PR #1, opened 2024-04-05
KAR-24 create helm chart to deploy ApiServerSource

skoved:


skoved:
I've add an explanation of what the helm chart deploys. I can edit that section if it confusing or if there is stuff that doesn't make sense.

Also, the ApiServerSource requires a that it is pointed to a sink so it has somewhere to send cloud events to. The helm chart deploys the example sink from the knative-eventing example, but we need replace it with a sink that we write for kubearchive

charts/kubearchive/crds/eventing-core.yaml
adambkaplan, line 1: It is most unfortunate that the "easy path" here is to install all of kNative eventing. I might file a feature request upstream to spin `APIServerSource` into its own deployment/helm chart.

adambkaplan (approved):
As good a start as any!

charts/kubearchive/crds/eventing-core.yaml
skoved: I agree. Do you need the rest of knative eventing to have an event source send a cloud event directly to a service that can receive cloud events or is the rest only necessary once you start using channels, brokers, etc.?

PR #15, opened 2024-04-29
Add integration tests base

rh-hemartin:
* Add example integration test file
* Add helper functions for integration tests


test/main.go
```diff
@@ -0,0 +1,51 @@
...
+	"os"
+	"os/exec"
+)
+
+const (
+	letterBytes   = "abcdefghijklmnopqrstuvwxyz"
+	randSuffixLen = 8
+)
+
+func RandomString() string {
```
skoved, line 16: I don't see where this function gets used

rh-hemartin: Sorry, I pushed it to show it you, but showed you my editor instead. The PR is a draft, I'm still considering some options when it comes to integration testing.

rh-hemartin: Now its being used.

test/integration/integration_test.go
```diff
@@ -0,0 +1,90 @@
...
+
+	client, err := test.GetKubernetesClient()
+
+	acc := 0.0
+	for {
+		if acc >= 30 {
+			t.Fatal("Timed out waiting for deployment to be ready.")
+		}
+
+		deploymentResource, err := client.AppsV1().Deployments(namespaceName).Get(context.TODO(), "kubearchive", metav1.GetOptions{})
```
skoved, line 71: I think that `context.Background()` should be used here instead
```suggestion
		deploymentResource, err := client.AppsV1().Deployments(namespaceName).Get(context.Background(), "kubearchive", metav1.GetOptions{})
```

from the [context docs](https://pkg.go.dev/context)

context.Background()

> It is typically used by the main function, initialization, and tests, and as the top-level Context for incoming requests

context.TODO()

> Code should use context.TODO when it's unclear which Context to use or it is not yet available (because the surrounding function has not yet been extended to accept a Context parameter).

go.mod
```diff
@@ -1,6 +1,8 @@
 module kubearchive
 
-go 1.21.9
+go 1.22.0
```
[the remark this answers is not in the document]
rh-hemartin: People can run

```
$ go install golang.org/dl/go1.22.2@latest
$ go1.22.2 download
```

 and that makes it available on the GOPATH. We will find a way to make these changes more explicit.

test/integration/integration_test.go
rh-hemartin: Done

skoved (approved):

PR #19, opened 2024-05-03
allow for the configuration of resources to be done per namespace

### LINKED ISSUE #16: one ApiServerSource per namespace
This will let will allow users to specify which resources should be archived in a namespace. 

Ex: resource a in namespace foo can be archived but resource a in namespace bar can be ignored

skoved:
This requires having one ApiServerSource per namespace where resources need to be archived. This has some impacts on what is deployed and how the ApiServerSources are configured.

- ApiServerSources no longer needs a namespace selector
- a service account must be created for each namespace
- a role must be created for each namespace
- a role binding must be created for each namespace

closes #16


maruiz93 (approved):

charts/kubearchive/values.yaml
```diff
@@ -8,33 +8,45 @@ kubearchive:
   # the name of the namespace to deploy kubearchive into
   namespace: "kubearchive"
   # configuration for a test namespace
```
beerparty, line 10: You removed the `testNamespace` section, so this comment should be moved to the `watchNamespaces` section where you define the `test` namespace.

beerparty (commented):
One minor suggestion for a comment.

charts/kubearchive/values.yaml
skoved: fixed!

PR #21, opened 2024-05-06
AuthZ middleware

### LINKED ISSUE #20: Add authorization middleware layer to the gin API
Add authorization middleware layer to the gin API

maruiz93:
Closes #20


skoved (approved):

cmd/api/main.go
```diff
@@ -4,42 +4,35 @@
...
-	pluralResourceType := c.Param("resourceType")
-	response := resources{
-		// FIXME The plural of the resource type can be different of adding an `s` to the kind
-		Kind:       pluralResourceType[:len(pluralResourceType)-1],
-		APIVersion: fmt.Sprintf("%s/%s", group, version),
-		Items:      []string{"resource1", "resource2"},
+var k8sClient = func() *kubernetes.Clientset {
+	config, err := rest.InClusterConfig()
+	if err != nil {
+		panic(err.Error())
```
skoved, line 21: Doesn't have to be right now, but at some point we should probably handle this more gracefully

cmd/api/main.go
```diff
@@ -4,42 +4,35 @@
...
 	}
-	c.IndentedJSON(http.StatusOK, response)
-}
+	client, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		panic(err.Error())
```
skoved, line 25: same as previous comment

cmd/api/auth/authz.go
```diff
@@ -0,0 +1,51 @@
...
+					Verb:      "get",
+				},
+			},
+		}, metav1.CreateOptions{})
+		if err != nil {
+			c.JSON(http.StatusInternalServerError, gin.H{"message": "Unexpected error on SAR"})
+			c.Abort()
+			return
+		}
+		if !sar.Status.Allowed {
```
rh-hemartin, line 44: Shouldn't we wait in between?

maruiz93: What do you mean? AFAIU, once we've got the SAR response we can check the status response (also, in the tests this works)

maruiz93: For reference and as I am still in the process of better understanding how go works regarding with routines and contexts, I analyzed the execution thread. AFAIU it is:

1. RBACAuthorization function is called with an implementation of [SubjectAccessReviewInterface](https://github.com/kubernetes/client-go/blob/master/kubernetes/typed/authorization/v1/subjectaccessreview.go#L37)
2. SubjectAccessReviewInterface is implemented by [subjectAccessReviews](https://github.com/kubernetes/client-go/blob/master/kubernetes/typed/authorization/v1/subjectaccessreview.go#L43) for the main execution.
3. authz middleware handler calls [Create](https://github.com/kubernetes/client-go/blob/master/kubernetes/typed/authorization/v1/subjectaccessreview.go#L55) of subjectaccessreview
4. In Create [Request](https://github.com/kubernetes/client-go/blob/master/rest/request.go#L99) is instantiated and from there, [Do](https://github.com/kubernetes/client-go/blob/master/rest/request.go#L1200) method is called.
5. Do calls the private method [request](https://github.com/kubernetes/client-go/blob/master/rest/request.go#L1104). This is the method in charge of performing the request, handling the retries and timeout, and canceling the context in case it's needed.

cmd/api/main.go
maruiz93: I've added a more verbose error in both panics. 
I used a panic as this would be on kubearchive API initialization error and if auth layer is enforced, there is no point on continue the API initialization.

Nevertheless, I think we may want to add a configuration parameter to k9e for allowing unauthenticated and unauthorized access.

Tekton results has this parameter in place. See related [code](https://github.com/tektoncd/results/blob/main/pkg/api/server/config/config.go#L30) and [docs](https://tekton.dev/docs/operator/tektonresult/#properties).

cmd/api/main.go
maruiz93: I've added a more verbose error message.

I've also included some comments to your suggestion. I'm resolving this conversation so we have only one thread.

cmd/api/auth/authz.go
rh-hemartin: Ok, if its a sync process from our perspective, that is fine.

cmd/api/routers/routers.go
```diff
@@ -0,0 +1,32 @@
...
+	group := c.Param("group")
+	version := c.Param("version")
+	pluralResourceType := c.Param("resourceType")
+	response := resources{
+		// FIXME The plural of the resource type can be different of adding an `s` to the kind
+		Kind:       pluralResourceType[:len(pluralResourceType)-1],
+		APIVersion: fmt.Sprintf("%s/%s", group, version),
+		Items:      []string{"resource1", "resource2"},
+	}
+	c.IndentedJSON(http.StatusOK, response)
```
beerparty, line 31: While useful for development, `IndentedJSON` should not be used here in the final code. If you want to keep it this way for development, you should add a `FIXME` so we don't forget to change it later.

```suggestion
        // FIXME Replace with 'c.JSON(http.StatusOK, response)' when development is complete
	c.IndentedJSON(http.StatusOK, response)
```

See https://pkg.go.dev/github.com/gin-gonic/gin#Context.IndentedJSON for details.

cmd/api/routers/routers.go
```diff
@@ -0,0 +1,32 @@
...
+}
+
+// GetAllResources responds with the list of resources of a specific type across all namespaces
+func GetAllResources(c *gin.Context) {
+	group := c.Param("group")
+	version := c.Param("version")
+	pluralResourceType := c.Param("resourceType")
+	response := resources{
+		// FIXME The plural of the resource type can be different of adding an `s` to the kind
+		Kind:       pluralResourceType[:len(pluralResourceType)-1],
```
beerparty, line 27: There is a go package you can use to convert a word from plural to singular form (instead of just trimming the last character):

https://github.com/gertd/go-pluralize

cmd/api/main.go
```diff
@@ -5,41 +5,35 @@ package main
...
+		panic(fmt.Sprintf("Error retrieving in-cluster k8s client config: %s", err.Error()))
 	}
-	c.IndentedJSON(http.StatusOK, response)
-}
+	client, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		panic(fmt.Sprintf("Error instantiating k8s from host %s: %s", config.Host, err.Error()))
+	}
+	return client
+}()
```
beerparty, line 29: Should these parentheses be here? I don't recognize this syntax.

beerparty (commented):
Some suggestions.

cmd/api/routers/routers.go
maruiz93: I've replaced IndentedJSON by JSON. It's short enough to be readable even if we are in a development phase. Also there are JSON parsers as browser plugins or we can use jq if we query the API from the command line.

cmd/api/routers/routers.go
maruiz93: I didn't want to use an external library for this. The resources can be made up names with a non-predictable singular/plural. 

As this is a mock response for now (we don't query the DB yet) I think we can go with whatever the name of the resource is, plural or singular, and we can face this in the future if we need to.

cmd/api/auth/authz.go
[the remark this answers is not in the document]
maruiz93: "Auth" is the way both "authentication" (usually shortened as authN) and "authorization" (authZ) is shortened. See [this blog post](https://dev.to/kleeut/authn-authz-what-is-42le).

We need both to be deferred to k8s, and this file just covers the authorization part (authZ). 
My plan is to add the authN file with the related middleware in the near future, as it is needed for the authorization.

There is no limitation to go with rbac_authorization.go but I think that it will be easier reading:

auth/authN.go
auth/authZ.go

cmd/api/auth/authz.go
```diff
@@ -0,0 +1,51 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package auth
+
+import (
+	"fmt"
+	"github.com/gin-gonic/gin"
+	authzv1 "k8s.io/api/authorization/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clientAuthzv1 "k8s.io/client-go/kubernetes/typed/authorization/v1"
```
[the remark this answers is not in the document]
maruiz93: The Z is because there is also the client for authentication which I will be using when I implement the middleware for the authN. See [authentication/v1](https://pkg.go.dev/k8s.io/client-go@v0.30.0/kubernetes/typed/authentication/v1)

The v1 is because the v1beta1 could be used, see [authorization/](https://pkg.go.dev/k8s.io/client-go/kubernetes/typed/authorization@v0.30.0)

Note that I'm also following the way this are used in the library. See [metav1 import](https://github.com/kubernetes/client-go/blob/v0.30.0/kubernetes/typed/authorization/v1/subjectaccessreview.go#L25).

It's also how things are imported in the [example](https://github.com/kubernetes/client-go/blob/v0.30.0/examples/in-cluster-client-configuration/main.go) of the library.

cmd/api/auth/authz.go
```diff
@@ -0,0 +1,51 @@
+// Copyright KubeArchive Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package auth
+
+import (
+	"fmt"
+	"github.com/gin-gonic/gin"
+	authzv1 "k8s.io/api/authorization/v1"
```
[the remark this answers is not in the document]
maruiz93: I'll solve this comment as it is being discussed in the other thread.

cmd/api/auth/authz.go
```diff
@@ -0,0 +1,51 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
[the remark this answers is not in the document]
maruiz93: I'll solve this as it is being discussed in a previous thread.

cmd/api/main.go
maruiz93: This parenthesis are for calling the function when it's declared.
This is called IIF (Immediately Invoked Function), a special usage of function.

You can read more about its usage [here](https://golangbyexample.com/function-golang-complete-guide/#IIF_or_Immediately_Invoked_Function)

cmd/api/auth/authz.go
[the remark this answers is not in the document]
maruiz93: I'll solve this but feel free to reopen it even if the PR is merged, I am open to do adjustments.

PR #23, opened 2024-05-06
utility functions for flushing spans and shutting down instrumentation

### LINKED ISSUE #22: provide utility functions for flushing completed spans and shutting down otel span processors
The [Batch Span Processor](https://pkg.go.dev/go.opentelemetry.io/otel/sdk@v1.26.0/trace#NewBatchSpanProcessor) runs concurrently with the program code and does not provide an guarantees for when spans will actually be processed and sent the exporter. This means that if the program shuts down, spans can be lost. If utility functions are provided to flush the span the batch span processors buffer or shut it down, this can be called in kubearchive to guarantee that spans get exported.

skoved:
closes #22

### DIFF
--- a/pkg/observability/trace.go
+++ b/pkg/observability/trace.go
@@ -5,6 +5,7 @@ package observability
 
 import (
 	"context"
+	"fmt"
 
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/exporters/stdout/stdouttrace"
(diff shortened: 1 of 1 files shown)


skoved:
I have them return errors if a tp is nil (only happens if Start has not been called), but they could be changed to become no-ops in that case instead. Not sure what the best behavior should be.

maruiz93 (approved):

PR #28, opened 2024-05-16
Use Server struct to share things

### LINKED ISSUE #27: Implement a Server struct
Currently we are instantiating the k8sClient globally, when the script is executed. There is an approach of creating a struct to share clients, loggers or other common pieces across handlers and other servers of the API[1]. We need to change the current implementation to make use of that pattern, so we can easily add stuff to share and we can easily mock them in tests.

[1]: https://grafana.com/blog/2024/02/09/how-i-write-http-services-in-go-after-13-years/

rh-hemartin:
* Add struct with k8sclient and router.
* Created a test passing a fake k8sclient.

Close #27

### DIFF
--- a/cmd/api/main_test.go
+++ b/cmd/api/main_test.go
@@ -1,43 +1,63 @@
 // Copyright KubeArchive Authors
 // SPDX-License-Identifier: Apache-2.0
-
 package main
 
 import (
-	"github.com/stretchr/testify/assert"
 	"net/http"
 	"net/http/httptest"
+	"net/http/httputil"
 	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"k8s.io/client-go/kubernetes/fake"
 )
 
-func TestRouterInvalidMethods(t *testing.T) {
-	router := setupRouter()
+func TestNewServer(t *testing.T) {
+	server := NewServer(fake.NewSimpleClientset())
+	res := httptest.NewRecorder()
+	req, _ := http.NewRequest("GET", "/", nil)
+	server.router.ServeHTTP(res, req)
 
+	assert.Equal(t, http.StatusUnauthorized, res.Code)
+}
+
+func TestAuthenticationMiddlewareIsConfigured(t *testing.T) {
 	tests := []struct {
-		name   string
-		method string
+		name     string
+		method   string
+		expected int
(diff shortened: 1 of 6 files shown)


cmd/api/main_test.go
```diff
@@ -1,43 +1,19 @@
-// Copyright KubeArchive Authors
-// SPDX-License-Identifier: Apache-2.0
```
maruiz93, line 2: You shouldn't remove license headers

cmd/api/routers/routers_test.go
```diff
@@ -31,7 +32,39 @@ func TestGetAllResources(t *testing.T) {
...
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			res := httptest.NewRecorder()
+			req, _ := http.NewRequest(tc.method, "/apis/stable.example.com/v1/crontabs", nil)
+			router.ServeHTTP(res, req)
+
+			assert.Equal(t, http.StatusNotFound, res.Code, res.Result())
+		})
+	}
+}
```
maruiz93, line 70: Does this work? You changed the name of `setupRouter()` function to `newServer()`.
This test tries out the router configuration, which is currently done on `main` package. With this refactor the logic is still on main, so I'd keep the test in `main_test.go` instead of `routers_test.go`.

Note that this isn't testing any of the logic defined on `routers/routers.go`

maruiz93 (changes requested):

cmd/api/main_test.go
rh-hemartin: Ops, my bad.

cmd/api/routers/routers_test.go
rh-hemartin: Due to the inclusion of the AuthN middleware it was not working, so I misunderstood what it was trying to test. You are right in that is not testing anything on the router.go, its just testing that the GET method from the library works as expected.

With the inclusion of the AuthN middleware, all unautenticated requests are unauthorized before they are rejected for being an invalid verb. So the test is not valid anymore. I recreated as "TestAuthenticationMiddlewareIsConfigured".

maruiz93 (approved):

PR #37, opened 2024-05-23
otel instrumentation only starts when environment variable is set

### LINKED ISSUE #36: only start otel if explicitly requested
Right now the otel instrumention for kubearchive api server is started automatically. It should only be started if user that installed kubearchive wants it. we could do this by having it look for an env var like `KUBEARCHIVE_OTEL_ENABLED=true`. The helm chart could be updated to set this env var automatically with the default being false

skoved:
OpenTelemetry instrumentation will only start if the environment variable `KUBEARCHIVE_OTEL_ENABLED=true`.

closes #36

### DIFF
--- a/charts/kubearchive/templates/api_server.yaml
+++ b/charts/kubearchive/templates/api_server.yaml
@@ -22,6 +22,8 @@ spec:
           command: ["./go/bin/dlv"]
           args: ["--listen=:40000", "--headless=true", "--api-version=2", "--log", "exec", "/ko-app/api"]
           {{- end}}
+          env:
+{{ include "kubearchive.v1.otel.env" .Values.apiServer | indent 12 }}
 ---
 kind: Service
 apiVersion: v1
--- a/charts/kubearchive/templates/_helpers.tpl
+++ b/charts/kubearchive/templates/_helpers.tpl
@@ -1,0 +1,18 @@
+{{/*
+Copyright KubeArchive Authors
+SPDX-License-Identifier: Apache-2.0
+*/}}
+
+{{/*
+Create environment variables for OpenTelemetry if .observability is set to true. Otherwise set KUBEARCHIVE_OTEL_ENABLED=false.
+This tells the OpenTelemtry instrumentation if it should start or not.
+*/}}
+{{- define "kubearchive.v1.otel.env" -}}
+{{- if .observability -}}
+- name: KUBEARCHIVE_OTEL_ENABLED
+  value: "true"
+{{- else -}}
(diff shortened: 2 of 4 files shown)


beerparty (approved):

skoved:
just a syncing the PR branch with `main`

maruiz93:
I the [Helm docs](https://helm.sh/docs/chart_template_guide/named_templates/#partials-and-_-files) I see that partials are supposed to have `.tpl` extension instead of `.tmpl`

charts/kubearchive/templates/api_server.yaml
```diff
@@ -22,6 +22,8 @@ spec:
           command: ["./go/bin/dlv"]
           args: ["--listen=:40000", "--headless=true", "--api-version=2", "--log", "exec", "/ko-app/api"]
           {{- end}}
+          env:
+{{ include "kubearchive.v1.otel.env" .Values.apiServer | indent 12 }}
```
maruiz93, line 26: I'm wondering if using the `template` directive will also work while keeping things simpler (I'm new to helm so I might not understand how this should be done, just read [this section](https://helm.sh/docs/chart_template_guide/named_templates/#declaring-and-using-templates-with-define-and-template) of the docs.

```suggestion
{{- template "kubearchive.v1.otel.env" .Values.apiServer }}
```

maruiz93 (changes requested):

skoved:
> I the [Helm docs](https://helm.sh/docs/chart_template_guide/named_templates/#partials-and-_-files) I see that partials are supposed to have `.tpl` extension instead of `.tmpl`

oops. that was a typo

charts/kubearchive/templates/api_server.yaml
skoved: I was going through that page on how templates work. The last section is about [include](https://helm.sh/docs/chart_template_guide/named_templates/#the-include-function). If we use the `template` directive it renders the template directly into the yaml and it will be formatted according to the indentation in its definition. The docs recommend using include because it will render the template into a string that can be passed into other functions. This allows the users of the templates to format the template as needed for their use. My goal is to be able to reuse this template for other deployments (like the sink) that will need otel environment variables and may have a different level of indentation. I know it's kind of ugly but this is what is stated as best practice and allows for easier reuse

maruiz93: ok now I understand better the reason, thank you for the thorough explanation :)

maruiz93 (approved):

PR #41, opened 2024-05-29
Add authentication middleware to the REST API

maruiz93:


skoved (approved):

cmd/api/auth/authn.go
```diff
@@ -0,0 +1,54 @@
...
+			Spec: authnv1.TokenReviewSpec{
+				Token: token,
+			},
+		}, metav1.CreateOptions{})
+		if err != nil {
+			abort(c, fmt.Sprintf("Unexpected error on TR: %s", err.Error()), http.StatusInternalServerError)
+			return
+		}
+		if !tr.Status.Authenticated {
+			abort(c, "Unauthenticated", http.StatusUnauthorized)
```
beerparty, line 48: Should this error message be "Unauthenticated" or should it be "Authentication failed"?

cmd/api/auth/authn_test.go
```diff
@@ -0,0 +1,108 @@
+// Copyright KubeArchive Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package auth
+
+import (
+	"context"
+	"github.com/gin-gonic/gin"
+	"github.com/stretchr/testify/assert"
+	v1 "k8s.io/api/authentication/v1"
```
beerparty, line 10:
```suggestion
	apiAuthn "k8s.io/api/authentication/v1"
```

Calling this package just "v1" is going to be confusing. I suggest "apiAuthn" to make it clear what package this is.

cmd/api/auth/authn_test.go
```diff
@@ -0,0 +1,108 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
beerparty, line 11:
```suggestion
	apisMeta "k8s.io/apimachinery/pkg/apis/meta/v1"
```

I think you can drop the "v1" from this abbreviation. I doubt we will be using v1 and v2 at the same time. But, you should add "apis" to make it more specific than "meta": I suggest "apisMeta".

cmd/api/auth/authz.go
```diff
@@ -4,49 +4,50 @@
 package auth
 
 import (
-	"log"
+	"fmt"
 	"net/http"
 
 	"github.com/gin-gonic/gin"
+	authnv1 "k8s.io/api/authentication/v1"
```
beerparty, line 11:
```suggestion
	apiAuthn "k8s.io/api/authentication/v1"
```

I think you can drop the "v1" from this abbreviation. I doubt we will be using v1 and v2 at the same time. But, you should add "api" to make it clearer what package this is.

cmd/api/auth/authz_test.go
```diff
@@ -5,9 +5,9 @@ package auth
 
 import (
 	"context"
-	"fmt"
 	"github.com/gin-gonic/gin"
 	"github.com/stretchr/testify/assert"
+	authnv1 "k8s.io/api/authentication/v1"
```
beerparty, line 10:
```suggestion
	apiAuthn "k8s.io/api/authentication/v1"
```

I think you can drop the "v1" from this abbreviation. I doubt we will be using v1 and v2 at the same time. But, you should add "api" to make it clearer what package this is.

cmd/api/auth/authn.go
rh-hemartin: We should use `http.StatusText(http.StatusUnauthorized)` if we don't want to use a custom text message.

cmd/api/auth/authn.go
```diff
@@ -0,0 +1,54 @@
...
+package auth
+
+import (
+	"errors"
+	"fmt"
+	"net/http"
+	"strings"
+
+	"github.com/gin-gonic/gin"
+	authnv1 "k8s.io/api/authentication/v1"
```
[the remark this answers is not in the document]
rh-hemartin: I think `authnv1` is the standard naming is used for it, so I would keep it.

cmd/api/auth/authn.go
```diff
@@ -0,0 +1,54 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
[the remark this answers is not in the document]
rh-hemartin: For example this: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/api/meta/interfaces.go#L20

cmd/api/auth/authn.go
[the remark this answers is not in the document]
rh-hemartin: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/api/meta/interfaces.go#L20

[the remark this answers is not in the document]
maruiz93: Agreed with Hector, @beerparty take a look at this other thread where I added some more context about this naming: https://github.com/kubearchive/kubearchive/pull/21#discussion_r1594201823

rh-hemartin (approved):

cmd/api/auth/authn_test.go
skoved: i think we should still keep `v1` in the name. using @beerparty's example: `apiAuthnv1`

cmd/api/auth/authz.go
skoved: i think that we should keep the `v1` in the name since the standard way of referring to k8s apis is to include the version number

rh-hemartin:
For the package import name lets use the style that Kubernetes is using, so its easier to recognize for people familiar with the Kubernetes Go ecosystem, we will get used to it eventually.

cmd/api/auth/authn.go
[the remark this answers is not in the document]
maruiz93: I'm solving this conversation as it seems that we agreed on keeping it with the standard naming

cmd/api/auth/authn_test.go
maruiz93: Solving it to keep with the standard practice

cmd/api/auth/authn.go
```diff
@@ -0,0 +1,54 @@
...
+	clientAuthnv1 "k8s.io/client-go/kubernetes/typed/authentication/v1"
```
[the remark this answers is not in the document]
maruiz93: Solving it to keep with the standard practice of showing the version (see other threads in the PR)

cmd/api/auth/authn.go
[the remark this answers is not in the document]
maruiz93: I will solve this as the standard way to alias those libraries is with `metav1`. See other threads of this PR for more context.

cmd/api/auth/authn.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+func extractBearerToken(header string) (string, error) {
+	if header == "" {
+		return "", errors.New("bad authorization bearer token given")
```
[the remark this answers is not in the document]
maruiz93: fixed!

cmd/api/auth/authn.go
maruiz93: I've chosen to use "Authentication failed" which is more accurate to what is happening. I don't want to set it as "Unauthorized" with `http.StatusText(http.StatusUnauthorized)` because the problem here would be the authentication and not the authorization.

cmd/api/auth/authn_test.go
maruiz93: I've changed it to Sam's proposal in this and other cases. I've chosen to include the `api` part as there is a `client` part (`clientAuthnv1 "k8s.io/client-go/kubernetes/typed/authentication/v1"`). Done this change in other similar cases throughout the `api` module.

cmd/api/auth/authz.go
maruiz93: Changed to `apiAuthnv1`. See previous thread for more info.

cmd/api/auth/authz_test.go
maruiz93: Changed to apiAuthnv1. See previous thread for more info.

cmd/api/auth/authn.go
```diff
@@ -0,0 +1,54 @@
...
+			abort(c, err.Error(), http.StatusBadRequest)
+			return
+		}
+		tr, err := tri.Create(c, &authnv1.TokenReview{
+			Spec: authnv1.TokenReviewSpec{
+				Token: token,
+			},
+		}, metav1.CreateOptions{})
+		if err != nil {
+			abort(c, fmt.Sprintf("Unexpected error on TR: %s", err.Error()), http.StatusInternalServerError)
```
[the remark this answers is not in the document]
maruiz93: Fixed!

PR #48, opened 2024-05-31
set propper env var on sink to enable otel instrumentation

### LINKED ISSUE #47: Update helm chart to add Otel env vars when observability is requested
the code to start the otel instrumentation expects that the the enviroment variable `KUBEARCHIVE_OTEL_ENABLED` is set to `true`. The helm chart needs to be updated so that environment variable can be properly set on the sink if the user requests instrumentation.

skoved:
closes #47

### DIFF
--- a/charts/kubearchive/templates/sink.yaml
+++ b/charts/kubearchive/templates/sink.yaml
@@ -1,4 +1,4 @@
-# yamllint disable rule:braces
+# yamllint disable-file
 ---
 apiVersion: apps/v1
 kind: Deployment
@@ -17,6 +17,8 @@ spec:
       containers:
         - name: {{ .Values.sink.name }}
           image: {{ .Values.sink.image }}
+          env:
+{{ include "kubearchive.v1.otel.env" .Values.sink | indent 12 }}
 ---
 kind: Service
 apiVersion: v1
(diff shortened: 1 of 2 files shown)


charts/kubearchive/templates/sink.yaml
```diff
@@ -17,6 +17,8 @@ spec:
       containers:
         - name: {{ .Values.sink.name }}
           image: {{ .Values.sink.image }}
+          env:
+{{ include "kubearchive.v1.otel.env" .Values.sink | indent 12 }}
```
rh-hemartin, line 21: I think we are hiding information in some way by doing it this way. I think that a more explicit approach works better:

```suggestion
          env:
          - name: KUBEARCHIVE_OTEL_ENABLED
            value: {{ .Values.sink.observability }}
```

While extracting the environment to a template is useful for very long environment scenarios, I don't think we are there yet, and we may never be.

rh-hemartin (approved):

charts/kubearchive/templates/sink.yaml
skoved: basically all configuration for otel instrumentation is done through environment variables. I haven't gotten to the point where we are deploying a collector, but that's the next thing on my list and having a template to make it easy to add the env vars to both the sink and the api server will make configuring the instrumentation for both components a lot easier to manage

PR #52, opened 2024-06-04
Add initial operator implementation.

ggallen:
This is the first cut at the operator. It needs work, but I want to get it committed and then start really integrating it into the project.  

The go files need work, so I'd rather not have comments now. They will be drastically altered moving ahead and there's plenty of time for that.

Based on a discussion with the operator SDK guys, this commit includes everything from the generation of the operator. The suggestion was to get it in the repository first, then start hacking.


rh-hemartin (approved):
I don't want to provide feedback because we could be blocking this for weeks before all the feedback from everyone is solved :rofl: I feel its better to move forward and solve problems we have using issues instead. Approved.

PR #55, opened 2024-06-07
Fix helm chart to include tokenreview permissions

maruiz93:
* Include permissions for the `kubearchive-api-server` service account to create TokenReview objects.
* Include a new service account called `kubearchive-privileged-user` with view permissions
* Update `DEVELOPMENT.md`

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -45,35 +45,40 @@ Install these tools:
 ## Install KubeArchive
 
 1. Use Helm to install KubeArchive:
-    ```bash
-    helm install kubearchive charts/kubearchive \
-        --set-string apiServer.image=$(ko build github.com/kubearchive/kubearchive/cmd/api) \
-        --set-string sink.image=$(ko build github.com/kubearchive/kubearchive/cmd/sink)
-    ```
+   ```bash
+   helm install kubearchive charts/kubearchive -n default \
+       --set-string apiServer.image=$(ko build github.com/kubearchive/kubearchive/cmd/api) \
+       --set-string sink.image=$(ko build github.com/kubearchive/kubearchive/cmd/sink)
+   ```
 1. Check that the KubeArchive deployments are Ready:
-    ```bash
-    kubectl get -n kubearchive deployments
-    ```
+   ```bash
+   kubectl get -n kubearchive deployments
+   ```
+   
+1. List the deployed helm chart
+   ```bash
+   helm list -n default  
+   ```
 
 ## Update KubeArchive
 
 After you make changes to the code use Helm to redeploy KubeArchive:
 
 ```bash
-helm upgrade kubearchive charts/kubearchive \
+helm upgrade kubearchive charts/kubearchive -n default \
     --set-string apiServer.image=$(ko build github.com/kubearchive/kubearchive/cmd/api) \
     --set-string sink.image=$(ko build github.com/kubearchive/kubearchive/cmd/sink)
 ```
 
-# Uninstall KubeArchive
+## Uninstall KubeArchive
 
 ```bash
-helm uninstall kubearchive
+helm uninstall -n default kubearchive
 ```
 
 ## Generate activity on the KubeArchive sink
 
-By default KubeArchive listens to `Event`s in the `test` namespace.
+By default, KubeArchive listens to `Event`s in the `test` namespace.
 
 1. Generate some activity creating a pod:
     ```bash
@@ -94,10 +99,17 @@ By default KubeArchive listens to `Event`s in the `test` namespace.
     ```bash
     kubectl get -n kubearchive secrets kubearchive-api-server-tls -o jsonpath='{.data.ca\.crt}' | base64 -d > ca.crt
     ```
+   
+1. **[ Optional ]** Create a service account with a specific role to test the REST API.
+   This Helm chart already provides `kubearchive-test-sa` with `view` privileges for testing purposes.
+    
 1. On a new terminal, use `curl` or your browser to perform a query:
     ```bash
-    curl --cacert ca.crt localhost:8081/apis/apps/v1/deployments
+    curl --cacert ca.crt localhost:8081/apis/apps/v1/deployments \
+    -H "Authorization: Bearer $(kubectl create token kubearchive-test-sa)" \
+   https://localhost:8081/apis/apps/v1/deployments
     ```
+
 1. Check the new logs on the KubeArchive API:
     ```bash
     kubectl logs -n kubearchive -l app=kubearchive-api-server
@@ -150,3 +162,12 @@ to start a debugger to which attach from your IDE.
     ```bash
     systemd-run -p Delegate=yes --user --scope kind create cluster
     ```
+1. Using KinD and Podman Desktop. If you get this error:
+   ```
+   Error: failed to publish images: error publishing 
+   ko://github.com/kubearchive/kubearchive/cmd/api: no nodes found for cluster "kind"
+   ```
+   expose the `KIND_CLUSTER_NAME` env variable with the appropriate name of the kind cluster:
+   ```bash
+   export KIND_CLUSTER_NAME=$(kubectl config get-clusters | grep -P '(?<=kind-).*' -o)
+   ```
--- a/DEVELOPER_RESOURCES.md
+++ b/DEVELOPER_RESOURCES.md
@@ -51,4 +51,6 @@ performance and behavior.
 Additional software that KubeArchive uses.
 
 * [Helm - Package Managment for Kubernetes](https://helm.sh/)
+* [Kind - Local Kubernetes Cluster](https://kind.sigs.k8s.io/)
+* [Podman Desktop - Includes Kind](https://podman-desktop.io/)
 
--- a/charts/kubearchive/templates/api_server/role.yaml
+++ b/charts/kubearchive/templates/api_server/role.yaml
@@ -6,7 +6,9 @@ metadata:
 rules:
   - apiGroups:
       - authorization.k8s.io
+      - authentication.k8s.io
     resources:
       - subjectaccessreviews
+      - tokenreviews
     verbs:
       - create
--- a/charts/kubearchive/templates/api_server/role_binding.yaml
+++ b/charts/kubearchive/templates/api_server/role_binding.yaml
@@ -11,3 +11,16 @@ roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
   name: {{ .Values.apiServer.name }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: {{ .Values.apiServer.testSA }}
+subjects:
+  - kind: ServiceAccount
+    name: {{ .Values.apiServer.testSA }}
+    namespace: {{ .Values.kubearchive.namespace }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: view
--- a/charts/kubearchive/templates/api_server/service_account.yaml
+++ b/charts/kubearchive/templates/api_server/service_account.yaml
@@ -4,3 +4,9 @@ kind: ServiceAccount
 metadata:
   name: {{ .Values.apiServer.name }}
   namespace: {{ .Values.kubearchive.namespace }}
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ .Values.apiServer.testSA }}
+  namespace: {{ .Values.kubearchive.namespace }}
--- a/charts/kubearchive/values.yaml
+++ b/charts/kubearchive/values.yaml
@@ -61,6 +61,7 @@ apiServer:
   debug: false
   # If true, OpenTelemetry instrumention will be enabled for the api server
   observability: false
+  testSA: "kubearchive-test-sa"
 
 # values used to create a sink
 sink:


rh-hemartin:
What about calling the user "kubearchive-test-user" instead of "privileged"? "privileged" sounds like its here to be used, while I think we are intend to use it for testing purposes.

DEVELOPMENT.md
```diff
@@ -35,45 +35,54 @@ Install these tools:
     ```bash
     kind create cluster
     ```
+   **NOTE**: The KinD cluster can also be created with tools like 
+   [Podman Desktop](https://podman-desktop.io/docs/kind/creating-a-kind-cluster).
+
```
rh-hemartin, line 40: I don't think its necessary, people that wants to customize things can test their own tools, for those that are new the easy way is to use kind.

DEVELOPMENT.md
```diff
@@ -35,45 +35,54 @@ Install these tools:
...
+   **NOTE**: The KinD cluster can also be created with tools like 
+   [Podman Desktop](https://podman-desktop.io/docs/kind/creating-a-kind-cluster).
+ 
 
 1. Set up `ko` to upload images to the KinD cluster, or any other registry you
   want to use:
     ```bash
     ```
+   **NOTE**: if your cluster doesn't have the name `kind` you need to specify the env var `KIND_CLUSTER_NAME`.
```
rh-hemartin, line 47: I don't think this is needed. People that want to go further and make an advanced usage of tools can do it by their own.

DEVELOPMENT.md
```diff
@@ -35,45 +35,54 @@ Install these tools:
...
 ## Install KubeArchive
 
 1. Use Helm to install KubeArchive:
-    ```bash
-    helm install kubearchive charts/kubearchive \
-        --set-string apiServer.image=$(ko build github.com/kubearchive/kubearchive/cmd/api) \
-        --set-string sink.image=$(ko build github.com/kubearchive/kubearchive/cmd/sink)
-    ```
+   ```bash
+   helm install kubearchive charts/kubearchive -n default \
```
rh-hemartin, line 53: This is not needed. `-n` populates the `{{ .Release.Namespace }}` variable which we are not using. We also don't want to install anything on default. If anything, this should be `-n kubearchive`, but again we are not using the variable.

https://helm.sh/docs/chart_template_guide/builtin_objects/

DEVELOPMENT.md
```diff
@@ -35,45 +35,54 @@ Install these tools:
...
-    kubectl get -n kubearchive deployments
-    ```
+   ```bash
+   kubectl get -n kubearchive deployments
+   ```
+   
+1. List the deployed helm chart
+   ```bash
+   helm list -n default  
+   ```
```
rh-hemartin, line 65: You are autoindenting on three spaces, we can create an .editorconfig to prevent these inconsistencies.

DEVELOPMENT.md
```diff
@@ -92,8 +101,13 @@ By default KubeArchive listens to `Event`s in the `test` namespace.
     ```
 1. On a new terminal, use `curl` or your browser to perform a query:
     ```bash
-    curl localhost:8081/apis/apps/v1/deployments
+    curl -H "Authorization: Bearer $(kubectl create token kubearchive-privileged-user)" \
+   localhost:8081/apis/apps/v1/deployments
     ```
+   **NOTE**: To test with an unauthorized user replace `kubearchive-privileged-user` with 
+   an unprivileged service account like `default`.
```
rh-hemartin, line 108: I don't think this belongs here, as we are not guiding the contributors into manual testing, but just teaching them how to generate traffic and do basic a basic GET on the API. If we want to add some manual testing section I would do that in the normal documentation (which is not added to the repository yet).

charts/kubearchive/templates/role_binding.yaml
```diff
@@ -28,3 +28,16 @@ roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
   name: {{ .Values.apiServer.name }}
+---
```
rh-hemartin, line 31: Giving a second thought, adding a SA with cluster-wide `view` in a non-specified namespace feels like a security issue. Remove it and we can discuss how to add it in a separate issue.

rh-hemartin (changes requested):
I have some things I want you to address. They are open to discussion.

DEVELOPMENT.md
maruiz93: I think that using podman desktop is pretty useful for testing purposes with local kubernetes cluster. Of course it isn't needed but it is a side note and I think it's useful to have this information somewhere (it is what I would like to read when onboarding).

I am open to change it to another section if you think it's more clear, e.g., an "Optional" section under "Requirements" at the beginning of this file.

DEVELOPMENT.md
maruiz93: I had to look for this when I started to use `ko` and not knowing this made face issues. I'm open to move it elsewhere, maybe to the "Known issues" section, but I feel this should be written somewhere.

DEVELOPMENT.md
maruiz93: I'm confused about this. AFAIU, when running `helm install` the chart will be installed in a namespace of the kubernetes cluster. If we run this just after we deploying the KinD cluster, the kubernetes context will be the `default` namespace so the chart will be installed there.

I think that it is better to explicitly state the namespace.

Note that after installing a chart, it can be listed by running `helm  list`. If you aren't in the same namespace that you where when running `helm install`, you won't be able to see the chart. 

If we add the namespace in all the instructions, they become more "reproducible" and prevent finding errors.

DEVELOPMENT.md
maruiz93: yes, my guess is that my editor is autoaligning the code blocks with the start of the paragraph. If we want to enforce it to a 4 space indent, it's better to set that in a `.editorconfig` file. I will check that out.

DEVELOPMENT.md
maruiz93: I can change the perspective and tell this information in the "known issues" section. 
I want people to know that if they don't use a token from a privileged service account, they won't get the expected output.

charts/kubearchive/templates/role_binding.yaml
maruiz93: Note that in a production cluster, the users will have different permissions and they will be probably linked to an oauth server. This is for testing purposes and there are no users deployed by default in the kind cluster apart from the admin one so we need to include one with some "get" privileges.

We could add a specific role for that or use the "view" one that is given by default in the cluster.

As the helm chart is used for deploying a testing environment I don't think that we should remove this and wait for another issue to be done because that will make the REST API unusable. I prefer to keep this and if we want to provide a more restrictive role to test the API in a different issue, remove this in that moment.

DEVELOPMENT.md
rh-hemartin: What about moving it to DEVELOPMENT_RESOURCES.md? I want to have this DEVELOPMENT.md guide as concise, narrow and slim as possible, to not overwhelm newcomers.

maruiz93:
> What about calling the user "kubearchive-test-sa" instead of "privileged"? "privileged" sounds like its here to be used, while I think we are intend to use it for testing purposes. Also, calling it "user" is kind of misleading as it not really a user, its a service account. I don't know how different they are from a authentication and autorization point of view.

I have no strong preferences, I can change it to `kubearchive-test-sa`.

BTW, a service account is only different from a user from the perspective of being defined inside the cluster instead of outside of it, but from the auth POV, they are all the same, a subject with some permissions assigned.

Note that for example, the [TokenReview status](https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/token-review-v1/#TokenReviewStatus) includes the information of this "subject" under the `user` field, it doesn't matter if the user is defined inside the k8s cluster (service account) or outside of it.

DEVELOPMENT.md
rh-hemartin: After testing a little bit... even if we declare `-n` to be default, the chart is installed on the kubearchive namespace as we have it coded in the files. If we want to use `-n` to install the chart somewhere we need to change our implementation.

So I think that both having `-n default` and not having it are not correct, because our implementation does not make good use of it. I suggest we don't specify it and address the problem in a separate issue.

charts/kubearchive/templates/role_binding.yaml
rh-hemartin: If the helm chart is just scoped for development then its fine to keep it as is.

DEVELOPMENT.md
rh-hemartin: I created #57 to address this, I assigned it to you, feel free to unassign it if you don't want to work on it.

charts/kubearchive/templates/service_account.yaml
```diff
@@ -19,3 +19,9 @@ kind: ServiceAccount
 metadata:
   name: {{ .Values.apiServer.name }}
   namespace: {{ .Values.kubearchive.namespace }}
+---
```
skoved, line 22: why have another service account `kubearchive-privileged-user` to use for token reviews instead of granting the service account ` kubearchive-api-server` the role to do token reviews instead

cmd/sink/main.go
```diff
@@ -28,7 +28,7 @@ func main() {
 	if err != nil {
 		logger.Printf("Could not start tracing: %s\n", err.Error())
 	}
-	httpClient, err := cloudevents.NewHTTP(
+	httpClient, _ := cloudevents.NewHTTP(
```
skoved, line 31: I agree that there is an issue here, but I think the right solution is to add the same kind of `err != nil` check that is on line 38 after the function call here. Not ignoring the returned error entirely. I must have missed this (and everyone that reviewed my PR :sweat_smile: )

skoved (changes requested):

charts/kubearchive/templates/service_account.yaml
rh-hemartin: I think the privileged user is granted the `view` cluster, not permissions to use token reviews. Its here: https://github.com/kubearchive/kubearchive/pull/55/files/dd96f229bbe3bc0f85742f18e514c4141af9bc73#diff-b68ce0b4f2cb84ca9c9429162a3b4fd00358b3e93c215148d26494da45e9ee99R33

skoved: you're right, `kubearchive-privileged-user` is granted the `view` role not token reviews.

Looking at the changes in development.md, is this service account just meant for testing?

rh-hemartin: Yes, just for testing. See https://github.com/kubearchive/kubearchive/pull/55#discussion_r1632650976

cmd/sink/main.go
maruiz93: yes!! I'll address the error as you suggest, just added this line because I was advertise of it in https://github.com/maruiz93/kubearchive/actions/runs/9416549251/job/25939886827

The good news is that with the latest updates on the CI we won't miss this kind of things anymore :)

charts/kubearchive/templates/service_account.yaml
maruiz93: I'm solving this as the user is for testing purposes. With the new name `kubearchive-test-sa` should be more clear

ggallen: Do we really need to have `-sa` on a service account name?

I personally am not a fan.

maruiz93: I initally thought about having `user` in the name, because a service account is a user. We can change it to `kubearchive-test-user`.

The `test` part over the `privileged` based on the discussions on this an other thread I think is a good replacement, as the main point is to state that is a service account for testing.

ggallen: My point is the `-sa` or -`user` is redundant.  It's a service account. If I try want to retrieve it, I have to specify that I want a service account, i.e. `kubectl get sa`.  If I run `kubectl get all`, it's listed under service accounts.

If we do it here, are we going to do it also for every role, rolebinding, deployment, etc?

I've added this to the agenda for Thursday.

cmd/sink/main.go
maruiz93: I've undone my change as the error management is already in place

charts/kubearchive/templates/service_account.yaml
maruiz93: Ok let's discuss it on Thursday.

BTW, I would usually say that we shouldn't specify what the resource is in the naming (the same way it isn't encourage to have the type included in a variable).

But, I found this a special case. The user is going to retrieve a token from a service account when doing a `curl` like the one described in the `DEVELOPMENT.md` file:

```
curl --cacert ca.crt localhost:8081/apis/apps/v1/deployments \
    -H "Authorization: Bearer $(kubectl create token kubearchive-test-sa)" \
   localhost:8081/apis/apps/v1/deployments
```

I think that having `kubearchive-test` is less clear. So I would normally go with your suggestion but treat this as an exception to the rule.

I think we aren't including the type of the resource in the rest of the helm chart values, this is the only one.

DEVELOPMENT.md
rh-hemartin: After talking with Marta, let's keep this with `-n default` as Marta introduced. For me its confusing that `-n` does not impact our resources, but not specifying it can lead to confusion when updating the helm chart if the user moved from the original namespace.

rh-hemartin (approved):

charts/kubearchive/templates/service_account.yaml
ggallen: It's on the agenda for tomorrow. But I still don't find your argument convincing. In your `curl` command is a nested `kubectl` command.  That command requires a service account. From the help:
```
[gallen@gallen-thinkpadp1gen5 kubearchive] (merge-operator-build)$ kubectl create token --help
Request a service account token.
...
[gallen@gallen-thinkpadp1gen5 kubearchive]
```

PR #59, opened 2024-06-11
Add cert-manager for API server

### LINKED ISSUE #58: Introduce cert-manager selfsigned for the API
We need to support TLS communication on the API. We will be using cert-manager and a self signed issuer for started, then we will refine the implementation to provide more options and to cover more components.

rh-hemartin:
* Add the cert-manager v1.15 version to helm crds
* Add a template file that generates a CA cert and a certificate for the API
* Change the API server to mount the certificate and use it
* Add instructions to get the CA certificate to query the API with curl

Closes #58

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -90,9 +90,13 @@ By default KubeArchive listens to `Event`s in the `test` namespace.
     ```bash
     kubectl port-forward -n kubearchive svc/kubearchive-api-server 8081:8081
     ```
+1. Get the Certificate Authority (CA) from the `kubearchive-api-server-tls` secret:
+    ```bash
+    kubectl get -n kubearchive secrets kubearchive-api-server-tls -o jsonpath='{.data.ca\.crt}' | base64 -d > ca.crt
+    ```
 1. On a new terminal, use `curl` or your browser to perform a query:
     ```bash
-    curl localhost:8081/apis/apps/v1/deployments
+    curl --cacert ca.crt localhost:8081/apis/apps/v1/deployments
     ```
 1. Check the new logs on the KubeArchive API:
     ```bash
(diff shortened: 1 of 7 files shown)


DEVELOPMENT.md
```diff
@@ -90,9 +90,13 @@ By default KubeArchive listens to `Event`s in the `test` namespace.
     ```bash
     kubectl port-forward -n kubearchive svc/kubearchive-api-server 8081:8081
     ```
+1. Get the Certificate Authority (CA) from the `kubearchive-api-server-tls` secret:
+    ```bash
+    kubectl get -n kubearchive secrets kubearchive-api-server-tls -o jsonpath='{.data.ca\.crt}' | base64 -d > ca.crt
```
mafh314, line 95: could you add '$' at the beginning of the command?

DEVELOPMENT.md
```diff
@@ -90,9 +90,13 @@ By default KubeArchive listens to `Event`s in the `test` namespace.
...
+    ```
 1. On a new terminal, use `curl` or your browser to perform a query:
     ```bash
-    curl localhost:8081/apis/apps/v1/deployments
+    curl --cacert ca.crt localhost:8081/apis/apps/v1/deployments
```
mafh314, line 99: could you add '$' at the beginning of the command?

DEVELOPMENT.md
skoved: for code blocks like this, github will generate a button to copy all of the text. If the `$` is added at the beginning, you have to go to the beginning of the line to remove the `$` if you paste the command into your terminal which I think is kind of annoying

rh-hemartin: That's it, I'm keeping the commands without dollar sign so they are easier to copy.

mafh314 (approved):

PR #63, opened 2024-06-12
add linters to golangci-lint

skoved:
bodyclose       - enforce that http response bodies get closed
containedctx  - enforce that contexts do not get placed as fields in a struct
fatcontext       - enforce that contexts do not get nested in loops
gofmt               - enforce the golang code style
nilerr                - enforce that an error is not returned as nil if the code checks the there error is not actually nil
usestdlibvars  - detect values that can be replaced with vars/consts from go standard library
wastedassign - detect vars that are reassigned and the new value is never used

Also fix code so that it passes all of the lint checks

### DIFF
--- a/.github/workflows/static_code_analysis.yml
+++ b/.github/workflows/static_code_analysis.yml
@@ -28,17 +28,3 @@ jobs:
         with:
           yamllint_strict: false
           yamllint_comment: false
-  goftm:
-    name: "Golang format test"
-    runs-on: ubuntu-latest
-    steps:
-      - run: echo "Job triggered by ${{ github.actor }}."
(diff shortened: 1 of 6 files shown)


skoved:
closes #63

skoved:
[you can see all linters avaible here](https://golangci-lint.run/usage/linters/)

mafh314 (approved):

skoved:
the pending check ****Golang format test**** is the gofmt action that I removed since I added gofmt to golangci-lint. When ready, I'll push the merge through

rh-hemartin (approved):

skoved:
syncing this PR with main

PR #65, opened 2024-06-12
organize templates used by kubearchive helm charts

### LINKED ISSUE #64: organize yaml templates in the helm chart
Right now the directory templates in the helm chart has a flat structure. All of the templates for all of the components are all mixed together. I think that they should be organized in the following way to make it easier to find resources and what they belong to

- each kubearchive component gets a sub directory of templates (ex: `templates/sink`, `templates/api`, `templates/operator`, etc.
- yaml template files in each sub directory should ideally contain either a single k8s resource or a set of k8s resources that are all related (ex: a `deployment` and it's associated `service` and `route` definitions or a template that will have helm loop to create multiple resources from the same template)

skoved:
Make it easier to find templates by putting them in folders based on which component of kubearchive they are for.

I didn't touch the rest of the templates since many of them will change when the helm chart will start deploying the operator

closes #64


mafh314 (approved):

skoved:
syncing this PR with main

PR #66, opened 2024-06-12
Refactor/merge operator code.

ggallen:
No real changes other than merging the operator code into the repository and cleaning up unnecessary files.


mafh314:
Hey!
I think that for the `[braces] too many spaces inside braces` yamllint errors, you can add
`# yamllint disable rule:braces` at the beginning of the yaml file

.gitignore
```diff
@@ -19,3 +19,8 @@
 
 # Go workspace file
 go.work
+
+# Files generated by the operator.
+charts/kubearchive/crds/kubearchive.kubearchive.org_kubearchiveconfigs.yaml
+charts/kubearchive/templates/operator/role.yaml
+cmd/operator/api/v1alpha1/zz_generated.deepcopy.go
```
rh-hemartin, line 26: This is the pattern that is included with the operator-sdk gitignore which covers for extra files in case we need them.
```suggestion
!vendor/**/zz_generated.*
```

charts/kubearchive/templates/operator/auth_proxy_service.yaml
```diff
@@ -0,0 +1,21 @@
+---
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    control-plane: controller-manager
+    app.kubernetes.io/name: {{ .Values.kubearchive.namespace }}-service
+    app.kubernetes.io/instance: {{ .Values.kubearchive.namespace }}-operator-metrics-service
```
rh-hemartin, line 8: You need to quote it to avoid the error. This will happen case by case until you cover them all, so you need to quote every "mixed" string (interpolation + literal string).

```suggestion
    app.kubernetes.io/instance: "{{ .Values.kubearchive.namespace }}-operator-metrics-service"
```

charts/kubearchive/templates/operator/kubearchiveconfig_editor_role.yaml
```diff
@@ -5,12 +5,11 @@ kind: ClusterRole
 metadata:
   labels:
     app.kubernetes.io/name: clusterrole
-    app.kubernetes.io/instance: kubearchiveconfig-editor-role
+    app.kubernetes.io/instance: {{ .Values.kubearchive.namespace }}config-editor-role
```
rh-hemartin, line 8: This looks like it is missing a dash.

```suggestion
    app.kubernetes.io/instance: {{ .Values.kubearchive.namespace }}-config-editor-role
```

cmd/operator/bin/controller-gen
rh-hemartin, line 1: I don't think we need to submit the binary itself. We can have some script that downloads the tools needed.

rh-hemartin (changes requested):
I left some comments

charts/kubearchive/templates/operator/auth_proxy_service.yaml
ggallen: We should be running yamllint on the files *after* they are processed.

cmd/operator/bin/controller-gen
ggallen: Actually, I have the script `generate.sh` in this PR which does just that. I'll remove the bin directory from git. 

Good catch.

go.mod
skoved, line 1: Since you made some changes to here, did you run `go mod tidy` after?

skoved (changes requested):

go.mod
```diff
@@ -1,49 +1,74 @@
 module github.com/kubearchive/kubearchive
 
-go 1.21.9
+go 1.22.0
```
skoved, line 3: I think we should leave of the `.0` at the end here. This will let compile with the latest version of go 1.22 without having to change this value. That's what most go projects do

.github/workflows/go_test.yml
```diff
@@ -29,6 +29,8 @@ jobs:
           go-version: '1.21.9'
```
skoved, line 29: you'll need to change the go version here

```suggestion
          go-version: '1.22'
```

cmd/operator/test/e2e/e2e_test.go
skoved, line 1: I think that instead of having the e2e tests in `cmd/operator/test` it should be in `test/e2e`

cmd/operator/test/e2e/e2e_test.go
```diff
@@ -11,7 +11,7 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 
-	"github.com/kubarchive/kubearchive/test/utils"
+	"github.com/kubearchive/kubearchive/operator/test/utils"
```
skoved, line 14: I think this needs to be updated

```suggestion
	"github.com/kubearchive/kubearchive/cmd/operator/test/utils"
```

cmd/operator/test/e2e/e2e_suite_test.go
skoved, line 1: I think that instead of having the e2e tests in `cmd/operator/test` it should be in `test/e2e`

cmd/operator/test/utils/utils.go
skoved, line 1: instead of naming the this `utils.go` is there a more descriptive name that can be used?

skoved: also, if this file is just helper functions for testing, it might make more sense to put it in the `test` directory at the root of the repo

cmd/operator/bin/controller-gen
skoved, line 1: should we be committing binary files?

charts/kubearchive/templates/operator/auth_proxy_service.yaml
ggallen: This is no longer an issue since https://github.com/kubearchive/kubearchive/pull/71 has been committed and CI now runs over the templated files.

cmd/operator/bin/controller-gen
ggallen: No. I have removed this.

.gitignore
ggallen: @rh-hemartin, if I make this change the file isn't ignored.

rh-hemartin: No problem then

ggallen: @rh-hemartin , does it work for you?

rh-hemartin: I tested it and it does not work, the thing that works is:

```suggestion
zz_generated.*
```

Maybe we can go with this and the moment we vendor then we add the other line to include generated files in vendorized dependencies.

cmd/operator/test/e2e/e2e_test.go
ggallen: I guess, but why?   These are operator-specific tests, and I think it makes sense to organize the tests by what they are testing.

cmd/operator/test/e2e/e2e_suite_test.go
ggallen: Again, why?  These are operator-specific tests.

These files are also generated initially by the operator-sdk.  I'd rather not move them around, yet, until I understand more of how they work.

cmd/operator/test/utils/utils.go
ggallen: They are helper functions for the operator tests. I'm not sure they make sense for any other tests.

And again, if we move the tests themselves later out of the operator directory, we can move this as well.

rh-hemartin (approved):
I share some of the ideas that Sam has about moving more things into the root directory. I also have some more things I would like to review, but I don't want to delay this too much. Approved

cmd/operator/test/e2e/e2e_suite_test.go
skoved: sure. we can leave them for now

cmd/operator/test/utils/utils.go
skoved: sure

cmd/operator/test/e2e/e2e_test.go
skoved: other go projects I've seen generally keep unit tests colocated with the code that they test, but e2e tests are usually all together in a separate `test` directory

skoved (approved):

PR #69, opened 2024-06-13
Run helm to generate templated files for yamllint.

ggallen:


charts/kubearchive/templates/api_server.yaml
```diff
@@ -14,7 +13,7 @@ spec:
     metadata:
       labels: *labels
     spec:
-      serviceAccountName: {{ .Values.apiServer.serviceAccount }}
+      serviceAccountName: {{ .Values.kubearchive.serviceAccount }}
```
skoved, line 16: I'm pretty sure the name of the service account is .Values.apiServer.name
https://github.com/kubearchive/kubearchive/blob/main/charts/kubearchive/templates/service_account.yaml#L20

ggallen: @maruiz93, can you confirm that this fix is correct?

ggallen: Ah, @skoved, I see what you are saying. I'll wait for @maruiz93 to confirm which is correct.

ggallen: @skoved, I think you are right. I'll change this!

.github/workflows/static_code_analysis.yml
```diff
@@ -23,6 +23,13 @@ jobs:
       - run: echo "Branch ${{ github.ref }} of repository ${{ github.repository }}."
       - name: Check out repository code
         uses: actions/checkout@v4
+      - name: Install helm
+        uses: azure/setup-helm@v4.2.0
+        with:
+          version: latest
+        id: install
+      - name: Run helm and template locally
+        run: helm template charts/kubearchive/ --output-dir charts/generated
```
skoved, line 32: would it make more sense to run `helm lint` instead of `helm template`?

https://helm.sh/docs/helm/helm_lint/

.github/workflows/static_code_analysis.yml
```diff
@@ -23,6 +23,13 @@ jobs:
       - run: echo "Branch ${{ github.ref }} of repository ${{ github.repository }}."
       - name: Check out repository code
         uses: actions/checkout@v4
+      - name: Install helm
+        uses: azure/setup-helm@v4.2.0
+        with:
+          version: latest
+        id: install
+      - name: Run helm and template locally
```
skoved, line 31: This should be its own job, not a step in the yamllint job

skoved (changes requested):

.github/workflows/static_code_analysis.yml
skoved, line 32: This should be its own job not a couple of steps in the `yamllint` job

skoved (changes requested):

.github/workflows/static_code_analysis.yml
ggallen: I don't think it does in the context of yamllint, which is what we are trying to run.

We're not linting the chart itself (which is what helm lint does), we're linting the output of the helm templating.

skoved: if the decision is made to switch to `helm lint` the helm stuff should get moved to its own job

ggallen: Unless `helm lint` runs `yamllint` under the covers, I think we'll always need to do this.

We have to run `helm` here to get the fully-templated file for `yamllint`.  I don't see that ever changing, unless we abandon `helm` completely.  This exercise already found one problem, which shows that it's important.

PR #82, opened 2024-06-18
sink writes cloudevents to the database

skoved:
Need to make this work with the full db schema, clean up the code, and figure out a better solution for connecting to the database. This works though. You just need to deploy the database before installing kubearchive and update the host to use the IP of the database service


cmd/sink/main.go
```diff
@@ -5,22 +5,151 @@ package main
...
+		if err != nil {
+			logger.Printf("failed to connect to database: %s", err)
+			return
+		}
+	}
 
+	eventExtensions := event.Extensions()
+	// since a call to validateCloudevent should already have passed we can assume that all of these values exist as
+	// strings
+	apiVersion, _ := eventExtensions[apiVersionExtension]
```
beerparty, line 93: To fix this so you don't get the security warning:

```suggestion
	apiVersion := eventExtensions[apiVersionExtension]
```

cmd/sink/main.go
```diff
@@ -5,22 +5,151 @@ package main
...
+	kind, _ := eventExtensions[kindExtension]
```
beerparty, line 94:
```suggestion
	kind := eventExtensions[kindExtension]
```

cmd/sink/main.go
```diff
@@ -5,22 +5,151 @@ package main
...
+	name, _ := eventExtensions[nameExtension]
```
beerparty, line 95:
```suggestion
	name := eventExtensions[nameExtension]
```

cmd/sink/main.go
```diff
@@ -5,22 +5,151 @@ package main
...
+	namespace, _ := eventExtensions[namespaceExtension]
```
beerparty, line 96:
```suggestion
	namespace := eventExtensions[namespaceExtension]
```

cmd/sink/main.go
```diff
@@ -5,22 +5,151 @@ package main
...
 	ceClient "github.com/cloudevents/sdk-go/v2/client"
+	"github.com/cloudevents/sdk-go/v2/types"
 	kaObservability "github.com/kubearchive/kubearchive/pkg/observability"
+	_ "github.com/lib/pq"
 	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
 )
 
-var logger = log.New(os.Stderr, "", log.LstdFlags|log.Lmicroseconds|log.LUTC)
+const (
+	connStr             = "user=kubearchive password=kubearchive dbname=kubearchive-db host=10.96.22.175 port=5432 sslmode=disable"
```
rh-hemartin, line 24: To avoid using the IP:

```suggestion
	connStr             = "user=kubearchive password=kubearchive dbname=kubearchive-db host=<db-service-name>.<db-namespace>.svc.cluster.local port=5432 sslmode=disable"
```

cmd/sink/main.go
skoved: done

cmd/sink/main.go
skoved: done

cmd/sink/main.go
skoved: done

cmd/sink/main.go
skoved: using the service url for the database now. I'm gonna modify the helm chart so that all of these values get added as env vars to the sink when it is deployed

skoved:
the sink connects to the database through the internal cluster url for the service. The url and other connection info is harded coded so I need to modify the helm chart to add that info as env vars when the sink and database are deployed. I also need to rebase the PR since there have been changes to the operator and helm chart since last week

skoved:
the test failures are in the operator code that has not been changed by this PR

.golangci.yml
```diff
@@ -13,3 +13,7 @@ linters:
     - nilerr
     - usestdlibvars
     - wastedassign
+issues:
+  exclude-files:
+    - '(.+)_test\.go'
+    - cmd/operator/internal/controller/kubearchiveconfig_controller.go
```
rh-hemartin, line 19: Could you add some comment to reflect WHY these files are being excluded?

skoved: my bad. Added explainations

skoved:
I got the integration test working by adding the environment variables that it expects to have to create a database connection. Since it's doesn't receive any cloudevents *yet* during the test this works fine. Once it starts receiving cloudevents, we might start to see issues since it will actually try to connect to a database

.golangci.yml
```diff
@@ -13,3 +13,6 @@ linters:
     - nilerr
     - usestdlibvars
     - wastedassign
+issues:
+  exclude-files:
+    - cmd/operator/internal/controller/kubearchiveconfig_controller.go # errcheck fails on lines 55-58. I tried to fix it but that caused tests to fail. TODO: fix file and remove this line
```
beerparty, line 18: To fix the lint error, you just need to add another space before the comment character.

```suggestion
    - cmd/operator/internal/controller/kubearchiveconfig_controller.go  # errcheck fails on lines 55-58. I tried to fix it but that caused tests to fail. TODO: fix file and remove this line
```

cmd/sink/event_data.go
```diff
@@ -0,0 +1,138 @@
...
+const (
+	apiVersionExtension = "apiversion"
+	kindExtension       = "kind"
+	nameExtension       = "name"
+	namespaceExtension  = "namespace"
+
+	writeQuery = "INSERT INTO ArchiveMeta (api_version, kind, name, namespace, resource_version, data, created_ts, updated_ts) Values ($1, $2, $3, $4, $5, $6, $7, $8)"
+)
+
+// Represents the fields in a resource's metadata that are need to write a resource into the database
```
beerparty, line 24: Tense, should be "needed":

```suggestion
// Represents the fields in a resource's metadata that are needed to write a resource into the database
```

cmd/sink/event_data.go
```diff
@@ -0,0 +1,138 @@
...
+func (entry *ArchiveEntry) LastUpdated() string {
+	return entry.Data.LastUpdated
+}
+
+// the version of the resources
+func (entry *ArchiveEntry) ResourceVersion() string {
+	return entry.Data.Metadata.ResourceVersion
+}
+
+// checks that the cloudevents has the appropriate extensions set and have values that are the right type. Additionally
```
beerparty, line 66: Should be "has":

```suggestion
// checks that the cloudevents has the appropriate extensions set and has values that are the right type. Additionally
```

cmd/sink/event_data.go
```diff
@@ -0,0 +1,138 @@
...
+			Data:       payload,
+			ApiVersion: apiVersion,
+			Kind:       kind,
+			Name:       name,
+			Namespace:  namespace,
+		},
+		nil
+}
+
+// Writes the kubernetes resource held by entry to the database using the provided connection string. Timeouts and
```
beerparty, line 103: The method takes a connection (dbConn), not a connection string:

```suggestion
// Writes the kubernetes resource held by entry to the database using the provided DB connection. Timeouts and
```

cmd/sink/event_data.go
```diff
@@ -0,0 +1,138 @@
...
+		return fmt.Errorf("write to database failed: %s", execErr)
+	}
+	execErr = tx.Commit()
+	if execErr != nil {
+		rollbackErr := tx.Rollback()
+		if rollbackErr != nil {
+			return fmt.Errorf("write to database failed: %s and unable to rollback transaction: %s", execErr, rollbackErr)
+		}
+		return fmt.Errorf("write to database failed: %s", execErr)
+	}
```
beerparty, line 136: The label in these error messages is the same as above, which means when they appear in the log you will not be able to tell which spot in the code failed. I suggest updating the text label to make the action clear.

```suggestion
	execErr = tx.Commit()
	if execErr != nil {
		rollbackErr := tx.Rollback()
		if rollbackErr != nil {
			return fmt.Errorf("commit to database failed: %s and unable to rollback transaction: %s", execErr, rollbackErr)
		}
		return fmt.Errorf("commit to database failed: %s", execErr)
	}
```

cmd/sink/main.go
```diff
@@ -5,29 +5,58 @@ package main
...
 	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
 )
 
-var logger = log.New(os.Stderr, "", log.LstdFlags|log.Lmicroseconds|log.LUTC)
+var (
+	logger         = log.New(os.Stderr, "", log.LstdFlags|log.Lmicroseconds|log.LUTC)
+	db     *sql.DB = nil
+)
 
+// Processes incoming cloudevents and
```
beerparty, line 28: Comment is incomplete (ends with "and")

cmd/sink/main.go
```diff
@@ -5,29 +5,58 @@ package main
...
 	kaObservability "github.com/kubearchive/kubearchive/pkg/observability"
+	_ "github.com/lib/pq"
 	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
 )
 
-var logger = log.New(os.Stderr, "", log.LstdFlags|log.Lmicroseconds|log.LUTC)
+var (
+	logger         = log.New(os.Stderr, "", log.LstdFlags|log.Lmicroseconds|log.LUTC)
+	db     *sql.DB = nil
+)
```
rh-hemartin, line 26: Don't use global variables, follow the same pattern than the API: create a type struct called `Server` (or `Sink` if you want) and have logger and db be two properties of that struct. Then `receive` would be a method of that struct and will get the database connection as `s.db` and the logger as `s.logger` instead of  just `db`. For reference:

https://github.com/kubearchive/kubearchive/blob/main/cmd/api/main.go#L44

This allows us for a better testing experience, in which we can initialize the `Sink` struct with a fake db connection on each test, instead on having a shared global variable that depends on the order of execution of the tests.

skoved:
testing this PR
1. check out the branch
2. revert commits that have the operator deploy the ApiServerSource (it is currently broken) `git revert 4674b82c1208423c1005d8c09b016c1377e01643` and `git revert 6c84ed5937d60da284b65cc71fd7b9878901851a`
3. deploy kubearchive: `./cmd/operator/generate.sh; helm install kubearchive charts/kubearchive --create-namespace -n kubearchive --set-string apiServer.image=$(ko build [github.com/kubearchive/kubearchive/cmd/api](http://github.com/kubearchive/kubearchive/cmd/api)) --set-string sink.image=$(ko build [github.com/kubearchive/kubearchive/cmd/sink](http://github.com/kubearchive/kubearchive/cmd/sink)) --set-string operator.image=$(ko build [github.com/kubearchive/kubearchive/cmd/operator](http://github.com/kubearchive/kubearchive/cmd/operator))`
4. get the name of the postgres pod `kubectl -n kubearchive get pods`
5. in a separate terminal, port forward the database `kubectl port-forward -n kubearchive svc/postgresql 5432:5432`
6. in another separate terminal, use psql to login to the database: `psql -h localhost -U ps_user --password -p 5432 postgresdb`  (default password can be found in the secret template)
7. initialize the ArchiveMeta table: `CREATE TABLE IF NOT EXISTS ArchiveMeta ("id" serial PRIMARY KEY, "api_version" varchar NOT NULL, "kind" varchar NOT NULL, "name" varchar NOT NULL, "namespace" varchar NOT NULL, "resource_version" varchar NULL, "created_ts" timestamp NOT NULL, "updated_ts" timestamp NOT NULL, "data" jsonb NOT NULL);`
8. run: `\dt` in psql to verify ArchiveMeta was created (its case insensitive)
9. back in the first terminal run: `kubectl run -n test busybox --image=busybox`
10. then run `kubectl logs -n kubearchive -l app=kubearchive-sink -f` you should see logs indicating that the sink receive cloudevents and that it successful wrote them to the database
11. in the psql terminal run: `SELECT * FROM ArchiveMeta;` This should print entries that have been added by the sink

rh-hemartin:
> 7\. nitialize the ArchiveMeta table: `CREATE TABLE IF NOT EXISTS ArchiveMeta ("id" serial PRIMARY KEY, "api_version" varchar NOT NULL, "kind" varchar NOT NULL, "name" varchar NOT NULL, "namespace" varchar NOT NULL, "resource_version" varchar NULL, "created_ts" timestamp NOT NULL, "updated_ts" timestamp NOT NULL, "data" jsonb NOT NULL);`

Rename the table we are creating with `database/init_db.go` and the table where the test objects are loaded.

cmd/sink/main.go
skoved: I'll see if I can get that approach to work with the cloudevent sdk

rh-hemartin: Ah, I didn't take that into account. Give it a go and if you find yourself spending too much, create a ticket and we handle it next.

.golangci.yml
skoved: fixed

cmd/sink/event_data.go
skoved: fixed

cmd/sink/event_data.go
skoved: fixed

cmd/sink/event_data.go
skoved: fixed

cmd/sink/main.go
skoved: finished the comment

skoved:
> > 7. nitialize the ArchiveMeta table: `CREATE TABLE IF NOT EXISTS ArchiveMeta ("id" serial PRIMARY KEY, "api_version" varchar NOT NULL, "kind" varchar NOT NULL, "name" varchar NOT NULL, "namespace" varchar NOT NULL, "resource_version" varchar NULL, "created_ts" timestamp NOT NULL, "updated_ts" timestamp NOT NULL, "data" jsonb NOT NULL);`
> 
> Rename the table we are creating with `database/init_db.go` and the table where the test objects are loaded.

I've changed the name of the table in `database/init_db.go` and `test_objects.sql`

cmd/sink/main.go
skoved: it works! the cloudevents sdk has no issue calling a function or a struct method and there's no more global variables

ggallen:
I thought we no longer had an `ArchiveMeta` table, but had just checked to `resource`?

skoved:
> I thought we no longer had an `ArchiveMeta` table, but had just checked to `resource`?

The last time I looked at [the database architecture](https://miro.com/app/board/uXjVNjKPncc=/), it still had the `ArchiveMeta` table. I must have missed the discussion where that changed. I'll change the table name to resource to `Resource`

ggallen:
I think postgres naming conventions are all lower case and also using underscores.

mafh314:
> I thought we no longer had an ArchiveMeta table, but had just checked to resource?

> The last time I looked at [the database architecture](https://miro.com/app/board/uXjVNjKPncc=/), it still had the ArchiveMeta table. I must have missed the discussion where that changed. I'll change the table name to resource

my bad, I should add a shadow to the old db model.

database/init_db.go
```diff
@@ -50,7 +50,7 @@ func main() {
 
 	// SQL instruction to create a table.
 	sqlStatement := `
-	CREATE TABLE IF NOT EXISTS public.test_objects (
+	CREATE TABLE IF NOT EXISTS resource (
```
mafh314, line 53: I'm not sure about what happens here if we create a table outside the `public` schema.

mafh314:
> I think postgres naming conventions are all lower case and also using underscores. Maybe @mafh314 has another change coming with the names changed? I'm not sure where the current schema stands.
Yes, I have a change with the table rename, 3 new fields (delete_cluster_ts, cluster, cluster_uid) and the required changes in the test objects.

database/init_db.go
skoved: I'll change it to public.resource

skoved:
> Yes, I have a change with the table rename, 3 new fields (delete_cluster_ts, cluster, cluster_uid) and the required changes in the test objects.

I'm not sure where to find the cluster name or cluster_uid. The cloudevents from ApiServerSource don't contain them

rh-hemartin (approved):

mafh314 (approved):

skoved:
synced with main

Issue #60, opened 2024-06-11
Linter for helm charts


maruiz93:
In the development process I've forgot to update a name of a value usage in the helm chart after a refactor.

This made me wonder about the possibility of adding a linter that for the helm chart in the Github CI.

Helm has the [lint](https://helm.sh/docs/helm/helm_lint/) command that allows to check the chart. This can be combined with the usage of the [required](https://helm.sh/docs/howto/charts_tips_and_tricks/#using-the-required-function) function when using mandatory values so the linter fails if no value is found.

- Include the required function when needed on the usage of helm values in the helm templates
- Add `helm lint` command in the Github actions CI

References:
- https://github.com/helm/helm/issues/12014

maruiz93:
While checking the best way of running `helm lint` I found this [Red Hat Communities of Practice Doc](https://redhat-cop.github.io/) for good practices on checking Helm charts.

maruiz93:
I'm going to close this issue and open separate ones for adding the `required` keyword now that the `helm lint` command is included in the CI.

PR #87, opened 2024-06-20
Include ct-lint in the static_code_analysis workflow

maruiz93:
ct-lint includes:
- helm lint
- version checking
- YAML schema validation on Chart.yaml
- YAML linting for Chart.yaml and values.yaml

### DIFF
--- a/.github/workflows/static_code_analysis.yml
+++ b/.github/workflows/static_code_analysis.yml
@@ -14,8 +14,8 @@ permissions:
   contents: read
   pull-requests: read
 jobs:
-  yamllint:
-    name: "YAML lint"
+  helm-lint:
+    name: "Helm lint"
     runs-on: ubuntu-latest
     steps:
       - run: echo "Job triggered by ${{ github.actor }}."
@@ -23,12 +23,15 @@ jobs:
       - run: echo "Branch ${{ github.ref }} of repository ${{ github.repository }}."
       - name: Check out repository code
         uses: actions/checkout@v4
-      - name: Install helm
+      - name: Set up Helm
         uses: azure/setup-helm@v4.2.0
         with:
           version: latest
-        id: install
-      - name: Run helm and template locally
+      - name: Set up chart-testing
+        uses: helm/chart-testing-action@main
+      - name: Run chart-testing (lint)
+        run: ct lint --charts charts/kubearchive --config ct.yaml
+      - name: Run helm template
         run: helm template charts/kubearchive/ --output-dir charts/generated
       - name: Run yamllint
         uses: karancode/yamllint-github-action@master
--- a/.yamllint.yml
+++ b/.yamllint.yml
@@ -3,7 +3,7 @@
 ---
 extends: default
 ignore:
-  - charts/kubearchive
+  - charts/kubearchive/
 rules:
   line-length: disable
   truthy: disable
--- a/charts/kubearchive/Chart.yaml
+++ b/charts/kubearchive/Chart.yaml
@@ -2,6 +2,7 @@
 apiVersion: v2
 name: kubearchive
 description: A Helm chart that deploys kubearchive on a Kubernetes cluster
+icon: https://avatars.githubusercontent.com/u/153536228?s=48&v=4
 
 # A chart can be either an 'application' or a 'library' chart.
 #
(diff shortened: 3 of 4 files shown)


maruiz93:
Related to #60

lintconf.yaml
```diff
@@ -0,0 +1,42 @@
...
+    max-spaces-after: 1
+  commas:
+    max-spaces-before: 0
+    min-spaces-after: 1
+    max-spaces-after: 1
+  comments:
+    require-starting-space: true
+    min-spaces-from-content: 2
+  document-end: disable
+  document-start: disable           # No --- to start a file
```
beerparty, line 24: Why are you disabling this check? Unless there is a reason I would prefer to enforce YAML document start requirement.

lintconf.yaml
```diff
@@ -0,0 +1,42 @@
...
+    max-end: 0
+  hyphens:
+    max-spaces-after: 1
+  indentation:
+    spaces: consistent
+    indent-sequences: whatever      # - list indentation will handle both indentation and without
+    check-multi-line-strings: false
+  key-duplicates: enable
+  line-length: disable              # Lines can be any length
+  new-line-at-end-of-file: disable
```
beerparty, line 37: Why are you disabling this check? A file without a final new line has caused us issues in the past, so if there is not a specific reason to disable this check I would prefer it stay enabled.

lintconf.yaml
ggallen: I also think it conflicts with a different linter that is requiring it!

lintconf.yaml
rh-hemartin, line 1: This file looks a lot as a .yamllint file. Any reason to use this over the current one (`.yamllint`)?

chart_schema.yaml
rh-hemartin, line 1: Why do we want to control the schema of the `Chart.yaml`? This file is a specification from Helm. Also when the `ct` tool is installed with the action, a `chart_schema.yaml` file is provided.

lintconf.yaml
maruiz93: I followed the instructions from https://redhat-cop.github.io/ci/linting-testing-helm-charts.html

I can try to disable it.

Note that the `lintconf.yaml` is used by ct to just check:
- values.yaml
- Chart.yaml

The rest is still being checked by the other yamllint

lintconf.yaml
maruiz93: I followed the instructions from https://redhat-cop.github.io/ci/linting-testing-helm-charts.html

I can try to disable it.

Note that the `lintconf.yaml` is used by `ct` to just check:

- values.yaml
- Chart.yaml

The rest is still being checked by the other yamllint.

lintconf.yaml
maruiz93: It is expected by the `ct lint` command used in the recommended practices for checking the Helm charts. See https://redhat-cop.github.io/ci/linting-testing-helm-charts.html

chart_schema.yaml
maruiz93: How is the `chart_schema.yaml` provided? By following [this instructions](https://redhat-cop.github.io/ci/linting-testing-helm-charts.html) I understood that `ct` needed that schema.

Also I modified it a bit because the `type` field wasn't included in those instructions schema.

Let me try to remove it and see if it's still checked or not.

maruiz93: It seems that I still see the following logs when the action runs:
```
Validating /home/runner/work/kubearchive/kubearchive/charts/kubearchive/Chart.yaml...
Validation success! 👍
```

But I'm still curious about where the `chart_schema.yaml` is provided in the action :thinking:

lintconf.yaml
maruiz93: But the tool used for linting is still `.yamllint`... let me see if I can just replace `lintconf.yaml` with `.yamllint`

maruiz93: Done!

lintconf.yaml
maruiz93: I use the `.yamllint` file for configuration to prevent having duplicated and different files.

chart_schema.yaml
rh-hemartin: Its here:

https://github.com/helm/chart-testing-action/blob/main/ct.sh#L130

They unpack their release tar.gz which contains the configuration files, then they set the configuration directory to the place they un packed the tar.gz.

rh-hemartin (approved):

PR #95, opened 2024-06-26
Add pre-install hook.

ggallen:
Add a pre-install hook to verify that all the pods in the `cert-manager` and `knative-eventing` namespaces are ready before installing resources. This hook runs after CRDs are installed but before helm installs any other resources.

### DIFF
--- a/charts/kubearchive/templates/pre-install.yaml
+++ b/charts/kubearchive/templates/pre-install.yaml
@@ -1,0 +1,77 @@
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: pre-install-view
+  annotations:
+    helm.sh/hook: pre-install
+    helm.sh/hook-delete-policy: hook-succeeded
+    helm.sh/hook-weight: "-10"
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: pre-install-view
+  annotations:
+    helm.sh/hook: pre-install
+    helm.sh/hook-delete-policy: hook-succeeded
+    helm.sh/hook-weight: "-10"
+rules:
+  - apiGroups:
+      - ""
+    resources:
+      - pods
+    verbs:
+      - get
+      - list
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: pre-install-view
+  annotations:
+    helm.sh/hook: pre-install
+    helm.sh/hook-delete-policy: hook-succeeded
+    helm.sh/hook-weight: "-10"
+subjects:
+  - kind: ServiceAccount
+    name: pre-install-view
+    namespace: "{{ .Release.Namespace }}"
+    roleRef:
+      kind: ClusterRole
+      name: pre-install-view
+      apiGroup: rbac.authorization.k8s.io
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: "{{ .Release.Name }}"
+  labels:
+    app.kubernetes.io/managed-by: {{ .Release.Service | quote }}
+    app.kubernetes.io/instance: {{ .Release.Name | quote }}
+    app.kubernetes.io/version: {{ .Chart.AppVersion }}
+    helm.sh/chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+  annotations:
+    # This is what defines this resource as a hook. Without this line, the
+    # job is considered part of the release.
+    helm.sh/hook: pre-install
+    helm.sh/hook-weight: "-5"
+    helm.sh/hook-delete-policy: hook-succeeded
+spec:
+  template:
+    metadata:
+      name: "{{ .Release.Name }}"
+      labels:
+        app.kubernetes.io/managed-by: {{ .Release.Service | quote }}
+        app.kubernetes.io/instance: {{ .Release.Name | quote }}
+        helm.sh/chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    spec:
+      serviceAccount: pre-install-view
+      restartPolicy: Never
+      containers:
+        - name: pre-install-cert-manager
+          image: "k8s.gcr.io/hyperkube:v1.18.20"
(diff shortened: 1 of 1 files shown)


charts/kubearchive/templates/pre-install.yaml
```diff
@@ -0,0 +1,77 @@
...
+kind: ServiceAccount
+metadata:
+  name: pre-install-view
+  annotations:
+    helm.sh/hook: pre-install
+    helm.sh/hook-delete-policy: hook-succeeded
+    helm.sh/hook-weight: "-10"
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
```
rh-hemartin, line 12: Could we use two roles instead of a ClusterRole? Given that we know the namespaces, the less permissions we provide the better.

rh-hemartin:
Create a GitHub issue to associate this PR

maruiz93:
I believe we can avoid the pre-hook workaround if we take advantage of the dependencies feature of helm.

Also, we are applying cert-manager Deployments and Services at the same time as CRDs and that shouldn't be its usage.

This is the way of deploying cert-manager as a dependency shown in [their docs](https://cert-manager.io/docs/installation/helm/#installing-cert-manager-as-subchart)

And here is a PR I opened with the fix: https://github.com/kubearchive/kubearchive/pull/99

charts/kubearchive/templates/pre-install.yaml
ggallen: This no longer matters, but this role only lives during the installation.

ggallen:
This is superseded by https://github.com/kubearchive/kubearchive/pull/99.

charts/kubearchive/templates/pre-install.yaml
```diff
@@ -0,0 +1,77 @@
...
+        app.kubernetes.io/managed-by: {{ .Release.Service | quote }}
+        app.kubernetes.io/instance: {{ .Release.Name | quote }}
+        helm.sh/chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    spec:
+      serviceAccount: pre-install-view
+      restartPolicy: Never
+      containers:
+        - name: pre-install-cert-manager
+          image: "k8s.gcr.io/hyperkube:v1.18.20"
+          command: ["kubectl", "wait", "pod", "--all", "--for=condition=Ready", "--namespace=cert-manager", "--timeout=30s"]
```
maruiz93, line 74: This can still have problems with race conditions if the Deployments didn't create the pods at this point:

```suggestion
          command: ["kubectl", "rollout", "status", "deployment", "--namespace=cert-manager"]
```

charts/kubearchive/templates/pre-install.yaml
```diff
@@ -0,0 +1,77 @@
...
+        - name: pre-install-knative-eventing
+          image: "k8s.gcr.io/hyperkube:v1.18.20"
+          command: ["kubectl", "wait", "pod", "--all", "--for=condition=Ready", "--namespace=knative-eventing", "--timeout=30s"]
```
maruiz93, line 77: This can still have problems with race conditions if the Deployments didn't create the pods at this point:

```suggestion
          command: ["kubectl", "rollout", "status", "deployment", "--namespace=knative-eventing"]
```

maruiz93 (changes requested):

charts/kubearchive/templates/pre-install.yaml
rh-hemartin: I edited to add the webhook name to the suggestion, as it was used in https://github.com/kubearchive/kubearchive/pull/110#issuecomment-2196314239

charts/kubearchive/templates/pre-install.yaml
rh-hemartin: I edited to add the webhook name to the suggestion, mimicking the usage in https://github.com/kubearchive/kubearchive/pull/110#issuecomment-2196314239

charts/kubearchive/templates/pre-install.yaml
rh-hemartin: Ah, then maybe its worth checking all the deployments given that we are checking. `rollout status` also support `--timeout` so that will be also useful not to block the whole thing and fail early. I think it waits forever if there is no timeout

maruiz93: good point with the timeout, let me add it to the suggestion

charts/kubearchive/templates/pre-install.yaml
maruiz93: Added the timeout in this suggestion

rh-hemartin (approved):

skoved:
> This is superseded by #99.

I'm a little confused why this got closed and then re-opened

maruiz93:
> I'm a little confused why this got closed and then re-opened

Yes... It's probably worth to add a little more context of what happened here...

KubeArchive needs two operators to work:
* Knative (for the events for the sink)
* Cert-manager (for the apiserver to have tls)

When doing:

```
helm install kubearchive charts/kubearchive --namespace kubearchive --create-namespace \
    --set-string apiServer.image=$(ko build github.com/kubearchive/kubearchive/cmd/api) \
    --set-string sink.image=$(ko build github.com/kubearchive/kubearchive/cmd/sink) \
    --set-string operator.image=$(ko build github.com/kubearchive/kubearchive/cmd/operator)
```

Sometimes a race condition happened in which the pods of the cert-manager weren't deployed yet, so kubearchive wasn't successfully installed.

To manage this race condition there were two suggestions:
* The pre-hook in this PR
* Use helm dependencies, because helm waits them to be ready before continue with the chart installation. That's when we created https://github.com/kubearchive/kubearchive/pull/99.

Then we found some issues with the dependencies, stated in https://github.com/kubearchive/kubearchive/issues/112, so we revrted the dependencies changes: https://github.com/kubearchive/kubearchive/pull/113 and reopened this PR

skoved:
I think I understand now. With the pre-install hook, we won't have to manually install the dependencies as described in #117?

ggallen:
This is all superseded by https://github.com/kubearchive/kubearchive/pull/117.

Closing this PR.

PR #99, opened 2024-06-27
Fix race condition of cert-manager installation

maruiz93:
Separate the CRDs and the rest of the resources from the cert-manager opertator.

The CRDs are in the `crds/` chart directory while the operator is built and installed through helm dependencies.

This solves the race condition of having the cert-managers in ready state before installing the kubearchive chart resources.

The `crds/` directory should only allocate CRDs and not Services and Deployments.

Prepare knative-events to be also a chart dependency.


ggallen:
What about `knative-eventing` or anything else that's needed or will be needed in the future?

charts/kubearchive/crds/cert-manager.yaml
rh-hemartin, line 1: Do we need to install the CRDs from `cert-manager` ourselves? I see their usage defaults to false, but I don't think we should keep here their list of CRDs. If they change them, we need to be aware of that and change them too.

ggallen (approved):

charts/kubearchive/crds/cert-manager.yaml
rh-hemartin: From the conversation we had, we don't mind having these CRDs in our repository. We will update them when we need.

maruiz93:
> What about knative-eventing or anything else that's needed or will be needed in the future?

Although knative doesn't provide a Helm chart, they do provide the CRDs separated from the rest of the resources needed for the operator, so I just changed the disposition of the resources to add it as a dependency.

I believe we can do this for future dependencies too.

charts/kubearchive/crds/cert-manager.yaml
maruiz93: Yes, and we already had them previously.

Also note that in the [installing instructions](https://artifacthub.io/packages/helm/cert-manager/cert-manager#installing-the-chart) of the `cert-manager` helm's chart they instruct to run this:

`$ kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.15.1/cert-manager.crds.yaml`

which is the equivalent of having the CRDs in our chart.

Of course, another alternative would be to download it as a previous step, but I think it's not a big problem to keep it.

PR #124, opened 2024-07-02
Fix to use correct API group for ApiServerSource role.

ggallen:


cmd/operator/internal/controller/kubearchiveconfig_controller.go
```diff
@@ -149,7 +150,12 @@ func (r *KubeArchiveConfigReconciler) reconcileRole(ctx context.Context, kaconfi
 func (r *KubeArchiveConfigReconciler) desiredRole(kaconfig *kubearchivev1alpha1.KubeArchiveConfig) (*rbacv1.Role, error) {
 	var rules []rbacv1.PolicyRule
 	for _, resource := range kaconfig.Spec.Resources {
-		rules = append(rules, rbacv1.PolicyRule{APIGroups: []string{resource.APIVersion}, Resources: []string{resource.Kind}, Verbs: []string{"get", "list", "watch"}})
+		apiGroup := ""
+		slash := strings.Index(resource.APIVersion, "/")
+		if slash > -1 {
+			apiGroup = resource.APIVersion[:slash]
+		}
```
beerparty, line 157: If there is no slash in `resource.APIVersion`, is it okay that `apiGroup` is an empty string? In the previous code, the value was the entire string of `resource.APIVersion`. It seems like the code should have an else section setting the value of `apiGroup`.

ggallen: No, this is correct.  The API version should specify both the group and the version.  If there is no group, it is assumed to be part of the "core", and the `ApiGroup` value is "".

rh-hemartin (approved):

Issue #126, opened 2024-07-03
ApiServerSource created by the operator does not reach the Ready state.


ggallen:
There is an issue with the ApiServerSource deployed by the operator. For some reason it doesn't get into the ready state.  From what I can tell, the resources are identical to other resources that work when applied manually.

This is the error I am seeing on the ApiServerSource object:
```
    - lastTransitionTime: "2024-07-03T11:18:53Z"
      message: User system:serviceaccount:test:kubearchive cannot get, list, watch
        resource "eventses" in API group "" in Namespace "test"
      status: "False"
      type: SufficientPermissions
```

rh-hemartin:
The name `eventses` does not look right to me. Looks as it something was trying to pluralize it from "events" to "eventses". I think the `kind` on the KubeArchiveConfig on the test namespace should not have trailing `s`:

```
---
apiVersion: kubearchive.kubearchive.org/v1alpha1
kind: KubeArchiveConfig
metadata:
  name: kubearchive
  namespace: test
spec:
  resources:
    - apiVersion: v1
      kind: Events <--
```

ggallen:
@rh-hemartin, yeah, it does have to do with that.

PR #128, opened 2024-07-03
Fix issues with plurals and rolebinding.

ggallen:
Resolves https://github.com/kubearchive/kubearchive/issues/126.

Note that I am trying to figure out how to determine the plural correctly (rather than just appending "s").  Putting the PR out now just so if people need it.

### DIFF
--- a/cmd/operator/internal/controller/kubearchiveconfig_controller.go
+++ b/cmd/operator/internal/controller/kubearchiveconfig_controller.go
@@ -155,7 +155,10 @@ func (r *KubeArchiveConfigReconciler) desiredRole(kaconfig *kubearchivev1alpha1.
 		if slash > -1 {
 			apiGroup = resource.APIVersion[:slash]
 		}
-		rules = append(rules, rbacv1.PolicyRule{APIGroups: []string{apiGroup}, Resources: []string{strings.ToLower(resource.Kind)}, Verbs: []string{"get", "list", "watch"}})
+		rules = append(rules, rbacv1.PolicyRule{
+			APIGroups: []string{apiGroup},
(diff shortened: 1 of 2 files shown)


cmd/operator/internal/controller/kubearchiveconfig_controller.go
```diff
@@ -155,7 +155,7 @@ func (r *KubeArchiveConfigReconciler) desiredRole(kaconfig *kubearchivev1alpha1.
 		if slash > -1 {
 			apiGroup = resource.APIVersion[:slash]
 		}
-		rules = append(rules, rbacv1.PolicyRule{APIGroups: []string{apiGroup}, Resources: []string{strings.ToLower(resource.Kind)}, Verbs: []string{"get", "list", "watch"}})
+		rules = append(rules, rbacv1.PolicyRule{APIGroups: []string{apiGroup}, Resources: []string{strings.ToLower(resource.Kind) + "s"}, Verbs: []string{"get", "list", "watch"}})
```
beerparty, line 158: Does the go formatter support wrapped lines?

```suggestion
		rules = append(rules, rbacv1.PolicyRule{
		        APIGroups: []string{apiGroup},
		        Resources: []string{strings.ToLower(resource.Kind) + "s"},
		        Verbs: []string{"get", "list", "watch"}})
```

ggallen: This is temporary, but I want to get it checked in so things will work for people. I'll open another issue to fix this the right way.

skoved (approved):
It works! And not just for events. I tested it for other resource types too

beerparty (approved):

Issue #67, opened 2024-06-13
Enable YAML linting for helm files


rh-hemartin:
Currently most of our Helm files are disabled for yamllint linting. We should remove the ignore lines and make sure those files are in line with the yamllint linter. The files that exist on `crds` should be ignored, as they are dependencies that we do not control and don't want to fix for correctness.

rh-hemartin:
It is worth to mention that currently yamllint does not like `{{ variable }}` and that instead wants `{{variable}}`. I think the prefer the first, as everyone wrote the YAML files that way, so we should change the configuration a bit.

skoved:
also `{{ variable }}` with the space is the preferred helm syntax

maruiz93:
AFAIK, we are already addressing this although we are including `charts/kubearchive` to be ignored in the `.yamllint` config file (see [this line](https://github.com/kubearchive/kubearchive/blob/main/.yamllint.yml#L6)). 
In the `static-code-analysis` [action](https://github.com/kubearchive/kubearchive/blob/main/.github/workflows/static_code_analysis.yml), we are doing the following:

1. `ct lint`
2. `helm template`  saving the output in `charts/generated/`
3. run `yamllint` --> this will run on the charts/generated new directory with all the templated files, so I believe we are covered in that regard, and we don't need to worry about the `{{ variable }}` or `{{variable}}` spacing because we are linting the templated final files and not the templates with the variables.

rh-hemartin:
Implemented in #87

rh-hemartin:
Implemented in #87

PR #130, opened 2024-07-03
chore: apply naming conventions to database resources in helm chart

### LINKED ISSUE #104: Adjust Database templates in Helm Chart to the defined naming conventions

```yaml
# values
operator: 

# template

# output
kubearchive-operator
```

## Goal

Review and update the Helm Chart templates related to the kubearchive database component to comply with the naming conventions.

mafh314:
Closes #104

### DIFF
--- a/charts/kubearchive/templates/database/ps-claim.yaml
+++ b/charts/kubearchive/templates/database/ps-claim.yaml
@@ -4,13 +4,12 @@ apiVersion: v1
 kind: PersistentVolumeClaim
 metadata:
   labels:
-    app: {{ .Values.database.name }}
-  name: {{ .Values.database.name | default "postgresql" }}-volume-claim
-  namespace: {{ .Values.kubearchive.namespace }}
+    app: {{ tpl .Values.database.name . }}
+  name: {{ tpl .Values.database.name . }}
 spec:
   storageClassName: manual
   accessModes:
     - ReadWriteMany
   resources:
     requests:
-      storage: {{ .Values.database.volumes.storage | default "5Gi" }}
+      storage: {{ .Values.database.volumes.storage }}
--- a/charts/kubearchive/templates/database/ps-deployment.yaml
+++ b/charts/kubearchive/templates/database/ps-deployment.yaml
@@ -4,32 +4,31 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   labels:
-    app: {{ .Values.database.name | default "postgresql" }}
-  name: {{ .Values.database.name | default "postgresql" }}-deployment
-  namespace: {{ .Values.kubearchive.namespace }}
+    app: {{ tpl .Values.database.name . }}
+  name: {{ tpl .Values.database.name . }}
 spec:
-  replicas: {{ .Values.database.replicaCount | default 1}}
+  replicas: {{ .Values.database.replicaCount }}
   selector:
     matchLabels:
-      app: {{ .Values.database.name | default "postgresql"}}
+      app: {{ tpl .Values.database.name . }}
   template:
     metadata:
       labels:
-        app: {{ .Values.database.name | default "postgresql" }}
+        app: {{ tpl .Values.database.name . }}
     spec:
       containers:
-        - name: {{ .Values.database.name | default "postgresql" }}
+        - name: {{ tpl .Values.database.name . }}
           image: "{{ .Values.database.image.repository }}:{{ .Values.database.image.tag }}"
           imagePullPolicy: {{ .Values.database.image.pullPolicy }}
           ports:
             - containerPort: {{ .Values.database.service.port }}
           envFrom:
             - secretRef:
-                name: {{ .Values.database.name | default "postgresql" }}-secret
+                name: {{ tpl .Values.database.secret . }}
           volumeMounts:
             - mountPath: "/var/lib/postgresql/data"
-              name: {{ .Values.database.volumes.name | default "postgresdata" }}
+              name: {{ .Values.database.volumes.name }}
       volumes:
-        - name: {{ .Values.database.volumes.name | default "postgresdata" }}
+        - name: {{ .Values.database.volumes.name }}
           persistentVolumeClaim:
-            claimName: {{ .Values.database.name | default "postgresql" }}-volume-claim
+            claimName: {{ tpl .Values.database.name . }}
--- a/charts/kubearchive/templates/database/ps-pv.yaml
+++ b/charts/kubearchive/templates/database/ps-pv.yaml
@@ -5,13 +5,12 @@ kind: PersistentVolume
 metadata:
   labels:
     type: local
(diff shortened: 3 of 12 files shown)


beerparty (approved):

charts/kubearchive/templates/database/ps-claim.yaml
```diff
@@ -4,13 +4,13 @@ apiVersion: v1
 kind: PersistentVolumeClaim
 metadata:
   labels:
-    app: {{ .Values.database.name }}
-  name: {{ .Values.database.name | default "postgresql" }}-volume-claim
-  namespace: {{ .Values.kubearchive.namespace }}
+    app: {{ tpl .Values.database.name . }}
+  name: {{ tpl .Values.database.name . }}
+  namespace: {{ .Release.Namespace }}
```
rh-hemartin, line 9: I think we decided to remove the namespaces. Also, this line is redundant with the default behaviour. If no namespace is defined, the namespace is the defined by the release.

maruiz93: +1 to the line being redundant. One of the points defined in the naming conventions is that we will remove the namespace of the resource when they are "kubearchive" as we will be providing it from the `-n` option on the `helm install` command.

But I don't think we agreed on removing the spaces. I think you might be referring to the issue https://github.com/kubearchive/kubearchive/issues/67, but as Sam mentions, `{{ variable }}` is the preferred Helm syntax. Also, note that this won't be a problem from the YAML linter POV as the `yamllint` will be applied after the `helm template` command render the values on the `charts/generated/` directory.

charts/kubearchive/templates/database/ps-deployment.yaml
```diff
@@ -4,32 +4,32 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   labels:
-    app: {{ .Values.database.name | default "postgresql" }}
-  name: {{ .Values.database.name | default "postgresql" }}-deployment
-  namespace: {{ .Values.kubearchive.namespace }}
+    app: {{ tpl .Values.database.name . }}
+  name: {{ tpl .Values.database.name . }}
+  namespace: {{ .Release.Namespace }}
```
maruiz93, line 9: I suggest removing this based on the fact that the `kubearchive` namespace is provided by the `-n` option in `helm install` command.
```suggestion
```

charts/kubearchive/templates/database/ps-pv.yaml
```diff
@@ -5,13 +5,13 @@ kind: PersistentVolume
 metadata:
   labels:
     type: local
-    app: {{ .Values.database.name | default "postgresql" }}
-  name: {{ .Values.database.name | default "postgresql" }}-volume
-  namespace: {{ .Values.kubearchive.namespace }}
+    app: {{ tpl .Values.database.name . }}
+  name: {{ tpl .Values.database.name . }}
+  namespace: {{ .Release.Namespace }}
```
maruiz93, line 10: I suggest removing this based on the fact that the `kubearchive` namespace is provided by the `-n` option in `helm install` command.

```suggestion
```

charts/kubearchive/templates/database/ps-secret.yaml
```diff
@@ -5,9 +5,9 @@ kind: Secret
 type: Opaque
 metadata:
   labels:
-    app: {{ .Values.database.name | default "postgresql" }}
-  name: {{ .Values.database.name | default "postgresql"}}-secret
-  namespace: {{ .Values.kubearchive.namespace }}
+    app: {{ tpl .Values.database.name . }}
+  name: {{ tpl .Values.database.name . }}
+  namespace: {{ .Release.Namespace }}
```
maruiz93, line 10: I suggest removing this based on the fact that the `kubearchive` namespace is provided by the `-n` option in `helm install` command.

```suggestion
```

charts/kubearchive/templates/database/ps-service.yaml
```diff
@@ -4,12 +4,12 @@ apiVersion: v1
 kind: Service
 metadata:
   labels:
-    app: {{ .Values.database.name | default "postgresql" }}
-  name: {{ .Values.database.name | default "postgresql" }}
-  namespace: {{ .Values.kubearchive.namespace }}
+    app: {{ tpl .Values.database.name .  }}
+  name: {{ tpl .Values.database.name . }}
+  namespace: {{ .Release.Namespace }}
```
maruiz93, line 9: I suggest removing this based on the fact that the `kubearchive` namespace is provided by the `-n` option in `helm install` command.

```suggestion
```

database/README.md
```diff
@@ -29,7 +29,7 @@ Install [DBeaver Community](https://dbeaver.io/).
 These queries can be executed from the database.
 Connect to the database.
 ```bash
-$ kubectl exec -it [postgres_pod] -n [namespace] -- psql -h localhost -U [db_user] --password -p 5432 [db_name]
+$ kubectl exec -it [postgres_pod] -n kubearchive -- psql -h localhost -U [db_user] --password -p 5432 [db_name]
```
maruiz93, line 32: To make it easier for the reader, we can provide a way to automatically get the db pod based on the current naming:
```suggestion
$ kubectl exec -it  $(kubectl get -n kubearchive pods --no-headers -o custom-columns=":metadata.name" | grep database) -n kubearchive -- psql -h localhost -U ps_user --password -p 5432 postgresdb
```

maruiz93 (changes requested):

charts/kubearchive/templates/database/ps-claim.yaml
mafh314: removed the namespace`

charts/kubearchive/templates/database/ps-deployment.yaml
mafh314: removed the `namespace`

charts/kubearchive/templates/database/ps-pv.yaml
mafh314: removed the `namespace`

charts/kubearchive/templates/database/ps-secret.yaml
mafh314: removed the `namespace`

charts/kubearchive/templates/database/ps-service.yaml
mafh314: removed the `namespace`

database/README.md
mafh314: Changed

rh-hemartin (approved):

maruiz93 (approved):

rh-hemartin:
I think that is failing the mix between using `Release.Name` as name and also using a define function in `_helpers.tpl`. I suggest you remove the define function and follow the same approach as the secret name for the api.

rh-hemartin (approved):

PR #133, opened 2024-07-03
Add DB integration to the API

maruiz93:


maruiz93:
The query isn't working yet. Work in progress....

pkg/db/db.go
```diff
@@ -0,0 +1,51 @@
...
+
+type Resource struct {
+	Kind       string                 `json:"kind"`
+	ApiVersion string                 `json:"apiVersion"`
+	Spec       map[string]interface{} `json:"spec"`
+	Status     map[string]interface{} `json:"status"`
+	Metadata   map[string]interface{} `json:"metadata"`
+}
+
+func (r Resource) Value() (driver.Value, error) {
```
skoved, line 18: It might be worth adding a `String()` method to Resource as well so that it would also implement the [Stringer interface](https://pkg.go.dev/fmt#Stringer). Since `json.Marshall()` returns `[]byte` a it would need to do is cast the return from `Value()` to a string

skoved, line 18: It might be worth adding a String() method to Resource as well so that it would also implement the [Stringer interface](https://pkg.go.dev/fmt#Stringer). Since json.Marshall() returns []byte a it would need to do is cast the return from Value() to a string

skoved:
What is the difference between `model.Resource` and `db.Resource`?

maruiz93:
> What is the difference between `model.Resource` and `db.Resource`?

oops! I pushed the db file by mistake, I'll remove it from the PR...

pkg/db/db.go
maruiz93: That was part of some tests I've added (the whole `db.go` file. But in this PR I just want to add the things currently needed by the API, so for now I will just add the `Scan` method as I need it to unmarshall the sql records.

But we can keep in mind what you mention

maruiz93:
> The query isn't working yet. Work in progress....

Made the query work, `?` are only valid placeholders for Mysql, `$1`and `$2` are the appropriate ones for postgresql...

This type of differences could be prevented with an ORM

DEVELOPMENT.md
```diff
@@ -159,27 +158,25 @@ to start a debugger to which attach from your IDE.
 
 1. Deploy the chart with `ko` and `helm` in debug mode using an image with `delve`:
    ```bash
-   helm install -n default [deployment name] charts/kubearchive \ 
+   helm install kubearchive charts/kubearchive --create-namespace -n kubearchive \ 
    --set apiServer.debug=true \
-   --set-string apiServer.image=$(KO_DEFAULTBASEIMAGE=gcr.io/k8s-skaffold/skaffold-debug-support/go:latest \
-   ko build --disable-optimizations github.com/kubearchive/kubearchive/cmd/api) \
+   --set-string apiServer.imae=$(KO_DEFAULTBASEIMAGE=gcr.io/k8s-skaffold/skaffold-debug-support/go:latest ko build --disable-optimizations github.com/kubearchive/kubearchive/cmd/api) \
```
rh-hemartin, line 163: The `--debug` command on `ko build` does something similar, maybe you want to check if it works, so we can reduce this line.

https://ko.build/reference/ko_build/

```suggestion
   --set-string apiServer.imae=$(ko build --debug --disable-optimizations github.com/kubearchive/kubearchive/cmd/api) \
```

database/init_db.go
```diff
@@ -69,7 +69,7 @@ func main() {
 	fmt.Println("table test_objects created in the DB.")
 
 	// load the test data from file
-	testData := "test_objects.sql"
+	testData := "database/test_objects.sql"
```
rh-hemartin, line 72: Yes, its easier to execute from the root directoy. If you want to keep this change, please change the DEVELOPMENT.md

database/init_db.go
```diff
@@ -50,7 +50,7 @@ func main() {
 
 	// SQL instruction to create a table.
 	sqlStatement := `
-	CREATE TABLE IF NOT EXISTS public.test_objects (
+	CREATE TABLE IF NOT EXISTS public.archive (
```
rh-hemartin, line 53: https://github.com/kubearchive/kubearchive/pull/134/

```suggestion
	CREATE TABLE IF NOT EXISTS public.resource (
```

pkg/database/fake/database_generated.go
rh-hemartin, line 1: This is called `_generated.go`, how is this generated ?

rh-hemartin (approved):
Looks good to me

DEVELOPMENT.md
skoved: there's also a typo

```suggestion
   --set-string apiServer.image=$(KO_DEFAULTBASEIMAGE=gcr.io/k8s-skaffold/skaffold-debug-support/go:latest ko build --disable-optimizations github.com/kubearchive/kubearchive/cmd/api) \
```

pkg/database/database.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+	if err = db.Ping(); err != nil {
+		return nil, err
+	}
+	return &Database{db, "archive"}, nil
+}
+
+func (db *Database) QueryResources(kind, group, version string) ([]models.Resource, error) {
+	query := fmt.Sprintf( //nolint:gosec
+		"SELECT data FROM %s WHERE kind=$1 AND api_version=$2", db.resourceTableName) //nolint:gosec
```
skoved, line 39: I think that format string should be made a constant that is declared outside of this function so it will be easier to find and modify if we need to in the future.

Also instead of using `Sprintf` to insert the values being looked for, i think we should use [sql.NamedArg instead](https://pkg.go.dev/database/sql#NamedArg)

pkg/database/database.go
```diff
@@ -0,0 +1,57 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if err = db.Ping(); err != nil {
+		return nil, err
+	}
+	return &Database{db, "archive"}, nil
+}
+
+func (db *Database) QueryResources(kind, group, version string) ([]models.Resource, error) {
```
skoved, line 37: can you add `ctx context.Context` as the first parameter to this method? This will allow the pass a timeout to the database. This will also make it easier to add otel instrumentation for this later

```suggestion
func (db *Database) QueryResources(ctx context.Context, kind, group, version string) ([]models.Resource, error) {
```

pkg/database/database.go
```diff
@@ -0,0 +1,57 @@
...
+	query := fmt.Sprintf( //nolint:gosec
+		"SELECT data FROM %s WHERE kind=$1 AND api_version=$2", db.resourceTableName) //nolint:gosec
+	apiVersion := fmt.Sprintf("%s/%s", group, version)
+	rows, err := db.db.Query(query, kind, apiVersion)
```
skoved, line 41: I think we should use queryContext instead so that any timeouts or deadlines specified by ctx will be respected

```suggestion
	rows, err := db.db.QueryContext(ctx, query, kind, apiVersion)
```

skoved (changes requested):

pkg/database/database.go
```diff
@@ -0,0 +1,57 @@
...
+		return nil, err
+	}
+	db, err := sql.Open("postgres", dataSource)
+	if err != nil {
+		return nil, err
+	}
+	if err = db.Ping(); err != nil {
+		return nil, err
+	}
+	return &Database{db, "archive"}, nil
```
skoved, line 34: the table name should be `"resource"`. Also I think it should be a constant that is deleclared outside of this function to make it easier to find and modify in the future

DEVELOPMENT.md
maruiz93: I see that there is a `--debug` flag in the [docs](https://ko.build/features/debugging/) but I tried the latest version of `ko` (v0.15.4) and the flag doesn't seem to be recognized: 

```bash
ko build --debug --disable-optimizations github.com/kubearchive/kubearchive/cmd/api
Error: unknown flag: --debug
```

I'll keep the env variable `KO_DEFAULTBASEIMAGE` with the fix of the typo that skoved is suggesting

database/init_db.go
maruiz93: Fixed when rebasing

database/init_db.go
maruiz93: Updated!

pkg/database/database.go
maruiz93: updated!

pkg/database/database.go
maruiz93: updated!

pkg/database/database.go
maruiz93: I tried to use an approach using the sql.NamedArg (with that I could remove the `//nolint:gosec` comment too) but I didn't find a way to do it for the name of the tables.

If I'm not mistaken, based  on [this post in StackOverflow](https://stackoverflow.com/questions/65327759/sql-table-name-as-variable-to-query) I think that there is no way to provide the table name as a variable without doing a string substitution, but please feel free to correct me if there is something I missunderstood or overlooked.

For extracting the SQL query, I think that is a good idea, and I did the change :)

pkg/database/fake/database_generated.go
maruiz93: I'm open to change the name, the `_generated` suffix follows the same strategy used by the k8s [client-go library](https://github.com/kubernetes/client-go). 

Note that in this library there is also a [fake package](https://github.com/kubernetes/client-go/tree/master/kubernetes/fake) with a [clientset_generated.go](https://github.com/kubernetes/client-go/blob/master/kubernetes/fake/clientset_generated.go) that implements the "dummy implementation" of the  `Clientset` struct in [clientset.go](https://github.com/kubernetes/client-go/blob/master/kubernetes/clientset.go) file.

maruiz93:
I've included some tests for the queryResources method using [go-sqlmock libbrary](https://pkg.go.dev/github.com/data-dog/go-sqlmock)

pkg/database/database.go
skoved: I didn't realize that you couldn't use NamedArg for the table name

skoved (approved):

pkg/database/fake/database_generated.go
rh-hemartin: I understand that is manually generated then, no problem, we can leave it as is.

PR #137, opened 2024-07-04
Add baseline for integration tests

### LINKED ISSUE #129: deploy a database with the sink in the integration tests
[this integration test](https://github.com/kubearchive/kubearchive/blob/main/test/integration/integration_test.go#L24) deploys the sink. When the sink starts, it looks for a database that it can connect to. If the e2e test is deploying the sink, then it also needs to deploy a database for the sink to connect to and provide the appropriate environment variables to the container so that the sink can build the connections string for the sql driver. The yaml defininition for the sink in the e2e test already has the environment variables added but their values need to be changed to match the ones for the database that will be deployed. The test should wait until the database pod is ready before deploying the sink to avoid race conditions

rh-hemartin:
* Deploy KubeArchive before starting the integration tests.
* Introduced a test that checks the deployments ready replicas.

Closes #129

### DIFF
--- a/test/integration/integration_test.go
+++ b/test/integration/integration_test.go
@@ -5,6 +5,7 @@ package main
 import (
 	"context"
 	"fmt"
+	"log"
 	"os"
 	"testing"
 	"time"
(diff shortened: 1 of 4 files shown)


test/integration/integration_test.go
```diff
@@ -21,71 +22,42 @@ func TestMain(m *testing.M) {
...
+			areAllReady = areAllReady && deployment.Status.ReadyReplicas == 1
+		}
+
+		if areAllReady {
+			log.Printf("All deployments ready.")
+			break
 		}
-	})
+
+		log.Printf("Not all deployments are ready, waiting 1 second...")
```
mafh314, line 59: The timeout will be 10 seconds, and if the resources are not ready the test will check the ready state each second, right? To me it's a like a short time. Why did you change the timeout from 30s to 10s? Maybe I'm missing something.

rh-hemartin: The test does not need to wait more and I expect it to pass the first cycle, as the test itself is done by the `rollout status` in the `hack/quick-install.sh` script. However I think you have a point in that it does not resemble a realistic timeout or sleep time. I increased the total to 30 seconds and the sleep time to 5.

mafh314 (approved):

PR #157, opened 2024-07-08
Add endpoint to query non core namespaced resources

### LINKED ISSUE #156: Add endpoint for querying non-core resources by namespace
Add an endpoint in the REST API server to query namespaced non-core resources

maruiz93:
Closes #156

### DIFF
--- a/cmd/api/routers/routers.go
+++ b/cmd/api/routers/routers.go
@@ -19,13 +19,32 @@ type Controller struct {
 func (c *Controller) GetAllResources(context *gin.Context) {
 	group := context.Param("group")
 	version := context.Param("version")
-	// FIXME - Title is deprecated and this conversion must be done in a different way
-	resource := strings.Title(context.Param("resourceType")) //nolint:staticcheck
-	resources, err := c.Database.QueryResources(context, resource[:len(resource)-1], group, version)
+	resource := getResourceKindFromPlural(context.Param("resourceType"))
+	resources, err := c.Database.QueryResources(context, resource, group, version)
 
 	if err != nil {
 		context.JSON(http.StatusInternalServerError, err.Error())
 	}
 
 	context.JSON(http.StatusOK, resources)
+}
+
+func (c *Controller) GetNamespacedResources(context *gin.Context) {
+	group := context.Param("group")
+	version := context.Param("version")
+	namespace := context.Param("namespace")
+	resource := getResourceKindFromPlural(context.Param("resourceType")) + "/" + context.Param("namespace")
+	resources, err := c.Database.QueryNamespacedResources(context, resource, group, version, namespace)
+
+	if err != nil {
+		context.JSON(http.StatusInternalServerError, err.Error())
+	}
+
+	context.JSON(http.StatusOK, resources)
+}
+
+func getResourceKindFromPlural(plural string) string {
+	// FIXME - Title is deprecated and this conversion must be done in a different way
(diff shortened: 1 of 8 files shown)


cmd/api/routers/routers.go
```diff
@@ -28,3 +28,18 @@ func (c *Controller) GetAllResources(context *gin.Context) {
...
 	context.JSON(http.StatusOK, resources)
 }
+
+func (c *Controller) GetNamespacedResources(context *gin.Context) {
+	group := context.Param("group")
+	version := context.Param("version")
+	namespace := context.Param("namespace")
+	// FIXME - Title is deprecated and this conversion must be done in a different way
+	resource := strings.Title(context.Param("resourceType")) //nolint:staticcheck
+	resources, err := c.Database.QueryNamespacedResources(context, resource[:len(resource)-1], group, version, namespace)
```
beerparty, line 38: Is `resource` a resource name? It looks like getting the resource name is going to be tricky for the `Database.QueryNamespacedResources` call. I suggest making a small function like `func GetResourceName(context *gin.Context)` to return the string that `QueryNamespacedResources` needs. That way you don't need to trim the name inside the `QueryNamespacedResources` call and you can create a unit test that verifies you are still getting the right string after you replace the deprecated `Title` code.

maruiz93: Yes, this logic will be updated soon, but I extracted it to a different method.

I will probably add a request to the kubernetes cluster as a middleware function and set the singular resource name in the context, but I think that for now it makes more sense having it in a separated private function.

Already updated so I'll close the thread

beerparty (approved):
Minor typo in a comment, not enough to request changes in this MR. But if something else has to change...

cmd/api/routers/routers.go
```diff
@@ -18,13 +18,32 @@ type Controller struct {
...
+		context.JSON(http.StatusInternalServerError, err.Error())
+	}
+
+	context.JSON(http.StatusOK, resources)
+}
+
+func getResourceKindFromPlural(plural string) string {
+	// FIXME - Title is deprecated and this conversion must bbe done in a different way
+	capitalized := strings.Title(plural) //nolint:staticcheck
+	return capitalized[:len(plural)-1]
```
ggallen, line 48: Are you just chopping of the last character here for now?

And can this be done all in one line?  Is there any advantage to creating `capitalized`?  Could it just become this?
```
return strings.Title(plural)[:len(plural)-1]
```

skoved (approved):

cmd/api/routers/routers.go
maruiz93: I've chosen to do in two steps because of the `//nolint:staticcheck` comment, IIUC, that way the linter is just disabled in the `Title` operation and not in the other one.

ggallen (approved):

cmd/api/routers/routers.go
```diff
@@ -18,13 +18,32 @@ type Controller struct {
...
+
+	if err != nil {
+		context.JSON(http.StatusInternalServerError, err.Error())
+	}
+
+	context.JSON(http.StatusOK, resources)
+}
+
+func getResourceKindFromPlural(plural string) string {
+	// FIXME - Title is deprecated and this conversion must bbe done in a different way
```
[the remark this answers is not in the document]
maruiz93: Fixed

PR #159, opened 2024-07-09
Handle errors on the controller

### LINKED ISSUE #125: resolve issues with golangci-lint errcheck
the function calls in this code block return errors that are not checked

https://github.com/kubearchive/kubearchive/blob/main/cmd/operator/internal/controller/kubearchiveconfig_controller.go#L54

this causes golangci-lint to fail. For now golangci-lint ignores that file. We should fix the code so that we can continue to run golangci-lint on this file

rh-hemartin:
* Add error handling for some functions on the Controller code.
* Skipped a problematic test.

Fixes #125

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -13,6 +13,3 @@ linters:
     - nilerr
     - usestdlibvars
     - wastedassign
-issues:
-  exclude-files:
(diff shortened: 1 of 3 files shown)


rh-hemartin:
See #160  for more information about the skipped test.

beerparty (approved):
LGTM

PR #170, opened 2024-07-10
Sink writes cluster name to datbase and updates entries

skoved:
The sink reads the name of the cluster from an environment variable and writes that value for entries that it addes to the database. The sink will also check if a resource has already been written to the database and update the entry if the resource has already been written. Otherwise it will create a new entry


skoved:
to test:

1. install KubeArchive using the helm chart
2. run `kubectl create ns pod-test`
3. initialize the database
4. run `kubectl apply -f - <<EOF
---
kind: KubeArchiveConfig
metadata:
  name: kubearchive
  namespace: pod-test
spec:
  resources:
    - apiVersion: v1
      kind: Pod
EOF`
5. run `kubectl run -n pod-test busybox --image=busybox --restart=Never -- ls`
6. access the database
7. from psql run `SELECT * FROM resource;`
8. you should see exactly one entry for the pod `busybox` in the namespace `pod-test`

cmd/sink/main.go
```diff
@@ -50,46 +60,75 @@ func NewSink(db *sql.DB, logger *log.Logger) *Sink {
...
 	if err != nil {
 		sink.Logger.Printf("cloudevent %s is malformed and will not be processed: %s\n", event.ID(), err)
 		return
 	}
 	sink.Logger.Printf("cloudevent %s contains all required fields. Attempting to write it to the database\n", event.ID())
+	sink.Lock.Lock()
+	defer sink.Lock.Unlock()
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
+	sink.Logger.Printf("checking if resource from cloudevent %s has been written to the database already\n", event.ID())
+	entryId, err := sink.Db.QueryResourceId(ctx, archiveEntry)
```
ggallen, line 84: You are querying the database to see if the resource already exists before writing it?  This is very inefficient, resulting in twice as many database queries as needed.

You should look at using an `upsert` instead.

See https://www.postgresql.org/docs/current/sql-insert.html.

pkg/database/database.go
```diff
@@ -7,17 +7,24 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
+
```
ggallen, line 10: Why the whitespace change here?

pkg/database/database.go
```diff
@@ -7,17 +7,24 @@ import (
...
 	"github.com/kubearchive/kubearchive/pkg/models"
 	_ "github.com/lib/pq"
 )
 
 const (
-	resourceTableName = "resource"
-	resourcesQuery    = "SELECT data FROM %s WHERE kind=$1 AND api_version=$2"
+	resourceTableName   = "resource"
+	resourcesQuery      = "SELECT data FROM %s WHERE kind=$1 AND api_version=$2"
+	resourceIdQuery     = `SELECT id FROM %s WHERE kind=$1 AND api_version=$2 AND cluster=$3 AND cluster_uid=$4 AND name=$5 AND namespace=$6`
```
ggallen, line 18: If you need this query (and I don't think you do, see my earlier comment) you should just be able to query by the resource UUID, which is guaranteed to be unique.  

In fact,  UUID should be the primary key on the table.

pkg/database/fake/database_generated.go
```diff
@@ -3,15 +3,17 @@ package fake
 import (
 	"context"
 	"fmt"
+
```
ggallen, line 6: Again, why the whitespace change?

pkg/database/database.go
```diff
@@ -7,17 +7,24 @@ import (
...
+	updateResourceQuery = `UPDATE %s SET resource_version=$1, updated_ts=$2, cluster_deleted_ts=$3, data=$4 WHERE id=$5`
```
ggallen, line 19: I think other fields need to be updated as well. What if the name changed? Or the namespace?

ggallen (changes requested):

cmd/sink/main.go
ggallen: Also see https://github.com/kubearchive/kubearchive/issues/172, which may be required for the `upsert`.

pkg/database/database.go
skoved: removed the whitespace

pkg/database/fake/database_generated.go
skoved: removed the whitespace

pkg/database/database.go
skoved: I've added using UUID as the primary key as a project discussion topic

cmd/sink/main.go
skoved: done!

pkg/database/database.go
skoved: uuid is now the primary key and upsert is done using uuid

pkg/database/database.go
skoved: added them

pkg/database/database.go
```diff
@@ -60,3 +66,85 @@ func (db *Database) QueryResources(ctx context.Context, kind, group, version str
...
+		return fmt.Errorf("commit to database failed and the transactions was rolled back: %s", execErr)
+	}
+	return nil
+}
+
+func (db *Database) WriteResource(ctx context.Context, entry *models.ResourceEntry) error {
+	query := fmt.Sprintf(writeResource, db.resourceTableName)
+	tx, err := db.db.BeginTx(ctx, nil)
+	if err != nil {
+		return fmt.Errorf("could not begin transaction: %s", err)
```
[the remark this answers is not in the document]
skoved: the context (ctx) only contains a timeout for the calls to the database but doesn't have any relevant info for the user. I've included the resource uuid in the error message to try to distinguish more about the operation

ggallen (approved):

PR #174, opened 2024-07-11
Use goimports instead of gofmt to format go code

### LINKED ISSUE #173: Switch from gofmt to goimports for enforcing go code style
goimports enforces the same style as gofmt but also will check imports. It removes unused imports, adds missing imports, and enforces ordering of imports. We should run goimports on all go code and update the github action to use goimports instead of gofmt

[more about goimports](https://pkg.go.dev/golang.org/x/tools/cmd/goimports)

skoved:
The github action that runs gofmt now runs goimports instead and all go code in the repo has been passed through goimports to ensure that it has the correct style.

resolves #173


rh-hemartin (approved):

skoved:
I messed up my first attempt at running `goimports` to format all of the go code. After the second attempt all go code have the same style for imports

maruiz93 (approved):

beerparty (approved):

Issue #136, opened 2024-07-03
Get the real plural name from the CRD, rather than blindly appending "s".


ggallen:
The configuration of the role required by `ApiServerSource` need so know the plural name of the resource to be configured correctly. Currently, the code just blindly appends an "s" to the kind.  It works, but it's not the correct solution.

The operator needs to take the API version and kind provided in the `KubeArchiveConfig` and look up the associated resource to retrieve the correct plural name.

This is happening on [this line](https://github.com/kubearchive/kubearchive/blob/main/cmd/operator/internal/controller/kubearchiveconfig_controller.go#L160).

ggallen:
The operator framework provides a client to make kubernetes calls from go, which is what I plan to use when and if I figure it out.

rh-hemartin:
Remember to implement (or add a TODO for implementing) a cache. Also if it comes to it, we can store the singular on the database, as we have that information in the events that the ApiServerSource sends us.

ggallen:
The operator framework takes care of the cache.

PR #175, opened 2024-07-15
Get the real plural name from the CRD, rather than blindly appending 's'.

ggallen:
Resolves https://github.com/kubearchive/kubearchive/issues/136.

### DIFF
--- a/cmd/operator/internal/controller/kubearchiveconfig_controller.go
+++ b/cmd/operator/internal/controller/kubearchiveconfig_controller.go
@@ -13,8 +13,10 @@ import (
 	duckv1 "knative.dev/pkg/apis/duck/v1"
 
 	"k8s.io/apimachinery/pkg/api/errors"
+	meta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	schema "k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -27,6 +29,7 @@ import (
 type KubeArchiveConfigReconciler struct {
 	client.Client
 	Scheme *runtime.Scheme
+	Mapper meta.RESTMapper
 }
 
 //+kubebuilder:rbac:groups=kubearchive.kubearchive.org,resources=kubearchiveconfigs,verbs=get;list;watch;create;update;patch;delete
@@ -135,7 +138,7 @@ func (r *KubeArchiveConfigReconciler) reconcileRole(ctx context.Context, kaconfi
 	log := log.FromContext(ctx)
 
 	log.Info("in reconcileRole")
-	role, err := r.desiredRole(kaconfig)
+	role, err := r.desiredRole(ctx, kaconfig)
 	if err != nil {
 		log.Error(err, "Unable to get desired Role")
 		return role, err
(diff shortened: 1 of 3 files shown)


cmd/operator/internal/controller/kubearchiveconfig_controller.go
```diff
@@ -162,18 +165,26 @@ func (r *KubeArchiveConfigReconciler) reconcileRole(ctx context.Context, kaconfi
...
-		slash := strings.Index(resource.APIVersion, "/")
-		if slash > -1 {
-			apiGroup = resource.APIVersion[:slash]
+		apiVersion := resource.APIVersion
+		data := strings.Split(resource.APIVersion, "/")
+		if len(data) > 1 {
+			apiGroup = data[0]
+			apiVersion = data[1]
+		}
+		gvr, err := r.Mapper.RESTMapping(schema.GroupKind{Group: apiGroup, Kind: resource.Kind}, apiVersion)
```
maruiz93, line 179: I think that we should add a comment to make it clear that gvr resources are expected to be the plural of the resource based on the kubernetes standard to use the plurals in the API endpoints.

Also, I think it's also worth to mention that k8s roles' rules expect the `resources` field to be filled with the plural lowercased version of the resource (as oposed to the singular PascalCase of the `kind` field of the resource)

maruiz93 (changes requested):

cmd/operator/internal/controller/kubearchiveconfig_controller.go
ggallen: Fixed.

skoved (commented):
[The documentation for configuring KubeArchive](https://kubearchive.github.io/kubearchive/main/configuration/kubearchiveconfig.html) Shows a `KubeArchiveConfig` with the `kind` of the resource pluralized. There were a series of issues we were having around getting the `ApiServerSource` to created by the operator to work and in that process we switched to have the resource be singular. Do we want to stick with the singular (in which case the docs need to be updated) or should we switch back to specifying the plural to match the docs?

ggallen:
@skoved, the `kind` should not be pluralized. It should be the real `kind` value.

I've updated the documentation. Nice catch.

maruiz93 (approved):

cmd/operator/internal/controller/kubearchiveconfig_controller.go
```diff
@@ -162,18 +165,28 @@ func (r *KubeArchiveConfigReconciler) reconcileRole(ctx context.Context, kaconfi
...
+		// The resource field in the GVR contains the plural version of the resource, and
+		// the kubernetes Role expects this lower-cased plural version.
+		gvr, err := r.Mapper.RESTMapping(schema.GroupKind{Group: apiGroup, Kind: resource.Kind}, apiVersion)
+		if err == nil {
+			rules = append(rules, rbacv1.PolicyRule{
+				APIGroups: []string{apiGroup},
+				Resources: []string{strings.ToLower(gvr.Resource.Resource)},
+				Verbs:     []string{"get", "list", "watch"}})
+		} else {
+			log.Error(err, "Failed to get GVR for "+resource.APIVersion)
```
mafh314, line 188: WDYT about this: `log.Error("Failed to get GVR for: %s", resource.APIVersion, err)`

skoved: `[log.Error](https://pkg.go.dev/github.com/go-logr/logr#Logger.Error)` does not support format verbs, like `%s` and `logr.Logger` does not appear to have an `Errorf` method

ggallen: This PR has been merged already.

PR #178, opened 2024-07-16
Fix how to get kind from resource name

maruiz93:


maruiz93:
Need to add tests

DEVELOPMENT.md
```diff
@@ -196,5 +196,6 @@ to start a debugger to which attach from your IDE.
    ```
    expose the `KIND_CLUSTER_NAME` env variable with the appropriate name of the kind cluster:
    ```bash
-   export KIND_CLUSTER_NAME=$(kubectl config get-clusters | grep -P '(?<=kind-).*' -o)
+   export KIND_CLUSTER_NAME=$(kind -q get clusters)
```
ggallen, line 199: How is this related to getting the kind from the resource name?

cmd/api/auth/authz.go
```diff
@@ -7,6 +7,8 @@ import (
 	"fmt"
 	"net/http"
 
+	"github.com/kubearchive/kubearchive/cmd/api/abort"
```
ggallen, line 10: How are all these changes in authz.go related to getting the kind from the resource name?

cmd/api/auth/authn.go
```diff
@@ -9,6 +9,8 @@ import (
 	"net/http"
 	"strings"
 
+	"github.com/kubearchive/kubearchive/cmd/api/abort"
```
ggallen, line 12: How are all these changes in authn.go related to getting the kind from the resource name?

cmd/api/routers/routers.go
```diff
@@ -19,7 +23,7 @@ type Controller struct {
 func (c *Controller) GetAllResources(context *gin.Context) {
 	group := context.Param("group")
 	version := context.Param("version")
-	resource := getResourceKindFromPlural(context.Param("resourceType"))
+	resource := getAPIResourceKind(context)
```
ggallen, line 26: Shouldn't this really be called `kind`?

cmd/api/routers/routers.go
```diff
@@ -33,7 +37,7 @@ func (c *Controller) GetNamespacedResources(context *gin.Context) {
 	group := context.Param("group")
 	version := context.Param("version")
 	namespace := context.Param("namespace")
-	resource := getResourceKindFromPlural(context.Param("resourceType")) + "/" + context.Param("namespace")
+	resource := getAPIResourceKind(context)
```
ggallen, line 40: Shouldn't this really be called `kind`?

ggallen (changes requested):

DEVELOPMENT.md
maruiz93: It is not, I just added the change because I faced problems when testing in my local with the other command.

It is ok for you to place it here or should I create separate commit in the PR? or a separate PR for this line?

I thought that with a separate PR it will be a little overkill as I guess I should create an issue for it.

cmd/api/auth/authn.go
maruiz93: I created a new go module for the middleware layer in charge of doing the request of the api resources called ` discover`. This module needs the `Abort` function so I exported it outside from the `auth` module to its own one called `abort`.
Exporting the `abort` function and moving it to its own module created the need of importing it in the `auth` module to keep things working.

cmd/api/auth/authz.go
maruiz93: See answer in the previous thread

skoved (approved):

cmd/api/routers/routers.go
maruiz93: Fixed

cmd/api/routers/routers.go
ggallen: No, I meant like this:
```
kind := getAPIResourceKind(context)
```

DEVELOPMENT.md
ggallen: I'm OK with it, but I've seen on other projects that people complain when things are mixed.  I think this is mainly for reverting.  

If you do make additional changes, I tend to just note them in the description so that the reviewer is aware.

cmd/api/auth/authn.go
ggallen: Understood.  I missed that link.

cmd/api/routers/routers.go
ggallen: No, I meant like this:
```
kind := getAPIResourceKind(context)
```

cmd/api/routers/routers.go
maruiz93: yes, that's the change that I did, don't you see it updated in the `Files Changed` section of this PR?
Note that this thread keeps the outdated code (there is the `outdated` label on it)

cmd/api/routers/routers.go
maruiz93: see above comment, I think you aren't checking the most updated changes

ggallen (approved):

PR #185, opened 2024-07-18
Deleted skipped operator tests

### LINKED ISSUE #160: Delete the Operator tests
Currently both test suites for the operator are skipped (`kubearchive_controller_test.go` and `e2e_test.go`). Both are integration tests, as they need infrastructure to run.

* `e2e_test.go` depends on having a `Makefile`, which we don't so we can discard it.
* `kubearchive_controller_test.go` is an integration test and needs proper setup, which we already do for `./test/integration/integration_test.go`

I propose that we:

* Delete the `e2e_test.go` as they just install the operator and check that its running.
* Delete the `kubearchive_controller_test.go` tests and implement a similar test in `./test/integration/integration_test.go` (covered by https://github.com/kubearchive/kubearchive/issues/127). We can repurpose the file for unit testing.

rh-hemartin:
Fixes #160


ggallen:
I'd rather keep these tests and delete them when the replacement is ready.

> Delete the e2e_test.go as they just install the operator and check that its running.

Why is this not valuable?  If there's a Makefile dependency we an just remove that and it will work. 

I was planning to get these tests working. I'm about to that point now to get started.

PR #204, opened 2024-07-19
Consistent GitHub action names, debug, and versions

### LINKED ISSUE #203: Consistent GitHub action names, debug, and version
The GitHub actions should be updated so they have consistent names and debug output when running. There is also an inconsistent version for "setup-go@v5", all the actions should be using "stable".

beerparty:
Updates the GitHub actions so they have consistent names and debug output when running. Also fixes an inconsistent version for "setup-go@v5", all the actions have been updated to use go-version "stable".

Fixes #203

```release-note
Updates GitHub actions for consistent action names, debug, and versions.
```

kind/documentation

### DIFF
--- a/.github/workflows/documentation-build.yaml
+++ b/.github/workflows/documentation-build.yaml
@@ -1,5 +1,6 @@
 ---
-name: Documentation build
+name: Documentation Build
+run-name: Documentation Build
 on:
   pull_request:
     branches: [main]
@@ -8,6 +9,10 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
+      - run: |
+          echo "Job triggered by ${{ github.actor }}."
+          echo "Running on a ${{ runner.os }} server hosted by GitHub."
+          echo "Branch ${{ github.ref }} of repository ${{ github.repository }}."
       - name: Checkout repository
         uses: actions/checkout@v4
       - name: Install Node.js
--- a/.github/workflows/documentation-release.yaml
+++ b/.github/workflows/documentation-release.yaml
@@ -1,5 +1,6 @@
 ---
-name: Documentation release
+name: Documentation Release
+run-name: Documentation Release
 on:
   push:
     branches: [main]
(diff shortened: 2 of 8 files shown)


rh-hemartin (approved):

.github/workflows/documentation-build.yaml
```diff
@@ -8,6 +9,10 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
+      - run: |
+          echo "Job triggered by ${{ github.actor }}."
+          echo "Running on a ${{ runner.os }} server hosted by GitHub."
+          echo "Branch ${{ github.ref }} of repository ${{ github.repository }}."
```
maruiz93, line 15: I see that we are including this descriptive first step in all the workflows, so I was wondering if there is any mechanism provided by Github Actions to reuse part of the workflows following the DRY principle.

I've found that there are actually 3 ways of doing that, described in [this post](https://cardinalby.github.io/blog/post/github-actions/dry-reusing-code-in-github-actions/)

Could we include something like that to prevent having this 3 lines of `echo` rewritten in every workflow?

maruiz93 (changes requested):

maruiz93:
DCO (Developer Certificate Origin) is failing as part of the CI of this PR. I believe the commit need to be signed to fix that.

.github/workflows/documentation-build.yaml
rh-hemartin: Because none of the methods to not repeat look complex, I think we can take a step back and ask ourselves if this information is useful. I think its not, here my reasoning:

* `github.actor` is already present on the list of actions.
* `runner.os` is already printed on the very first "embedded" step called "Set up job".
* `github.ref` is already printed on the checkout step.
* `github.repository` is already also printed on the checkout step.

So maybe deleting is better than finding a way to DRY.

maruiz93: Removing the step like @rh-hemartin is suggesting makes a lot of sense for me

beerparty: I am going to create a separate task to remove the duplication. Including the removal of duplicate code here will require PR testing, which I don't want to do with a push request already in progress.

maruiz93 (approved):

PR #205, opened 2024-07-22
Apply Helm naming conventions to general templates

### LINKED ISSUE #105: Adjust general templates in Helm Chart to the defined naming conventions

```yaml
# values
operator: 

# template

# output
kubearchive-operator
```

## Goal

Review and update the Helm Chart templates related to the kubearchive general templates (no subfolder) to comply with the naming conventions.

rh-hemartin:

Fixes #105 

## Release Note

```release-note
NONE
```

### DIFF
--- a/charts/kubearchive/templates/certificates.yaml
+++ b/charts/kubearchive/templates/certificates.yaml
@@ -1,9 +1,10 @@
 # yamllint disable rule:braces
+{{- $caName := tpl "{{ .Release.Name }}-ca" . -}}
 ---
 apiVersion: cert-manager.io/v1
 kind: Issuer
 metadata:
-  name: kubearchive-ca-issuer
+  name: "{{ .Release.Name }}-ca"
 spec:
   selfSigned: {}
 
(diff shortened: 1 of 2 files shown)


charts/kubearchive/templates/certificates.yaml
```diff
@@ -3,32 +3,32 @@
...
 ---
 apiVersion: cert-manager.io/v1
 kind: Issuer
 metadata:
-  name: kubearchive-cert-issuer
+  name: "{{ .Release.Name }}"
 spec:
   ca:
-    secretName: kubearchive-ca-secret
+    secretName: "{{ .Release.Name }}-ca"
```
maruiz93, line 34: `{{ .Release.Name }}-ca` appears 5 times here. Based on that and on how we refactored the other helm templates I think we should include a variable in the values with that template and use `tpl` to reference it.

Something like:

values.yaml
```yaml
kubearchive:
  ca: "{{ .Release.Name }}-ca"
...
```

certificates.yaml
```yaml
...
spec:
  ca:
    secretName: "{{ tpl .Values.kubearchive.ca }}"
...
```

maruiz93 (changes requested):

charts/kubearchive/templates/certificates.yaml
rh-hemartin: Instead of creating a new variable in the values, I created the variable at the top of the file with an assignment. This is not a variable for the user to modify.

maruiz93 (approved):

Issue #171, opened 2024-07-10
`created_ts` and `updated_ts` are automatic


ggallen:
I still don't like the field names for timestamps, particularly `cluster_deleted_ts`.  I don't know what makes this field so different that it has to have "cluster" in it. I see this in the documentation:
```
created_ts: Time when KubeArchive received the resource for the first time.
updated_ts: Last time KubeArchive received an update for the resource.
cluster_deleted_ts: Timestamp when KubeArchive deleted the resource from the cluster.
```
Why not just `deleted_ts` again? Shouldn't this just be the "Time when KubeArchive received the delete event"? Or "Time when KubeArchive submitted the deletion of the resource"?

maruiz93:
I don't think we should name the field `deleted_ts`.

It is quite a standardize thing to include in the DB tables the fields `created_at` and `updated_at` (that relates to our `created_ts` and `updated_ts` fields). 
The meaning of these fields are:

* created_at --> When the record was first inserted in the table
* updated_at --> Last time the record is updated in the table

It is quite common to include these fields in all the tables of a database, because they are metadata of the record per se, and not related to the business logic.

There is no `deleted_at` because following the logic of the other timestamps, that would mean "When the record is deleted from the table" and that makes no sense if the record is still there.

In our case we want to have:

*cluster_deleted_ts --> When kubearchive prune the resource from etcd

If you don't like `cluster_deleted_ts` we can change it to a different one, maybe `etcd_deleted_ts` or `etcd_prune_ts` but I will encourage not to use `deleted_ts` because that will be confusing giving the standardize usage of `created_at` and `updated_at` standardize usage as metadata for the record in the tables.

maruiz93:
Other things to consider are:

* In the future we may want to address the actual removal of old records in the kubearchive database. We should be carefull of the word "delete" either providing context of "what we delete where" or by using different words like prune/delete, being prune the removal of resources from etcd to get it cleaned and delete the complete removal of the resources

* I've also seen triggers defined in the database like [these ones](https://medium.com/@alexishevia/automatic-timestamps-with-postgresql-8a5fc499d887) to automatically fill the `updated_at` and `created_at` fields in the DB and not having to set them manually. As I see it, this also helps stating the different nature of the fields (`updated_ts` `created_ts`) and `cluster_deleted_ts` because the last one can't be automated based what happens with the record in relation with the table.

skoved:
[...]

PR #210, opened 2024-07-22
Automate created_at and updated_at DB fields

maruiz93:

Related to #171 

## Release Note

```release-note
NONE
```

## Notes for Reviewers

This PR covers:
* Renaming `updated_ts` and `created_ts` to `updated_at` and `created_at` to comply with the standard name of these fields
* Set these fields with a default value of `now()` so the writer don't need to bother about them
* Create a trigger linked to the `resources` table so the `updated_at` field gets automatically updated

I followed [this guide](https://medium.com/@alexishevia/automatic-timestamps-with-postgresql-8a5fc499d887)

The sink (writer of the DB) need to add some changes to remove those fields from the INSERT statements as they are going to be automatically set. 
This is why I am merging this changes against `db_triggers` branch and not `created_updated_ts`. Once we have both PR we will be able to marge this with main.


rh-hemartin:
The destination branch does not look correct, please check if it is and fix it if needed.

rh-hemartin:
The destination branch does not look correct (as its not main), please check if it is and fix it if needed.

rh-hemartin (approved):

skoved (approved):

PR #211, opened 2024-07-22
Add required helm kw to api templates

### LINKED ISSUE #91: Add required function to API REST templates in Helm Chart
## Required function in Helm Templates
The optional fields in the kubernetes resources that we expect to be filled through the `helm template` command should include this function to detect typos or similar errors in the CI.

More information on the function in [here](https://helm.sh/docs/howto/charts_tips_and_tricks/#using-the-required-function)

## Goal
Include this keyword in the templates of the kubearchive REST API component.

maruiz93:

Fixes #91 

## Release Note

```release-note
NONE
```

## Notes for Reviewers

This PR is the first of several that we should create if we agree to go on with the usage of the [required](https://helm.sh/docs/howto/charts_tips_and_tricks/#using-the-required-function) helm keyword.

The main goal is to prevent introducing typos in the templated variables of the templates in the charts.

The required should be introduced in non-mandatory fields (AFAIK, the only field that files an error with `helm lint --strict` if it's not provided is `.metadata.name`) that don't use `tpl` function. This way we will prevent having empty values after `helm template` by mistake.

With the usage of `tpl` the `required` function is not needed because if the variable to be templated not exist, it files an error.

### DIFF
--- a/charts/kubearchive/templates/api_server/api_server.yaml
+++ b/charts/kubearchive/templates/api_server/api_server.yaml
@@ -19,7 +19,7 @@ spec:
             secretName: {{ tpl .Values.apiServer.secret . }}
       containers:
         - name: {{ tpl .Values.apiServer.name . }}
-          image: {{ .Values.apiServer.image }}
+          image: {{ required "An API Server container image is required" .Values.apiServer.image }}
           volumeMounts:
             - name: tls-secret
               readOnly: true
@@ -47,7 +47,7 @@ spec:
             - name: POSTGRES_URL
               value: {{ tpl .Values.database.url . }}
             - name: POSTGRES_PORT
-              value: "{{ .Values.database.service.port }}"
+              value: "{{ required "A Postgres port must be specified" .Values.database.service.port }}"
 {{ include "kubearchive.v1.otel.env" .Values.apiServer | indent 12 }}
 ---
 kind: Service
@@ -59,8 +59,8 @@ spec:
     app: {{ tpl .Values.apiServer.name . }}
   ports:
     - protocol: TCP
-      port: {{ .Values.apiServer.port }}
-      targetPort: {{ .Values.apiServer.port }}
+      port: {{ required "An API Server port must be specified" .Values.apiServer.port }}
+      targetPort: {{ required "An API Server targetPort must be specified" .Values.apiServer.port }}
       name: server
     {{- if .Values.apiServer.debug }}
     - protocol: TCP
--- a/charts/kubearchive/templates/api_server/certificates.yaml
+++ b/charts/kubearchive/templates/api_server/certificates.yaml
@@ -11,7 +11,7 @@ spec:
   renewBefore: 360h  # 15 days
   subject:
     organizations:
-      - {{ .Release.Name }}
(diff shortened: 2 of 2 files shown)


skoved (approved):

rh-hemartin:
I don't like having this scattered in the different templates, I'll leave a piece for discussion in the ticket.

maruiz93:
I have nothing against the schema but, I think it does not cover the same functionality.

With the `required` function I am preventing that if I make a typo on the name of the variable either in the `values.yaml` or in a template where I reference a variable name, it will fail on CI. 

With the schema, we make sure that several values for variables are provided, that covers having typos or lacking a value for a variable in the `values.yaml`, but what if I make a typo referencing the variable in the template? It will be rendered as an empty value.

The `required` function is the better I found for covering that, but maybe there are alternatives. For me, the best would be an option in the `helm lint` command that ensures that every variable referenced in a template has to be defined, but  the more similar thing I found for that is the `required` function.

rh-hemartin (approved):

charts/kubearchive/templates/api_server/certificates.yaml
```diff
@@ -11,7 +11,7 @@ spec:
   renewBefore: 360h  # 15 days
   subject:
     organizations:
-      - {{ .Release.Name }}
+      - {{ required "An organization in the api server certificate must be specified" .Release.Name }}
```
mafh314, line 14: `api server -> `API Server` . To maintain consistent.

maruiz93: Fixed!

mafh314 (approved):

charts/kubearchive/templates/api_server/api_server.yaml
```diff
@@ -47,7 +47,7 @@ spec:
             - name: POSTGRES_URL
               value: {{ tpl .Values.database.url . }}
             - name: POSTGRES_PORT
-              value: "{{ .Values.database.service.port }}"
+              value: "{{ required "A postgres port must be specified" .Values.database.service.port }}"
```
beerparty, line 50: Since you capitalize "API Server", you should capitalize "Postgres" as well:

```suggestion
              value: "{{ required "A Postgres port must be specified" .Values.database.service.port }}"
```

charts/kubearchive/templates/api_server/api_server.yaml
```diff
@@ -47,7 +47,7 @@ spec:
             - name: POSTGRES_URL
               value: {{ tpl .Values.database.url . }}
```
beerparty, line 48: Is there a default URL for Postgres? If not, since you are requiring a Postgres port you will need to require the URL as well.

maruiz93: Note that we use `tpl` in there, using `tpl` implies using `required`. If `.Values.database.url` isn't provided, the `helm lint` and `helm template` commands will fail.

ggallen (approved):

charts/kubearchive/templates/api_server/api_server.yaml
maruiz93: Fix applied

PR #214, opened 2024-07-22
Add Jaeger deployment

### LINKED ISSUE #213: Add a telemetry endpoint to send traces
While stdout is fine for seeing if traces are being sent or not, parent-child relationships are slower to track in stdout. We need to include an endpoint for telemetry that we deploy for development purposes.

rh-hemartin:

Fixes #213 

## Release Note

```release-note
NONE
```

## Notes for Reviewers

* This is still a Draft, looking to share the code but a lot of things need to be solved first. See the issue for more information.

### DIFF
--- a/charts/kubearchive/templates/_helpers.tpl
+++ b/charts/kubearchive/templates/_helpers.tpl
@@ -3,17 +3,19 @@ Copyright KubeArchive Authors
 SPDX-License-Identifier: Apache-2.0
 */}}
 
-{{/*
-Create environment variables for OpenTelemetry if .observability is set to true. Otherwise set KUBEARCHIVE_OTEL_ENABLED=false.
-This tells the OpenTelemtry instrumentation if it should start or not.
-*/}}
+
+{{/* Create environment variables for OpenTelemetry */}}
 {{- define "kubearchive.v1.otel.env" -}}
-{{- if .observability -}}
-- name: KUBEARCHIVE_OTEL_ENABLED
-  value: "true"
-{{- else -}}
-- name: KUBEARCHIVE_OTEL_ENABLED
-  value: "false"
-{{- end -}}
+{{- $enabled := .Values.integrations.observability.enabled -}}
+{{- $endpoint := .Values.integrations.observability.endpoint -}}
+
+{{/* If 'enabled' we deploy a traces collector, so override the endpoint to use it */}}
+{{- if (eq $enabled true) -}}
+    {{- $endpoint = tpl "traces-collector.{{ .Release.Namespace }}.svc.cluster.local:4318" . -}}
 {{- end -}}
 
+- name: KUBEARCHIVE_OTEL_ENABLED
+  value: '{{ ternary "false" "true" (eq $endpoint "") }}'
+- name: OTEL_EXPORTER_OTLP_ENDPOINT
+  value: "{{ $endpoint }}"
+{{- end -}}
--- a/charts/kubearchive/templates/api_server/api_server.yaml
+++ b/charts/kubearchive/templates/api_server/api_server.yaml
@@ -48,7 +48,7 @@ spec:
               value: {{ tpl .Values.database.url . }}
             - name: POSTGRES_PORT
               value: "{{ required "A Postgres port must be specified" .Values.database.service.port }}"
-{{ include "kubearchive.v1.otel.env" .Values.apiServer | indent 12 }}
+            {{- include "kubearchive.v1.otel.env" . | nindent 12 }}
 ---
 kind: Service
 apiVersion: v1
--- a/charts/kubearchive/values.yaml
+++ b/charts/kubearchive/values.yaml
@@ -24,8 +24,6 @@ apiServer:
   # KO_DEFAULTBASEIMAGE=gcr.io/k8s-skaffold/skaffold-debug-support/go:latest ko build --disable-optimizations \
   # github.com/kubearchive/kubearchive/cmd/api
   debug: false
-  # If true, OpenTelemetry instrumentation will be enabled for the api server
-  observability: false
   # NOTE - This resource must include the certificate suffix to work
   cert: "{{ tpl .Values.apiServer.name . }}-certificate"
   secret: "{{ tpl .Values.apiServer.name . }}-tls"
@@ -46,9 +44,6 @@ sink:
   targetPort: 8080
   # number of kubearchive sink pods that should be deployed
   replicas: 1
-  # If true, OpenTelemetry instrumentation will be enabled for the sink
-  observability: false
-  # a json path that if it returns a value, will cause the sink to delete the resource
   deleteWhen: ".status.completionTime"
 
 operator:
@@ -93,3 +88,14 @@ database:
     dbUser: "ps_user"
     # database password
     dbPassword: "P0stgr3sdbP@ssword"  # notsecret
+
+integrations:
+  observability:
+    # If "true" a Jaeger all-in-one Deployment is created in the '{{ .Release.Namespace }}' namespace
+    enabled: false
+    # Jaeger all-in-one image tag used for the Deploy if 'enabled' is true
+    imageTag: 1.57.0
+    # Value to which OTEL_EXPORTER_OTLP_ENDPOINT is set. If it is set KubeArchive tries to send telemetry data
+    # If you set `enabled` to true, this is set to the deployed Jaeger all-in-one instance
+    # see `_helpers.tpl` for more information.
+    endpoint: ""
--- a/pkg/observability/trace.go
+++ b/pkg/observability/trace.go
@@ -10,7 +10,7 @@ import (
 	"strings"
 
 	"go.opentelemetry.io/otel"
-	"go.opentelemetry.io/otel/exporters/stdout/stdouttrace"
+	"go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracehttp"
 	"go.opentelemetry.io/otel/propagation"
 	"go.opentelemetry.io/otel/sdk/resource"
 	sdkTrace "go.opentelemetry.io/otel/sdk/trace"
@@ -28,7 +28,7 @@ func Start() error {
 		return nil
 	}
 
-	exporter, err := stdouttrace.New(stdouttrace.WithPrettyPrint())
+	exporter, err := otlptracehttp.New(context.Background())
 	if err != nil {
 		return err
 	}
--- a/charts/kubearchive/templates/sink/sink.yaml
+++ b/charts/kubearchive/templates/sink/sink.yaml
@@ -38,7 +38,7 @@ spec:
               value: "{{ .Values.database.service.port }}"
             - name: DELETE_WHEN
               value: "{{ .Values.sink.deleteWhen }}"
-{{ include "kubearchive.v1.otel.env" .Values.sink | indent 12 }}
+            {{- include "kubearchive.v1.otel.env" . | nindent 12 }}
 ---
 kind: Service
 apiVersion: v1
--- a/charts/kubearchive/templates/traces-collector.yaml
+++ b/charts/kubearchive/templates/traces-collector.yaml
@@ -1,0 +1,77 @@
+{{- if .Values.integrations.observability.enabled }}
+{{ $appName := "traces-collector" }}
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: "{{ $appName }}"
+  labels:
+    app: "{{ $appName }}"
+spec:
+  selector:
+    matchLabels:
+      app: "{{ $appName }}"
+  template:
+    metadata:
+      labels:
+        app: "{{ $appName }}"
+    spec:
+      containers:
+      - env:
+        - name: SPAN_STORAGE_TYPE
+          value: memory
+        - name: JAEGER_DISABLED
+          value: "false"
+        - name: COLLECTOR_OTLP_ENABLED
+          value: "true"
+        image: "jaegertracing/all-in-one:{{ .Values.integrations.observability.imageTag }}"
+        imagePullPolicy: IfNotPresent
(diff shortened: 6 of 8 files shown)


hack/quick-install.sh
```diff
@@ -5,12 +5,29 @@ set -o xtrace
...
 kubectl apply -f https://github.com/knative/eventing/releases/download/knative-${KNATIVE_EVENTING_VERSION}/eventing-core.yaml
 kubectl rollout status deployment --namespace=cert-manager --timeout=30s
 kubectl rollout status deployment --namespace=knative-eventing --timeout=30s
 
+# Development dependencies
+# Avoid problems on namespace creation when re-executing this file
+kubectl create namespace observability --dry-run=client -o yaml | kubectl apply -f -
+kubectl apply -f https://github.com/jaegertracing/jaeger-operator/releases/download/${JAEGER_OPERATOR_VERSION}/jaeger-operator.yaml -n observability
+kubectl rollout status deployment --namespace=observability --timeout=30s
+sleep 5  # The jaeger operator webhook its not fast enough :(
```
beerparty, line 21: Since this is a script that developers will likely run and then wait for it to finish, when you have a pause like this it is a good idea to put something on the console so the developers know why the script has "stopped". 

```suggestion
echo "Wait a few seconds for the Jaeger Operator Webhook to finish initialization."
sleep 5
```

hack/quick-install.sh
```diff
@@ -5,12 +5,29 @@ set -o xtrace
 
 export CERT_MANAGER_VERSION=v1.9.1
 export KNATIVE_EVENTING_VERSION=v1.14.3
+export JAEGER_OPERATOR_VERSION=v1.57.0
 
+# Production dependencies
```
beerparty, line 10: Rather then just comment the code, add a log message so people running this script also know what is going on.

```suggestion
echo "Installing production dependencies"
```

hack/quick-install.sh
```diff
@@ -5,12 +5,29 @@ set -o xtrace
...
 kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/${CERT_MANAGER_VERSION}/cert-manager.yaml
 kubectl apply -f https://github.com/knative/eventing/releases/download/knative-${KNATIVE_EVENTING_VERSION}/eventing-core.yaml
 kubectl rollout status deployment --namespace=cert-manager --timeout=30s
 kubectl rollout status deployment --namespace=knative-eventing --timeout=30s
 
+# Development dependencies
```
beerparty, line 16: Rather then just comment the code, add a log message so people running this script also know what is going on.

```suggestion
echo "Installing development dependencies"
```

hack/quick-install.sh
```diff
@@ -5,12 +5,29 @@ set -o xtrace
...
+apiVersion: jaegertracing.io/v1
+kind: Jaeger
+metadata:
+  name: jaeger
+  namespace: observability
+EOF
+
 bash cmd/operator/generate.sh
 
 helm install kubearchive charts/kubearchive --create-namespace -n kubearchive \
```
beerparty, line 33: This block could use  a logging message as well.

```suggestion
echo "Installing KubeArchive"
helm install kubearchive charts/kubearchive --create-namespace -n kubearchive \
```

beerparty (commented):
Some suggestions for the quick install script.

ggallen (commented):
How is Konflux implementing this? Are they use jaeger?

Is this development only? If so, what's the plan for production and will it be different?

hack/quick-install.sh
rh-hemartin: Sorry, I changed the approach. I was not convinced of adding it on the script, so I deleted the whole installation of the jaeger operator in favor of a deployment with the all-in-one image.

hack/quick-install.sh
rh-hemartin: While theoretically this make sense, practically this gets lost in the output of the script so its not useful. Also, we don't want to redirect the output to somewhere so its more clear if something goes wrong.

hack/quick-install.sh
rh-hemartin: Sorry, I changed the approach. I was not convinced of adding it on the script, so I deleted the whole installation of the jaeger operator in favor of a deployment with the all-in-one image.

hack/quick-install.sh
rh-hemartin: While theoretically this make sense, practically this gets lost in the output of the script so its not useful. Also, we don't want to redirect the output to somewhere so its more clear if something goes wrong.

rh-hemartin:
> How is Konflux implementing this? Are they use jaeger?

As far as I know they are sending traces to the RedHat HoneyComb instance for staging and production.

> Is this development only? If so, what's the plan for production and will it be different?
This is intended for development only. For production you may set the endpoint you want, as long as is compatible with OTLP.

rh-hemartin:
I think I'm good with the current PR code, so I'm opening it for a more formal review.

charts/kubearchive/templates/api_server/api_server.yaml
```diff
@@ -48,7 +48,10 @@ spec:
               value: {{ tpl .Values.database.url . }}
             - name: POSTGRES_PORT
               value: "{{ .Values.database.service.port }}"
-{{ include "kubearchive.v1.otel.env" .Values.apiServer | indent 12 }}
+            - name: KUBEARCHIVE_OTEL_ENABLED
+              value: '{{ ternary "false" "true" (eq .Values.observability.endpoint "") }}'
+            - name: OTEL_EXPORTER_OTLP_ENDPOINT
+              value: "{{ tpl .Values.observability.endpoint . }}"
```
skoved, line 54: Since the same think added here is also being added to the sink, i think that it makes sense to have a template for this (like was defined in `_helpers.tpl`

pkg/observability/trace.go
```diff
@@ -28,7 +28,7 @@ func Start() error {
 		return nil
 	}
 
-	exporter, err := stdouttrace.New(stdouttrace.WithPrettyPrint())
+	exporter, err := otlptracehttp.New(context.TODO())
```
skoved, line 31: I little nitpick-y but i think this should be `context.Background()` since I don't _think_ the intention is to eventually have `Start` function here accept a context

```suggestion
	exporter, err := otlptracehttp.New(context.Background())
```

charts/kubearchive/templates/api_server/api_server.yaml
rh-hemartin: Added the include back.

charts/kubearchive/templates/jaeger.yaml
```diff
@@ -0,0 +1,91 @@
...
+spec:
+  type: ClusterIP
+  selector:
+    app: jaeger
+  ports:
+  - name: http-query
+    port: 16686
+    protocol: TCP
+    targetPort: 16686
+{{- end }}
```
maruiz93, line 91: I would like to suggest several changes related to naming:

* Instead of `jaeger`, I would name the resources and the file with a more generic name like `trace-collector` and maybe extract it into a variable.

* I think that the image should be extracted as a variable in `values.yaml` we potentially would like to change the full image or at least the version.

charts/kubearchive/templates/jaeger.yaml
```diff
@@ -0,0 +1,91 @@
...
+    protocol: TCP
+    targetPort: 4318
+---
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: jaeger
+  name: jaeger-query
+  namespace: {{ .Release.Namespace }}
```
maruiz93, line 81: This is redundant as if it's not provided it will be deployed in this namespace.
```suggestion
```

charts/kubearchive/templates/jaeger.yaml
```diff
@@ -0,0 +1,91 @@
...
+          successThreshold: 1
+          timeoutSeconds: 1
+---
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: jaeger
+  name: jaeger-collector
+  namespace: {{ .Release.Namespace }}
```
maruiz93, line 64: This is redundant as if it's not provided it will be deployed in the same namespace.
```suggestion
```

maruiz93 (changes requested):

charts/kubearchive/templates/jaeger.yaml
rh-hemartin: I used `traces-collector` as name, and exported the image tag for the image.

charts/kubearchive/templates/jaeger.yaml
rh-hemartin: Removed

charts/kubearchive/templates/jaeger.yaml
rh-hemartin: Removed

charts/kubearchive/templates/jaeger.yaml
maruiz93: Then I think we should also rename the name of the file from `jaeger.yaml` to `traces-collector.yaml`

charts/kubearchive/values.yaml
```diff
@@ -91,3 +87,14 @@ database:
...
     # database password
     dbPassword: "P0stgr3sdbP@ssword"  # notsecret
+
+observability:
+  # If "true" a Jaeger all-in-one Deployment is created in the '{{ .Release.Namespace }}' namespace
+  enabled: false
+  # Jaeger all-in-one image tag used for the Deploy if 'enabled' is true
+  imageTag: 1.57.0
+  # Value to which OTEL_EXPORTER_OTLP_ENDPOINT is set. If set KubeArchive tries to send telemetry data
+  # If you set `enabled` to true, use
```
ggallen, line 97: Why can't you just do this conditionally if that value is set?  I'd rather just set enabled to `true` and have it work, rather than having to read the rest of the comments to see that I have to do additional stuff.

ggallen (changes requested):
I would like jaeger turned off (and not installed) by default in development. There should be an easy way to do this.  Most times I don't think developers will want or need this, and it's just something else that can go wrong.

It looks like this can be done, and is the default. I just want to confirm it.

charts/kubearchive/values.yaml
rh-hemartin: I changed it so if you set enabled: true whatever is on the endpoint is replaced by the endpoint of the jaeger Deployment. This way it will be easier to use in development. If enabled is false the value of endpoint is used to populate the variable.

rh-hemartin:
> I would like jaeger turned off (and not installed) by default in development. There should be an easy way to do this. Most times I don't think developers will want or need this, and it's just something else that can go wrong.
> 
> It looks like this can be done, and is the default. I just want to confirm it.

Confirmed, its disabled by default so nothing is deployed. If you need it, set enabled to true.

charts/kubearchive/templates/jaeger.yaml
rh-hemartin: Done as well.

maruiz93 (approved):

charts/kubearchive/templates/_helpers.tpl
```diff
@@ -3,17 +3,19 @@ Copyright KubeArchive Authors
...
-  value: "true"
-{{- else -}}
-- name: KUBEARCHIVE_OTEL_ENABLED
-  value: "false"
-{{- end -}}
+{{- $enabled := .Values.observability.enabled -}}
+{{- $endpoint := .Values.observability.endpoint -}}
+
+{{/* If 'enabled' we deploy a traces collector, so override the endpoint to use it */}}
+{{- if and (eq $enabled true) -}}
```
beerparty, line 13: Why is this "if and"? Shouldn't it just be `if (eq $enabled true)`?

rh-hemartin: Good catch! You are right. It was a leftover from the changes, I removed the `and`.

beerparty (approved):

rh-hemartin:
@ggallen could you review this again? I made some changes due your comments.

mafh314 (approved):

charts/kubearchive/values.yaml
```diff
@@ -93,3 +88,13 @@ database:
     dbUser: "ps_user"
     # database password
     dbPassword: "P0stgr3sdbP@ssword"  # notsecret
+
+observability:
```
ggallen, line 92: I was just thinking about this myself.  I think we should create an `integrations` key and put `observability` under that.  That would also allow it to be used to contain `splunk` or any other integration we support.

charts/kubearchive/values.yaml
```diff
@@ -93,3 +88,13 @@ database:
...
+  # If "true" a Jaeger all-in-one Deployment is created in the '{{ .Release.Namespace }}' namespace
+  enabled: false
```
ggallen, line 94: If it's enabled here, then shouldn't the `enabled` keys on the individual components be removed as well?  I don't see the point in having them in both places, and I think it should be on or off for all of k9e, not on a component-by-component basis.

rh-hemartin: If I understood you correctly, they are already removed. See https://github.com/kubearchive/kubearchive/pull/214/files#diff-860af8aee46eeef818486b58498628115cc3a9427545ea9e0f32b90d4b7e5089L50 and https://github.com/kubearchive/kubearchive/pull/214/files#diff-860af8aee46eeef818486b58498628115cc3a9427545ea9e0f32b90d4b7e5089L28

charts/kubearchive/values.yaml
rh-hemartin: I think it makes sense, we can always go back if we find something wrong with the nesting.

rh-hemartin: Introduced an `integrations` key.

ggallen (approved):

PR #216, opened 2024-07-22
Apply naming conventions for operator in chart

### LINKED ISSUE #103: Adjust operator templates in Helm Chart to the defined naming conventions

```yaml
# values
operator: 

# template

# output
kubearchive-operator
```

## Goal

Review and update the Helm Chart templates related to the kubearchive operator component to comply with the naming conventions.

maruiz93:

Fixes #103 

## Release Note

```release-note
NONE
```

## Notes for Reviewers

Applied the naming conventions described in the linked issue.

Note the following:
* Some resource names have changed to remove redundant suffix
* The label `app.kubernetes.io/name` has been changed to `{{ .Release.Name }}. Based on [the docs](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels) I believe this value makes more sense than the kind of the resource.
* As the values under the `kubearchive` section weren't used anymore I removed them

There is a line out of the scope of the issue (but related). I removed the namespace of the sink service as it should have been removed in [naming conventions PR for the sink](https://github.com/kubearchive/kubearchive/pull/135).

### DIFF
--- a/charts/kubearchive/values.yaml
+++ b/charts/kubearchive/values.yaml
@@ -3,15 +3,6 @@
 # Declare variables to be passed into your templates.
 
 ---
-# template values relevant to all of kubearchive
-kubearchive:
-  # the name of the namespace to deploy kubearchive into
-  namespace: "kubearchive"
-  # RoleBinding name
-  roleBinding: "kubearchive"
-  # Name of Service Account
-  serviceAccount: "kubearchive"
-
 # values used to create the api-server
 apiServer:
   name: "{{ .Release.Name }}-api-server"
@@ -47,7 +38,9 @@ sink:
   # the path that the sink will mount the KubeArchive ConfigMap
   mountPath: "/data/sink-filters"
 
+# values used to create the operator
 operator:
+  name: "{{ .Release.Name }}-operator"
   # NOTE - Helm does not resolve the `ko` path. The path needs to be set on the `helm install` command.
   # For example:
   # helm install -n default <chart-name> charts/kubearchive \
--- a/cmd/operator/generate.sh
+++ b/cmd/operator/generate.sh
@@ -18,10 +18,10 @@ ${LOCALBIN}/controller-gen crd paths="./..." output:dir=../../charts/kubearchive
 
 # Generate role.
 echo "Generating role."
-${LOCALBIN}/controller-gen rbac:roleName="replaceme-operator" \
+${LOCALBIN}/controller-gen rbac:roleName="replaceme" \
     paths="./..." output:stdout | \
     sed -e 's/---/---\n# yamllint disable rule:indentation/' | \
-    sed -e 's/replaceme/{{ .Values.kubearchive.namespace }}/' > ../../charts/kubearchive/templates/operator/role.yaml
+    sed -e 's/replaceme/{{ tpl .Values.operator.name . }}/' > ../../charts/kubearchive/templates/operator/role.yaml
 
 echo "Generating deep copy code."
 # Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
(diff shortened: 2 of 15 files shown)


skoved (approved):

rh-hemartin (approved):

cmd/operator/generate.sh
```diff
@@ -21,7 +21,7 @@ echo "Generating role."
 ${LOCALBIN}/controller-gen rbac:roleName="replaceme-operator" \
     paths="./..." output:stdout | \
     sed -e 's/---/---\n# yamllint disable rule:indentation/' | \
-    sed -e 's/replaceme/{{ .Values.kubearchive.namespace }}/' > ../../charts/kubearchive/templates/operator/role.yaml
+    sed -e 's/replaceme/{{ tpl .Values.operator.name . }}/' > ../../charts/kubearchive/templates/operator/role.yaml
```
maruiz93, line 24: There is an issue here I need to address

maruiz93: Fixed!

beerparty (approved):

charts/kubearchive/values.yaml
```diff
@@ -49,18 +40,29 @@ sink:
...
 
+# values used to create the operator
 operator:
+  name: "{{ .Release.Name }}-operator"
   # NOTE - Helm does not resolve the `ko` path. The path needs to be set on the `helm install` command.
   # For example:
   # helm install -n default <chart-name> charts/kubearchive \
   # --set-string operator.image=$(ko build github.com/kubearchive/kubearchive/cmd/operator)
   image: ko://github.com/kubearchive/kubearchive/cmd/operator
+  leaderElection: "{{ tpl .Values.operator.name . }}-leader-election"
```
ggallen, line 51: Why are all these being added?

ggallen (approved):

charts/kubearchive/values.yaml
maruiz93: What do you mean? I extracted this value to the `values.yaml` to avoid repeating `{{ tpl .Values.operator.name . }}-leader-election` in the 6 places that it appears now in the files `leader_election_role.yaml` and `leader_election_role_binding.yaml`.

maruiz93: I've tried to comply with the conventions described in [here](https://kubearchive.github.io/kubearchive/main/contributors/naming-conventions.html)

mafh314 (approved):

charts/kubearchive/values.yaml
ggallen: I have to say as I look at this change that I don't like it.  I just don't see the value of extracting all these names individually to the `values.yaml` file.  The only part of these names that I think will ever change is the first part, and I don't think we'll ever change the name values defined in `values.yaml`.

I also don't see the need to put the additional settings (that weren't templated before) in the`values.yaml` either.  We don't have to have everything there.

I think we should try to keep `values.yaml` small in general,and only include things there that are likely to change.

ggallen (changes requested):

charts/kubearchive/values.yaml
```diff
@@ -51,18 +42,22 @@ sink:
...
 
+# values used to create the operator
 operator:
+  name: "{{ .Release.Name }}-operator"
   # NOTE - Helm does not resolve the `ko` path. The path needs to be set on the `helm install` command.
   # For example:
   # helm install -n default <chart-name> charts/kubearchive \
   # --set-string operator.image=$(ko build github.com/kubearchive/kubearchive/cmd/operator)
   image: ko://github.com/kubearchive/kubearchive/cmd/operator
+  metricsPort: 8443
```
ggallen, line 53: Why are these still exposed while all the other vars have gone elsewhere?

Again, I think these are unlikely to ever change, certainly not by a user.

ggallen (approved):
I left one other comment.

I still fail to see any real value in these changes, but it's not worth wasting any more time on it.  Feel free to commit.

charts/kubearchive/values.yaml
maruiz93: Fixed. I also removed the ports

PR #246, opened 2024-07-24
Enable 'go mod tidy' for renovate

### LINKED ISSUE #201: Enable go mod tidy for renovate
The renovate bot can be configured to tidy go modules before submitting a PR, we should enable that now that we are implementing a check for it in the CI in #179 . See the docs: https://docs.renovatebot.com/golang/#module-tidying

rh-hemartin:

Fixes #201 

## Release Note

```release-note
NONE
```

## Notes for Reviewers

* The configuration read was from the `main` branch, which did not contain the `goModTidy` option. I had to use the parameter `configurationFile` to make the changes on this branch matter. I removed it after testing because we don't need to specify it.
* This was tested at https://github.com/kubearchive/kubearchive/pull/245 by running manually the Renovate workflow from a branch with these changes.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,5 +1,8 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "osvVulnerabilityAlerts": true,
-  "semanticCommits": "disabled"
+  "semanticCommits": "disabled",
+  "postUpdateOptions": [
+    "gomodTidy"
+  ]
 }


.github/workflows/renovate.yml
```diff
@@ -33,6 +33,7 @@ jobs:
         uses: renovatebot/github-action@v40.2.3
         with:
           token: '${{ steps.get_token.outputs.token }}'
+          configurationFile: "./renovate.json"
```
skoved, line 36: `configurationFile` is for the [global config](https://docs.renovatebot.com/self-hosted-configuration/) which is a different config file than `renovate.json`. Since `postUpdateOptions` is a setting in `renovate.json`, this line should be removed

skoved (changes requested):

.github/workflows/renovate.yml
rh-hemartin: Ok, so I found that it always reads the configuration from the default branch, so then I need to remove the variable and force-merge this PR. That sounds right?

skoved: The global config is being read from the `env` of this action. We don't nee don't need to specify the config file. After renovate is started it reads our `renovate.json` and creates the PRs. See here:

https://github.com/kubearchive/kubearchive/actions/runs/10074716026/job/27851462762#step:5:175


We don't need `configurationFile`

skoved: As for testing changes to renovate.json, I'm not sure how to do it. Maybe there is a renovate linter?

skoved: it looks like this might help https://docs.renovatebot.com/config-validation/

beerparty (approved):

.github/workflows/renovate.yml
rh-hemartin: My claim that the configuration was not read is not true. However using `configurationFile` helped me test the behaviour of adding `goModTidy` as post-action. I will remove the parameter now that I tested it, but it will be useful in the future if we want to test behavior before merging.

skoved (approved):

PR #247, opened 2024-07-24
sink only attempts to delete a k8s object once

### LINKED ISSUE #232: Sink should not try to delete a resource from a DELETE cloudevent
after the sink deletes a resource, it receives a delete event for that same resource from ApiServerSource. This event needs to be processed to obtain the timestamp for when that resource was deleted. However, the sink needs to skip trying to delete the resource in this case

**Acceptance Criteria**:

- the sink processes delete events
- the sink does not try to delete already deleted resources

skoved:

Fixes #232 

## Release Note

```release-note
NONE
```

## Notes for Reviewers

The sink will stops processing `delete` type cloudevents after its payload is written into the database. Deletion criteria does not need to be checked since the object in the payload of the cloudevent has already been deleted

### DIFF
--- a/cmd/sink/main.go
+++ b/cmd/sink/main.go
@@ -13,6 +13,7 @@ import (
 	ceOtelObs "github.com/cloudevents/sdk-go/observability/opentelemetry/v2/client"
 	cloudevents "github.com/cloudevents/sdk-go/v2"
 	ceClient "github.com/cloudevents/sdk-go/v2/client"
+	"github.com/kubearchive/kubearchive/cmd/sink/events"
 	jsonpath "github.com/kubearchive/kubearchive/cmd/sink/jsonPath"
 	"github.com/kubearchive/kubearchive/pkg/database"
 	"github.com/kubearchive/kubearchive/pkg/models"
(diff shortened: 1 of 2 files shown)


beerparty (approved):

cmd/sink/main.go
```diff
@@ -104,6 +105,10 @@ func (sink *Sink) Receive(ctx context.Context, event cloudevents.Event) {
 	}
 	sink.Logger.Printf("successfully wrote cloudevent %s to the database\n", event.ID())
 	sink.Logger.Printf("checking if resource from cloudevent %s needs to be deleted\n", event.ID())
+	if events.IsDeleteType(event) {
```
rh-hemartin, line 108: What about:

```suggestion
	if event.Type() == eventingSources.ApiServerSourceDeleteEventType {
```

Where "eventingSources" is "knative.dev/eventing/pkg/apis/sources".

They have the strings also hardcoded and we already depend on the knative/eventing package. This will remove the need to create our own events pacakge.

skoved:
closing this PR in favor of the solution discussed earlier today

skoved:
> closing this PR in favor of the solution discussed earlier today

Ignore that comment

PR #248, opened 2024-07-25
Add kubectl-archive plugin with get command

### LINKED ISSUE #79: cli can get resources from kubearchive database
The cli can call appropriate endpoints provided by the api server to print out archived resources. Related to #78

rh-hemartin:

Fixes #79 

## Release Note

```release-note
NONE
```

## Notes for Reviewers

* We need to support namespaced and non-namespaced, to do in another issue.
* We need think about how to pass resource names. The problem with that twofold: pluralizing and api group. We may try to access the Kubernetes server to solve this problem. To do in another issue.


pkg/cmd/get.go
```diff
@@ -0,0 +1,182 @@
...
+
+func NewGetCmd() *cobra.Command {
+	o := NewGetOptions()
+
+	cmd := &cobra.Command{
+		Use:   "get [RESOURCES]",
+		Short: "Command to get resources from KubeArchive",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			var err error
+			err = o.Validate(args)
```
skoved, line 46: Instead of doing the validation of the args for this function in the `RunE` method, you could use [`cobra.Command.Args`](https://pkg.go.dev/github.com/spf13/cobra#Command.Args) ([see this example](https://github.com/spf13/cobra/blob/main/site/content/user_guide.md#positional-and-custom-arguments)). All you would need to do is add a `*cobra.Cmd` arg to `GetOptions.Validate` so that it matches the type for [`cobra.PositionalArgs` ](https://pkg.go.dev/github.com/spf13/cobra#PositionalArgs)

pkg/cmd/get.go
```diff
@@ -0,0 +1,182 @@
...
+			if err != nil {
+				return err
+			}
+
+			err = o.Complete(args)
```
skoved, line 51: It might be worth getting o.Complete to work as a [`cobra.Command.ValidArgsFunction`](https://pkg.go.dev/github.com/spf13/cobra#Command.ValidArgsFunction) ([see example](https://github.com/spf13/cobra/blob/main/site/content/completions/_index.md#dynamic-completion-of-nouns))

skoved (commented):
We might be able to leverage more of the features that cobra provides and simplify the `RunE` method

pkg/cmd/get.go
```diff
@@ -0,0 +1,182 @@
...
+		return nil, err
+	}
+	defer response.Body.Close()
+
+	if response.StatusCode != http.StatusOK {
+		bBytes, err := io.ReadAll(response.Body)
+		if err != nil {
+			return nil, err
+		}
+		return nil, fmt.Errorf("Status code != 200, but %d with body %s", response.StatusCode, string(bBytes))
```
beerparty, line 124: Does the `client.Get` call log information about the call? If not, this error message is not very useful. It says "status code != 200", but it doesn't say what you were doing. It doesn't even say that it was the return code from an HTTP call! It should at least say you were making an HTTP call to get a resource.

```suggestion
		return nil, fmt.Errorf("Return code from HTTP call to get resource %s was not 200, but %d with body %s", url, response.StatusCode, string(bBytes))
```

pkg/cmd/get.go
```diff
@@ -0,0 +1,182 @@
...
+		return list.Items, nil
+	})
+}
+
+func (o *GetOptions) getKubearchiveResources() ([]unstructured.Unstructured, error) {
+	o.RESTConfig.CAData = nil      // Remove CA data from the Kubeconfig
+	o.RESTConfig.CAFile = "ca.crt" // This expects you to have extracted the CA, see DEVELOPMENT.md
+	o.RESTConfig.BearerToken = *o.kubeFlags.BearerToken
+
+	// TODO: when our API returns a List resource, we can remove the decode function
```
beerparty, line 152: I don't see a `decode` function in the code below this comment. Is this a leftover TODO that has already been done?

pkg/cmd/get.go
rh-hemartin: I changed it by `cobra.ExactArgs(1)`, for now that will suffice. The moment we need to validate that the args have a certain pattern to them, I think we will need to extract it back.

pkg/cmd/get.go
rh-hemartin: `ValidArgsFunction` is supposed to return values for shell completion, while `Complete` fills out fields in options for later processing. I don't think its worth investing time here.

pkg/cmd/get.go
rh-hemartin: I improved all the error handling, included this. I included something similar where the url, the status code and the string body is printed.

pkg/cmd/get.go
rh-hemartin: I changed these comments slightly, the `decode` function is the inlined function below. It gets passed as a parameter called `decode` to `o.getResources` because it decodes the body into a certain type. Its needed because our API returns a list of resources, and the K8s API returns a resource `List`.

beerparty (approved):

DEVELOPMENT.md
```diff
@@ -130,19 +129,42 @@ By default, KubeArchive listens to `Event`s in the `test` namespace.
...
+    ```
+
+1. Get the Certificate Authority (CA) from the `kubearchive-api-server-tls` secret:
+    ```bash
+    kubectl get -n kubearchive secrets kubearchive-api-server-tls -o jsonpath='{.data.ca\.crt}' | base64 -d > ca.crt
+    ```
+
+1. Run the CLI:
+    ```bash
+    go run cmd/kubectl-archive/main.go get batch/v1/jobs --token $(kubectl -n kubearchive create token kubearchive-test)
```
mafh314, line 159: I was reading the whole DEVELOPMENT.md, and we mentioned the namespaces: `test` and kubearchive-test`. In the charts, we have defined  the `test` [namespace](https://github.com/kubearchive/kubearchive/blob/a00ecced3f7c18911558f55cdd25a76ee9b91e57/charts/kubearchive/templates/test/namespace.yaml). 
Maybe we need to add the creation of that namespace before these steps, or should we change it to `test` ?

maruiz93: `kubearchive-test` is the service account name, the namespace is `kubearchive`

we could move `kubearchive-test` to the `test` namespace though

mafh314: My bad, when I read this, I read: `-n kubearchive-test` :/

pkg/cmd/get.go
```diff
@@ -0,0 +1,162 @@
...
+		var list unstructured.UnstructuredList
+		err := json.Unmarshal(bodyBytes, &list)
+		if err != nil {
+			return nil, fmt.Errorf("error deserializing the body into unstructured.UnstructuredList: %w", err)
+		}
+		return list.Items, nil
+	})
+}
+
+func (o *GetOptions) getKubearchiveResources() ([]unstructured.Unstructured, error) {
```
maruiz93, line 127: shouldn't this be `getKubeArchiveResources` with the `A` capitalized?

maruiz93 (approved):

pkg/cmd/get.go
rh-hemartin: Fixed!

PR #251, opened 2024-07-25
introduce the first integration test for the operator

### LINKED ISSUE #127: Introduce the first integration test for the operator
The test should ensure that when a KubeArchiveConfig is created, all the required resources are created on the destination namespace.

mafh314:

Fixes #127 

## Release Note

```release-note
NONE
```

## Notes for Reviewers

The integration test checks for the following objects in the test-<random-string> namespace.
- Deployment
- ReplicaSet
- Pod
- ApiServerSource
- EventType (add, delete, update).
For the ApiServerSource and the EventType objects, the status check is done by a kubectl command + jsonpath filtering.

### DIFF
--- a/test/integration/integration_test.go
+++ b/test/integration/integration_test.go
@@ -7,11 +7,12 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"os/exec"
+	"strings"
 	"testing"
 	"time"
 
 	"github.com/kubearchive/kubearchive/test"
-
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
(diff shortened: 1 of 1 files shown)


rh-hemartin (approved):

test/integration/integration_test.go
```diff
@@ -7,11 +7,12 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"os/exec"
```
maruiz93, line 10: I see that this integration test runs `kubectl` CLI to interact with the cluster, is there any reason to go with this approach instead of using the [go kubernetes client ](https://github.com/kubernetes/client-go) that we are already using in the operator, the apiserver and the sink?

maruiz93 (changes requested):

test/integration/integration_test.go
mafh314: I use the go client to check the status of the core objects (Deployment, Replica set and the Pod). But for the ApiServerSource and the EvenType, I tried to set up a [Dynamic Client](https://pkg.go.dev/k8s.io/client-go/dynamic) to interact with those objects but I couldn't manage to set that correctly. So, I decided to check the status of those using kubectl.

maruiz93: I see that we are also using `kubectl`  and `ko` for creating, deleting and retrieving resources from the cluster in the [main file of the test package](https://github.com/kubearchive/kubearchive/blob/main/test/main.go#L67) so I guess it's ok to use it here too.

If we want to add a refactor for that in the future and use a go client for everything we can address that in a separate issue.

I've created https://github.com/kubearchive/kubearchive/issues/275 for that.

maruiz93 (approved):

test/integration/integration_test.go
rh-hemartin: Just to add a little bit of context...

On of the reasons to use the kubectl command is to deal with non-standard CRDs, its a pain to build the objects using the unstructured structs. On the other hand, building the YAML file and apply it is easier. We can use the go client for everything, but sometimes that is a pain.

And ko is more of a residue of the first implementation, that applied one of our images using `ko` to build it before using it. That can be removed for sure.

PR #97, opened 2024-06-27
Add release process

rh-hemartin:
Fixes #98 

```release-note
NONE
```

PR #257, opened 2024-07-25
Operator creates and managers a filter ConfigMap

### LINKED ISSUE #221: Operator creates and manages filter `ConfigMap`.
Until `ApiServerSource` can be enhanced to support the type of message filtering needed by `KubeArchive`, `KubeArchive` needs another way to let the sink know when the archive or delete a resource.

This can be done by having the operator create and manage a `ConfigMap` that will contain information that the sink can use to determine if a resource should be archived and then optionally deleted.

This `ConfigMap` will be named `filters` and reside in the `kubearchive` (or `KubeArchive` installation) namespace. It will take the following format:
```
---
kind: ConfigMap
apiVersion: v1
metadata:
  name: filters
  namespace: kubearchive
data:
  _global:
    - archive-when: .spec.kind == 'PipelineRun' and .status.state in ('Aborted, 'Completed', 'Initializing', 'Failed', 'Running')
      delete-when: .spec.kind == 'PipelineRun' and .status.state in ('Aborted, 'Completed', 'Failed')
  gallen:
    - archive-when: .spec.kind == 'CronJob' 
      delete-when: .spec.kind == 'CronJob' and .status.state in ('Completed')
    - archive-when: .spec.kind == 'Job' 
      delete-when: .spec.kind == 'Job' and .status.state in ('Completed', 'Failed')
  
```
In this `ConfigMap`, the `_global` entry lists filtering rules that should be applied to every namespace that `KubeArchive` is monitoring. This is set by the system administrator and cannot be overriden by users.  All other keys are namespace names, and the list under each namespace represents additional filter rule pairs to control archival and deletion for that particular namespace.

For namespace-specific rules, users would modify the `KubeArchiveConfig` resources in their namespace to add the filtering rules to be applied to that namespace. The operator will then update the `ConfigMap` with the latest rules during its reconciliation.

The operator will create and manage this `ConfigMap`. The sink is the consumer of the data in the `ConfigMap`, and uses the data to determine whether or not to archive the resource in the message.

ggallen:

Fixes #221

## Release Note

```release-note
NONE
```

## Notes for Reviewers

See details in https://github.com/kubearchive/kubearchive/issues/221.

### DIFF
--- a/cmd/operator/internal/controller/kubearchiveconfig_controller.go
+++ b/cmd/operator/internal/controller/kubearchiveconfig_controller.go
@@ -11,6 +11,8 @@ import (
 	"os"
 	"strings"
 
+	"gopkg.in/yaml.v3"
+
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	sourcesv1 "knative.dev/eventing/pkg/apis/sources/v1"
@@ -24,10 +26,13 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
 	kubearchivev1alpha1 "github.com/kubearchive/kubearchive/cmd/operator/api/v1alpha1"
 )
+
+const SinkFilterConfigMapName = "sink-filters"
 
 var k9eNs = os.Getenv("KUBEARCHIVE_NAMESPACE")
 var a14eName = k9eNs + "-a14e"
@@ -46,6 +51,7 @@ type KubeArchiveConfigReconciler struct {
 //+kubebuilder:rbac:groups=core,resources=serviceaccounts,verbs=create;delete;get;list;update;watch
 //+kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=roles;rolebindings,verbs=bind;create;delete;escalate;get;list;update;watch
 //+kubebuilder:rbac:groups=sources.knative.dev,resources=apiserversources,verbs=create;delete;get;list;update;watch
+//+kubebuilder:rbac:groups="",resources=configmaps,verbs=create;delete;get;list;update;watch
 
 func (r *KubeArchiveConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	log := log.FromContext(ctx)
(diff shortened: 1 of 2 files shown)


cmd/operator/internal/controller/kubearchiveconfig_controller.go
```diff
@@ -46,24 +50,51 @@ type KubeArchiveConfigReconciler struct {
...
+				return ctrl.Result{}, err
+			}
+		}
+
+		// Stop reconciliation as the resource is being deleted.
+		return ctrl.Result{}, nil
 	}
 
-	_, err = r.reconcileA14eServiceAccount(ctx, kaconfig)
+	_, err := r.reconcileA14eServiceAccount(ctx, kaconfig)
```
beerparty, line 97: Why do you need a colon here? `err` is defined on line 61 of the new code.

ggallen: The compiler gives an error. I believe the `err` defined on line 61 is scoped to the `if` statement and not outside.

maruiz93:
Aren't we using the [PR template](https://github.com/kubearchive/kubearchive/blob/main/.github/pull_request_template.md) anymore for new PRs?

ggallen:
> Aren't we using the [PR template](https://github.com/kubearchive/kubearchive/blob/main/.github/pull_request_template.md) anymore for new PRs?

I didn't know we had or agreed to use any PR template.  If it continues to make the git comments unreadable, then I say no.

maruiz93:
> I didn't know we had or agreed to use any PR template. If it continues to make the git comments unreadable, then I say no.

If I'm not mistaken is part of the [release process](https://github.com/kubearchive/kubearchive/pull/97) that we recently introduced. 
As I see it, it makes no sense to have the template in place if we are going to ignore it.

Also, I think that @rh-hemartin made some changes to fix the readability of the commits.

ggallen:
> As I see it, it makes no sense to have the template in place if we are going to ignore it.

I didn't see the PR and didn't review it, so I was completely unaware of this change. The title is release process, which I guess includes git commits but that's not how I thought about it.

I'm happy to use the template, but I'd like to see how it actually works.  Here's a single commit message. Do you think it's readable?
```
commit 368153791e9e24ea9786210d4c6ddaf52c65289f
Merge: f5ca644 575e80d
Author: Sam Koved <skoved@redhat.com>
Date:   Wed Jul 24 15:05:46 2024 -0400

    sink deletes k8s objects when json path returns a value (#217)
    
    <!--  Thanks for sending a pull request!
    
    Please read our contributor guidelines:
    https://kubearchive.github.io/kubearchive/main/contributors/guide.html
    -->
    
    <!-- Please make sure to create an issue you can link to and mention it
    here
    to automatically close it. If this PR covers part of the work, instead
    of
    "Fixes #", use "Related to #"
    
    Usage: `Fixes #<issue number>`, or `Fixes (paste link of issue)`.
    -->
    
    Fixes #191
    
    ## Release Note
    <!--
    If this change has user-visible impact, write a release note in the
    block
    below. Include the string "action required" if additional action is
    required of
    users switching to the new release, for example in case of a breaking
    change.
    
    Write as if you are speaking to users, not other KubeArchive
    contributors. If this
    change has no user-visible impact, leave the code block as is.
    -->
    
    ```release-note
    NONE
    ```
    
    ## Notes for Reviewers
    
    The operator grants the sink's ServiceAccount permissions to delete the
    resource type(s) being watched. When deploying kubearchive, the user can
    specify a JSON Path that, if it returns a value when applied to a
    kubernetes object, will cause the sink to delete the kubernetes object
    from the cluster.
    
```

cmd/operator/internal/controller/kubearchiveconfig_controller.go
```diff
@@ -358,3 +394,93 @@ func (r *KubeArchiveConfigReconciler) desiredA14e(kaconfig *kubearchivev1alpha1.
...
+		cm = &corev1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      SinkFilterConfigMapName,
+				Namespace: k9eNs,
+			},
+			Data: map[string]string{},
+		}
+	}
+
+	jsonBytes, err := json.Marshal(kaconfig.Spec.Resources)
```
maruiz93, line 448: PTAL to my last comment in the issue: https://github.com/kubearchive/kubearchive/issues/221#issuecomment-2252643480

ggallen: Yeah, I replied there. The YAML suggest is great.  I'll try that.

cmd/operator/internal/controller/kubearchiveconfig_controller.go
beerparty: Ah, okay. Thanks for checking.

cmd/operator/internal/controller/kubearchiveconfig_controller.go

maruiz93 (approved):

PR #265, opened 2024-07-26
sink does not try to delete k8s objects twice

### LINKED ISSUE #232: Sink should not try to delete a resource from a DELETE cloudevent

skoved:

Fixes #232 

The after the sink deletes a kubernetes object, it will update the entry in the database for that resource with timestamp of when the resource was deleted. Bumped the version of knative-eventing to v1.15.0 so that the operator can apply a filter to all ApiServerSources so that they do not send delete type cloudevents.

## Release Note

```release-note
NONE
```

## Notes for Reviewers

- knative-eventing version bump `v1.14.3` -> `v1.15.0`
- operator applies a filter to ApiServerSources so they do not send `delete` type cloudevents

To test:
- run `hack/quick-install.sh`
- initialize the database
- run `kubectl apply -f https://storage.googleapis.com/tekton-releases/pipeline/latest/release.yaml`
- run
  ```bash
  kubectl apply -f - << EOF
  ---
  kind: KubeArchiveConfig
  metadata:
    name: kubearchive
    namespace: demo-tekton
  spec:
    resources:
      - selector:
          apiVersion: tekton.dev/v1beta1
          kind: PipelineRun
        archive-when: .status != 'Completed'
        delete-when: .status == 'Completed'
  ---
  apiVersion: v1
  kind: ServiceAccount
  metadata:
    name: demo-tekton
    namespace: demo-tekton
  ---
  apiVersion: rbac.authorization.k8s.io/v1
  kind: ClusterRole
  metadata:
    name: pipelineruns-reader
  rules:
  - apiGroups: ["tekton.dev"]
    resources: ["pipelineruns"]
    verbs: ["get", "list"]
  ---
  apiVersion: rbac.authorization.k8s.io/v1
  kind: ClusterRoleBinding
  metadata:
    name: read-pipelineruns
  subjects:
  - kind: ServiceAccount
    name: demo-tekton
    namespace: demo-tekton
    apiGroup: ""
  roleRef:
    kind: ClusterRole
    name: pipelineruns-reader
  EOF
  ```
- in a separate terminal run `kubectl logs -n kubearchive -l app=kubearchive-sink -f`
- in the first terminal run `kubectl apply -f https://raw.githubusercontent.com/skoved/kubearchive/first-demo/tekton.yaml`

After 10-20 seconds you will see only one log for deleting the PipelineRun and you will see a log after it for updating it in the database after deleting it

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -42,15 +42,29 @@ Install these tools:
     export KO_DOCKER_REPO="kind.local"
     ```
 
-1. Install knative-eventing core and cert-manager and wait for them to be ready:
+1. Install knative-eventing core and cert-manager, wait for them to be ready, and enable new-apiserversource-filters:
     ```bash
     export CERT_MANAGER_VERSION=v1.9.1
-    export KNATIVE_EVENTING_VERSION=v1.14.3
+    export KNATIVE_EVENTING_VERSION=v1.15.0
 
     kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/${CERT_MANAGER_VERSION}/cert-manager.yaml
     kubectl apply -f https://github.com/knative/eventing/releases/download/knative-${KNATIVE_EVENTING_VERSION}/eventing-core.yaml
     kubectl rollout status deployment --namespace=cert-manager --timeout=30s
     kubectl rollout status deployment --namespace=knative-eventing --timeout=30s
+
+    kubectl apply -f - << EOF
+    apiVersion: v1
+    kind: ConfigMap
+    metadata:
+      name: config-features
+      namespace: knative-eventing
+      labels:
+        eventing.knative.dev/release: devel
+        knative.dev/config-propagation: original
+        knative.dev/config-category: eventing
+    data:
+      new-apiserversource-filters: enabled
+    EOF
     ```
 
 1. Generate operator code:
--- a/cmd/sink/main.go
+++ b/cmd/sink/main.go
@@ -108,17 +108,28 @@ func (sink *Sink) Receive(ctx context.Context, event cloudevents.Event) {
 		)
 	}
 	if mustDelete {
-		sink.Logger.Println("attempting to delete kubernetes object:", k8sObj.GetUID())
+		sink.Logger.Println("requesting to delete kubernetes object:", k8sObj.GetUID())
 		kind := k8sObj.GetObjectKind().GroupVersionKind()
 		resource, _ := meta.UnsafeGuessKindToResource(kind) // we only need the plural resource
 		k8sCtx, k8sCancel := context.WithTimeout(context.Background(), time.Second*5)
 		defer k8sCancel()
 		err = sink.K8sClient.Resource(resource).Namespace(k8sObj.GetNamespace()).Delete(k8sCtx, k8sObj.GetName(), metav1.DeleteOptions{})
 		if err != nil {
-			sink.Logger.Printf("could not delete resource %s: %s\n", k8sObj.GetUID(), err)
+			sink.Logger.Printf("failed to request resource %s be deleted: %s\n", k8sObj.GetUID(), err)
 			return
 		}
-		sink.Logger.Println("successfully deleted kubernetes object:", k8sObj.GetUID())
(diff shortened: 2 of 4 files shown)


cmd/sink/main.go
```diff
@@ -119,6 +119,17 @@ func (sink *Sink) Receive(ctx context.Context, event cloudevents.Event) {
...
 		}
 		sink.Logger.Println("successfully deleted kubernetes object:", k8sObj.GetUID())
+		deleteTs := metav1.Now()
+		k8sObj.SetDeletionTimestamp(&deleteTs)
+		sink.Logger.Printf("updating database entry for kubernetes object %s after successful deletion\n", k8sObj.GetUID())
+		updateCtx, updateCancel := context.WithTimeout(context.Background(), time.Second*5)
+		err = sink.Db.WriteResource(updateCtx, k8sObj, event.Data())
+		defer updateCancel()
+		if err != nil {
+			sink.Logger.Printf("failed to update database entry for kubernetes object %s after successful deletion\n", k8sObj.GetUID())
```
ggallen, line 129: This comment applies to this log message and also one earlier in the file that is not part of this change.

It's not really a successful deletion.  The message implies the deletion is done, but it's not.  It's really a successful submit of the deletion request to kubernetes.

I think this message could actually say something shorter, like "failed to update `ts_deleted`".  And earlier "updating `ts_deletion` for resource uuid  <primary-key>"

ggallen (changes requested):

cmd/sink/main.go
skoved: fixed the log messages

ggallen (approved):

DEVELOPMENT.md
```diff
@@ -42,15 +42,29 @@ Install these tools:
     export KO_DOCKER_REPO="kind.local"
     ```
 
-1. Install knative-eventing core and cert-manager and wait for them to be ready:
+1. Install knative-eventing core and cert-manager, wait for them to be ready, and enable new-apiserversource-filters:
```
beerparty, line 45: Rather than have this code in the DEVELOPMENT.md file, why not refer them to the `hack/quick-install.sh` script? Isn't this the same thing?

rh-hemartin: The hack/quick-install is used on the CI. We may have different instructions for CI than for development, as we mentioned yesterday in the meeting (CI to test production procedure).

After we make changes for production release at #277 , maybe we can see if it makes sense to replace some DEVELOPMENT steps by the hack/quick-install.sh

skoved: I'm going to resolve this thread for now and we can discuss doing this after #277

cmd/sink/main.go
```diff
@@ -108,17 +108,28 @@ func (sink *Sink) Receive(ctx context.Context, event cloudevents.Event) {
...
-			sink.Logger.Printf("could not delete resource %s: %s\n", k8sObj.GetUID(), err)
+			sink.Logger.Printf("failed to request resource %s be deleted: %s\n", k8sObj.GetUID(), err)
 			return
 		}
-		sink.Logger.Println("successfully deleted kubernetes object:", k8sObj.GetUID())
+		sink.Logger.Printf("successfully requested kubernetes object %s be deleted\n", k8sObj.GetUID())
+		deleteTs := metav1.Now()
+		k8sObj.SetDeletionTimestamp(&deleteTs)
+		sink.Logger.Println("updating cluster_deleted_ts for kubernetes object:", k8sObj.GetUID())
+		updateCtx, updateCancel := context.WithTimeout(context.Background(), time.Second*5)
```
rh-hemartin, line 125: You need to use the context from the request to bundle the traces correctly.

```suggestion
		updateCtx, updateCancel := context.WithTimeout(ctx, time.Second*5)
```

There is another coincidence of this for the `k8sCancel` a few lines above this, I know its not part of this PR, so feel free to keep it as is.

rh-hemartin (approved):
A thing related with tracing that does not impact behavior, otherwise looks good to me.

beerparty (approved):

Issue #85, opened 2024-06-19
Initialize the DB when the container starts


mafh314:
Create the necessary tables and load the data for testing mounting .sql files under /docker-entrypoint-initdb.d 
(https://github.com/docker-library/docs/blob/master/postgres/README.md#initialization-scripts)

rh-hemartin:
I think this will have to be manual in some way, I wouldn't push too much for this one. If we need to make a manual step, the manual step can be to apply those files and that it, no need to think the process to be able to mount the files before the container is created.

rh-hemartin:
For the future, I think this should not live in the database side. I think the responsible should be the sink/api that runs the database initialization before starting the application, as any other database application does. This will ensure that the moment the sink/api is started, the database is in the correct state.

maruiz93:
I wouldn't let the API (a reader with only read permissions on the DB) should be able to create schemas / tables.

I have my doubts on the sink too. I prefer to have everything decoupled, and I would say that both sink and api should check the connection to the db when starting and if fails fail to start the container.

I get your point though of not having the tables provided in the database when the pod starts because I think that the user of kubearchive may provide its own DB without the schema, just provide the permissions in a specific database.

I think that in a production environment the operator should have the responsability of setting the tables.

Now, if we accept that the operator would be the responsible of configuring the DB schema, as in development we aren't using it and we are using helm, the equivalent procedure would be one of the following (as I see it):

* Run the init_db script manually
* Mount the .sql files in the DB pod as suggested in this issue

maruiz93:
> We are using helm to initialize the database?

I meant that we are using helm instead an operator. 

If we agree on the operator to be the one that configure the schema in the database  (I see that we haven't yet) we should do something in development for initializing it. The two options that I see are:

* Run the script manually (this is what we are doing now, and this is outside of the helm install command)
* Make the DB container to automatically set up the kubearchive schema (so this will be done as part of the helm install command)

> I'm not sure I agree with this. The operator, at least currently, isn't doing any sort of dependency configuration/installation.

It's not a problem that the operator doesn't do it right now but this is a good discussion to set up the expectations on what the operator should and shouldn't do in the future.

As I stated before, for me, configuring the schema makes sense to be done out from the sink/api, and as I see the operator as the "deployer" I also see it as "the one who configure every kubearchive component to get things working". I envision it as:

- [DB] Create the appropriate tables/index (and lookup data if we have it at any point)
- [Events] Create the APIServerSource CRs
[...]

PR #269, opened 2024-07-26
Extract DDL to its own file

### LINKED ISSUE #258: Don't initialize "public.resource", use "resource" instead
The KubeArchive init process creates the table "public.resource". "public" is the default namespace, but database admins may want to restrict user access to a certain namespace. To allow administrators to manage better the kubearchive user permissions, KubeArchive needs to drop the "public". This change will make KubeArchive compatible with PostgreSQL schemas if they administrators want to.

PostgreSQL schemas: https://www.postgresql.org/docs/current/ddl-schemas.html

**Edit**: changed the approach to drop "public" first.

maruiz93:

Fixes #258
Related to #85

## Release Note

```release-note
Move resources table from public to kubearchive schema.
```

## Notes for Reviewers

* Extracted the SQL code embedded in init_db.go to ddl.sql
* Move the `resources` table from `public` default schema to `kubearchive` schema.


skoved (approved):

database/ddl.sql
```diff
@@ -0,0 +1,30 @@
...
+    "namespace" varchar NOT NULL,
+    "resource_version" varchar NULL,
+    "created_at" timestamp NOT NULL DEFAULT now(),
+    "updated_at" timestamp NOT NULL DEFAULT now(),
+    "cluster_deleted_ts" timestamp NULL,
+    "data" jsonb NOT NULL
+    );
+
+CREATE OR REPLACE TRIGGER set_timestamp
+BEFORE UPDATE ON kubearchive.resource
```
ggallen, line 28: I thought we were going to set `created_at` via a trigger as well?

ggallen (changes requested):

beerparty (approved):
I support the creation of a file for just the DDL and the other updates. I have no opinion on the content of the new DDL.

database/ddl.sql
maruiz93: We have `created_at` value automated, but as this should only be set when the registry is first created, the default value of the field gets the work done.

Note that we have both `created_at` and `updated_at` defaults to `now()`.

You can find more information in this PR https://github.com/kubearchive/kubearchive/pull/210 where I followed the steps from [this guide](https://medium.com/@alexishevia/automatic-timestamps-with-postgresql-8a5fc499d887)

rh-hemartin (approved):

mafh314 (approved):

ggallen (approved):

PR #273, opened 2024-07-29
Set service name for telemetry

### LINKED ISSUE #207: Set service.name for OpenTelemetry spans
Currently we are sending telemetry spans with an empty service name from both the KubeArchive sink and API, we need to add its name. Probably `kubearchive.<component>` is the right move for the name, so `kubearchive.sink` and `kubearchive.api`, and eventually `kubearchive.operator`.

**Notes**

* The `kubearchive.api` string we are sending to the otelgin library its not setting the service name.

rh-hemartin:

Resolves #207 

## Release Note

```release-note
NONE
```

## Notes for Reviewers

* The change has been made in code even if `OTEL_SERVICE_NAME` can be set because the value of the name is more a constant than a variable, so it made more sense to me to add it within the code.
* I can extract the name to a variable at the top of the file instead, but hardcoding it on the function didn't look crazy to me.

### DIFF
--- a/cmd/api/main.go
+++ b/cmd/api/main.go
@@ -18,6 +18,8 @@ import (
 	"k8s.io/client-go/rest"
 )
 
+const otelServiceName = "kubearchive.api"
+
 type Server struct {
 	k8sClient kubernetes.Interface
 	router    *gin.Engine
@@ -37,7 +39,7 @@ func getKubernetesClient() *kubernetes.Clientset {
 
 func NewServer(k8sClient kubernetes.Interface, controller routers.Controller) *Server {
 	router := gin.Default()
-	router.Use(otelgin.Middleware("kubearchive.api"))
+	router.Use(otelgin.Middleware("")) // Empty string so the library sets the proper server
 	router.Use(auth.Authentication(k8sClient.AuthenticationV1().TokenReviews()))
 	router.Use(auth.RBACAuthorization(k8sClient.AuthorizationV1().SubjectAccessReviews()))
 	// TODO - Probably want to use cache for the discovery client
@@ -53,7 +55,7 @@ func NewServer(k8sClient kubernetes.Interface, controller routers.Controller) *S
 }
 
 func main() {
-	err := observability.Start()
+	err := observability.Start(otelServiceName)
 	if err != nil {
 		log.Printf("Could not start opentelemetry: %s", err)
 	}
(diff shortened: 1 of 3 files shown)


pkg/observability/trace.go
```diff
@@ -39,6 +40,9 @@ func Start() error {
 		resource.WithProcessRuntimeName(),
 		resource.WithProcessRuntimeVersion(),
 		resource.WithFromEnv(),
+		resource.WithAttributes(
+			semconv.ServiceName(serviceName),
+		),
```
skoved, line 45: `resource.WithFromEnv()` gets called here. It can set the service name from the `OTEL_SERVICE_NAME` environment variable. That is the way that I think we should set the service name.

https://github.com/open-telemetry/opentelemetry-go/blob/main/sdk/resource/env.go#L40

skoved (changes requested):

cmd/api/main.go
```diff
@@ -53,7 +53,7 @@ func NewServer(k8sClient kubernetes.Interface, controller routers.Controller) *S
 }
 
 func main() {
-	err := observability.Start()
+	err := observability.Start("kubearchive.api")
```
skoved, line 56: I think `"kubearchive.api"` should be a constant instead of a magic string

cmd/sink/main.go
```diff
@@ -124,7 +124,7 @@ func (sink *Sink) Receive(ctx context.Context, event cloudevents.Event) {
 
 func main() {
 	logger := log.New(os.Stderr, "", log.LstdFlags|log.Lmicroseconds|log.LUTC)
-	err := kaObservability.Start()
+	err := kaObservability.Start("kubearchive.sink")
```
skoved, line 127: I think `"kubearchive.sink"` should be a constant instead of a magic string

skoved (changes requested):

pkg/observability/trace.go
skoved: I realize you mentioned this in the PR description

cmd/api/main.go
rh-hemartin: Moved to the top as a constant.

cmd/sink/main.go
rh-hemartin: Moved to the top as a constant.

mafh314 (approved):

beerparty (approved):

skoved (approved):

Issue #253, opened 2024-07-25
Images are set to the production ones


rh-hemartin:
Currently our Helm values contain `ko` image references. To be installed easily in production, those need to be set to something other than ko. In this case, the release process sets the appVersion and version on the helm chart, and those are the same used for image tagging. One solution would be to default the images to something like:

```
quay.io/kubearchive/operator:{{ .Chart.appVersion}}
```

So the version set by the release is used.

rh-hemartin:
This maybe does not make sense now that we agreed to use a single YAML for release. The reason would be that we probably use `helm template | ko resolve -f -` to generate the YAML for release, so the current image references are OK and they don't need to change. See #277 for information about the single YAML for release.

rh-hemartin:
This is not needed anymore, as we will use the `helm template | ko resolve` we need the references to be `ko` images. See #328.

PR #274, opened 2024-07-29
Add installation docs

### LINKED ISSUE #252: Installation documentation
We need to start thinking about installation for production. We need some starting documentation so people interested can install it and configure it to a bare minimum.

rh-hemartin:

Resolves #252 

## Release Note

```release-note
NONE
```

## Notes for Reviewers

* Needs #258  and #253
* There have been discussions about doing an installation with a single YAML instead of Helm. While that is possible, it need changes on multiple places. I prefer to document the current installation state and then work towards moving to a new installation process.

### DIFF
--- a/docs/modules/ROOT/pages/getting-started/installation.adoc
+++ b/docs/modules/ROOT/pages/getting-started/installation.adoc
@@ -1,0 +1,45 @@
+= Installation
+
+This document explains how to install and configure KubeArchive in your Kubernetes cluster.
+
+== Prerequisites
+
+* The `helm` CLI
+* A PostgreSQL instance
+* CertManager is installed on the Kubernetes cluster (+v1.9.1)
+* Knative Eventing is installed on the Kubernetes cluster (+v1.14.3)
+
+== Prepare the Database
+
+1. Create a user and a password for KubeArchive to use.
+1. Create a database for the user to use.
+1. As the user, execute the `ddl.sql` we provide in the database, located at:
++
+[source,bash]
+----
+https://github.com/kubearchive/kubearchive/blob/<VERSION>/database/ddl.sql
+----
++
+This file creates a PostgreSQL schema, a table for KubeArchive and other stuff KubeArchive
+requires.
+
+== Install KubeArchive
+
+1. Create a [filename]`values.yml` file and add the database information:
++
+[source,yaml]
+----
+---
+database:
+  url: "<postresql.example.com>"
+  postgresData:
+    dbName: <example-db>
+    dbUser: <example-user>
+    dbPassword: <example-password>
+----
+
+1. Install KubeArchive using `helm`:
+[source,bash]
+----
+helm install kubearchive -n kubearchive quay.io/kubearchive/kubearchive-helm:<VERSION> -f values.yml --create-namespace
+----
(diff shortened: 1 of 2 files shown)


docs/modules/ROOT/pages/getting-started/installation.adoc
```diff
@@ -0,0 +1,45 @@
...
+This document explains how to install and configure KubeArchive in your Kubernetes cluster.
+
+== Prerequisites
+
+* The `helm` CLI
+* A PostgreSQL instance
+* CertManager is installed on the Kubernetes cluster (+v1.9.1)
+* Knative Eventing is installed on the Kubernetes cluster (+v1.14.3)
+
+== Prepare the database
```
beerparty, line 12: Title capitalization:

```suggestion
== Prepare the Database
```

docs/modules/ROOT/pages/getting-started/installation.adoc
```diff
@@ -0,0 +1,45 @@
...
+
+1. Create a user and a password for KubeArchive to use.
+1. Create a database for the user to use.
+1. As the user, execute the `ddl.sql` we provide in the database, located at:
```
beerparty, line 16: Since you are providing the location right below this, providing the location in the text is redundant. Also, the file name needs to be tagged:

```suggestion
1. As the user, execute the [filename]`ddl.sql` we provide, located at:
```

docs/modules/ROOT/pages/getting-started/installation.adoc
```diff
@@ -0,0 +1,45 @@
...
+---
+database:
+  url: "<postresql.example.com>"
+  postgresData:
+    dbName: <example-db>
+    dbUser: <example-user>
+    dbPassword: <example-password>
+----
+
+1. Install using helm
```
beerparty, line 41: In the previous examples, you ended the text with a colon and added a blank line (with the "+") before the example code. I think you should do the same here for consistency.  I also think you should explicitly say "Install KubeArchive" in the text.

```suggestion
1. Install KubeArchive using `helm`:
+
```

beerparty (commented):
Some suggestions included.

docs/modules/ROOT/pages/getting-started/installation.adoc
rh-hemartin: Using `[filename]` is on the Red Hat Asciidoc guide, but why it is used? Do you know? I rather not use it, if it does not add value.

maruiz93: The square brackets used before a word in backticks are for role assignment, what in html means the `class` in the HTML element.

The specific docs are [here](https://docs.asciidoctor.org/asciidoc/latest/attributes/role/#assign-roles-to-formatted-inline-elements).

It is currently being rendered like that:
image

Now about the question of if we should keep adding this? It doesn't make any difference adding this class because there is no specific `CSS` rule for it but we may want to introduce it at some point. See how to do this in Antora [here](https://docs.antora.org/antora-ui-default/stylesheets/#add-a-new-css-rule).

rh-hemartin: Ah, I see.

PR #281, opened 2024-07-30
Add instrumentation for HTTP client

### LINKED ISSUE #202: Add instrumentation for the http client embedded in the k8s client for kubearchive rest api server
We need logs for the requests done from the api server to the kubernetes API.

That imply configuring the kubernetes client with the proper [Transport](https://github.com/kubernetes/client-go/blob/master/rest/config.go#L104)

I think that we can follow [this piece of code](https://uptrace.dev/get/instrument/opentelemetry-net-http.html#instrumenting-http-client) to add the otel transport for instrumenting the http client of the k8s client

rh-hemartin:

Resolves #202 

## Release Note

```release-note
NONE
```

## Notes for Reviewers

* The approach for testing (creating a RESTClient) has been inspired by code that tests the kubectl code. Its bulky, but gets the work done. I didn't look at it enough to see how to make it easier, if that is possible.

### DIFF
--- a/cmd/api/discovery/apiresources.go
+++ b/cmd/api/discovery/apiresources.go
@@ -5,22 +5,44 @@ import (
 	"net/http"
 
 	"github.com/gin-gonic/gin"
-	"k8s.io/client-go/discovery"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/rest"
 
 	"github.com/kubearchive/kubearchive/cmd/api/abort"
 )
 
 // GetAPIResource set apiResource attribute in the context based on the path parameters group version and resourceType
-func GetAPIResource(discoveryInterface discovery.DiscoveryInterface) gin.HandlerFunc {
+//
+// This function relied on the DiscoveryClient's ServerResourcesForGroupVersion function. However it does not accept a context
+// so we had to implement the same call to be able to pass the context so telemetry traces are tied together
+// @rh-hemartin opened a ticket to allow for a context on that function, see https://github.com/kubernetes/client-go/issues/1370
+func GetAPIResource(client rest.Interface) gin.HandlerFunc {
 	return func(c *gin.Context) {
 		groupVersion := fmt.Sprintf("%s/%s", c.Param("group"), c.Param("version"))
-		resourceList, err := discoveryInterface.ServerResourcesForGroupVersion(groupVersion)
-		if err != nil {
-			abort.Abort(c, err.Error(), http.StatusBadRequest)
+
+		// TODO: when we support the core API we need to change this
+		url := fmt.Sprintf("/apis/%s", groupVersion)
+		result := client.Get().AbsPath(url).Do(c.Request.Context())
+
+		if result.Error() != nil {
+			status := 0
+			result.StatusCode(&status)
+			abort.Abort(c, fmt.Errorf("Unable to retrieve information from '%s', error: %w", url, result.Error()).Error(), status)
 			return
 		}
+
+		resources := &metav1.APIResourceList{
+			GroupVersion: groupVersion,
+		}
+
+		err := result.Into(resources)
+		if err != nil {
+			abort.Abort(c, fmt.Errorf("Unable to deserialize result from '%s', error: %w", url, err).Error(), http.StatusInternalServerError)
+			return
+		}
+
 		resourceName := c.Param("resourceType")
-		for _, resource := range resourceList.APIResources {
+		for _, resource := range resources.APIResources {
 			if resource.Name == resourceName {
 				c.Set("apiResource", resource)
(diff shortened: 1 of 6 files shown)


rh-hemartin:
I'm having some problems related with testing. Due to the implementation of the fake library, I can't access the RESTClient on the Discovery field of the fake Clientset.

cmd/api/discovery/apiresources_test.go
```diff
@@ -33,7 +42,33 @@ func TestGetAPIResource(t *testing.T) {
...
+				default:
+					t.Fatalf("unexpected request: %#v\n%#v", req.URL, req)
+					return nil, nil
+				}
+			default:
+				t.Fatalf("unexpected request: %s %#v\n%#v", req.Method, req.URL, req)
+				return nil, nil
+			}
+		}),
+	}
```
maruiz93, line 71: I also played around with how to mock the restClient and I came out with a different solution.
I wanted to take advantage of the faked `ServerResourcesForGroupVersion` that was being used before, so I call it and mock the RestClient after that:

```suggestion
           // call ServerResourcesForGroupVersion and properly set the response and the status of the server
			apiResources, err := fakeDiscovery.ServerResourcesForGroupVersion(fmt.Sprintf("%s/%s", tc.group, tc.version))
			status := http.StatusOK
			response := ""
			if err != nil {
				response = err.Error()
				status = http.StatusBadRequest
			} else {
				buf := bytes.NewBufferString("")
				serializer.Encode(apiResources, buf)
				response = buf.String()
			}
           // mock the restClient with the given status and response
			restClient := &fakerest.RESTClient{
				Client: fakerest.CreateHTTPClient(func(request *http.Request) (*http.Response, error) {
					resp := &http.Response{
						StatusCode: status,
						Body: io.NopCloser(
							strings.NewReader(response)),
					}
					return resp, nil
				}),
				NegotiatedSerializer: scheme.Codecs.WithoutConversion(),
				GroupVersion:         schema.GroupVersion{Version: "v1"},
				VersionedAPIPath:     fmt.Sprintf("/apis/%s/%s", tc.group, tc.version),
			}
```

For that to work is needed also the serializer:

```
	s := runtime.NewScheme()
	serializer := json.NewSerializerWithOptions(json.DefaultMetaFactory, s, s, json.SerializerOptions{
		Pretty: true,
		Yaml:   false,
		Strict: true,
	})
```

What I like of this solution is that the logic is mostly deferred to the fake module, I just put in place the glue between the faked `ServerResourcesForGroupVersion` and the RestClient

maruiz93 (changes requested):

cmd/api/discovery/apiresources.go
```diff
@@ -5,22 +5,29 @@ import (
...
+			GroupVersion: groupVersion,
+		}
+		// TODO: when we support the core API we need to change this
+		url := fmt.Sprintf("/apis/%s", groupVersion)
+		err := client.Get().AbsPath(url).Do(c.Request.Context()).Into(resources)
 		if err != nil {
-			abort.Abort(c, err.Error(), http.StatusBadRequest)
-			return
+			abort.Abort(c, fmt.Errorf("Unable to retrieve information from '%s', error: %w", url, err).Error(), http.StatusBadRequest)
 		}
```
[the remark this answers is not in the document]
rh-hemartin: I separated both stages, as you suggested.

cmd/api/discovery/apiresources_test.go
rh-hemartin: Good idea! I went with this idea but dropped the serializer in favor of just using `json.Marshal`.

maruiz93: Nice! I like how it looks now :)

cmd/api/discovery/apiresources.go
```diff
@@ -5,22 +5,40 @@ import (
...
-	"k8s.io/client-go/discovery"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/rest"
 
 	"github.com/kubearchive/kubearchive/cmd/api/abort"
 )
 
 // GetAPIResource set apiResource attribute in the context based on the path parameters group version and resourceType
-func GetAPIResource(discoveryInterface discovery.DiscoveryInterface) gin.HandlerFunc {
+func GetAPIResource(client rest.Interface) gin.HandlerFunc {
```
maruiz93, line 15: How about adding a comment referencing the issue you opened to get DiscoveryInterface to admit a context?

maruiz93 (approved):

cmd/api/discovery/apiresources.go
rh-hemartin: Added a paragraph for context.

skoved (approved):

PR #282, opened 2024-07-30
Fix integration tests

### LINKED ISSUE #127: Introduce the first integration test for the operator

rh-hemartin:

Fixes #127 

## Release Note

```release-note
NONE
```

## Notes for Reviewers

* Today I merged #251 but I missed that the schema for KubeArchiveConfig changed in the last days. The integration tests now fail because of that. I'm submitting this to fix it.

### DIFF
--- a/test/integration/integration_test.go
+++ b/test/integration/integration_test.go
@@ -84,8 +84,11 @@ metadata:
   namespace: %s
 spec:
   resources:
-    - apiVersion: batch/v1
-      kind: Deployment
+    - selector:
+        apiVersion: v1
+        kind: Event
+      archive-when: .status != 'Completed'
+      delete-when: .status == 'Completed'
 `, namespaceName)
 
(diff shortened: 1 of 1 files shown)


maruiz93 (approved):

mafh314 (approved):

rh-hemartin:
> I think we should see what we can do to prevent this kind of thing to happen.
> 
> The problem was that the CI in #251 didn't ran with the latest changes. Maybe there is an option to enforce having the PR rebased with main before allowing the merge?
> 
> Maybe there are other options? what do you think?

Good question. I think we can force every PR to be up to date with main, but given how fast we are merging these days, I think it maybe become a problem. Maybe we can open it for discussion and see if we want to test the approach and then evaluate the impact if any.

Issue #285, opened 2024-07-30
operator creates invalid cel expressions in KubeArchiveConfig


skoved:
The operator creates a KubeArchiveConfig in the `test` namespace with invalid cel expressions

```yaml
archiveWhen: .status != 'Completed'
deleteWhen: .status == 'Completed'
```

The error returned by the cel language parser is 

```bash
ERROR: <input>:1:9: found no matching overload for '_==_' applied to '(map(string, dyn), string)'
| .status == 'Completed'
| ........^
```

This causes errors in the sink when generating the cel expressions from the `sink-filters` ConfigMap. Additionally the operator should validate all cel expressions in KubeArchiveConfigs during reconciliation and refuse to create/update a KubeArchiveConfig if any of the cel expressions in it are invalid

PR #286, opened 2024-07-30
Sink archives and deletes objects based on CEL expressions

### LINKED ISSUE #238: Sink should read filters out of the ConfigMap
Related to #237 #222 #221 

The sink needs to mount the Filters ConfigMap, and read the filters out of it. Hopefully it can be parse in yaml format. Then we can have a struct that can be used to load the data into. There should be a separate thread that checks if there is a change to the mounted configmap and updates the filter struct.

The filter struct will need to be thread safe. We can use a RWMutex for that.

**Acceptance Criteria**:

- the sink can read the filters out the of the configmap
- the sink can update the filter data when the configmap is updated

skoved:
Sink archives and deletes resources on the cluster depending on if they
cause a cel expression to return to true. The cel expressions that are run
depend on the namespace and apiVersion of the k8s object

Resolves #238

## Release Note

```release-note
NONE
```

## Notes for Reviewers

- Sink creates cel expressions from the `sink-filters` ConfigMap that is managed by the operator
- Sink mounts the ConfigMap so that every key-value pair in the ConfigMap's data becomes a file that will be updated if the ConfigMap changes
- Sink can handle invalid cel expressions. If one or more cel expressions are not valid the sink will log the error and skip the expression
- The sink can handle the mounted ConfigMap files being unreadable with out crashing
- The sink can handle the ConfigMap not being mounted without crashing

WIP:
- Updating the cel expressions when the ConfigMap changes

**Important**: If you test this change, you will see the following error in the sink logs

```bash
ERROR: <input>:1:9: found no matching overload for '_==_' applied to '(map(string, dyn), string)'
| .status == 'Completed'
| ........^
```

This is a bug in the operator code. I created #285 to fix it. This error does not impact the operation of the sink

To test:

- install KubeArchive following instructions in DEVELOPMENT.md including initializing the database
- run `kubectl apply -f https://storage.googleapis.com/tekton-releases/pipeline/latest/release.yaml`
- run `kubectl create ns demo-tekton`
- run

```bash
kc apply -f - << EOF
---
kind: KubeArchiveConfig
metadata:
  name: kubearchive
  namespace: demo-tekton
spec:
  resources:
    - selector:
        apiVersion: tekton.dev/v1beta1
        kind: PipelineRun
      archive-when: .status.startTime != ''
      delete-when: .status.completionTime != ''
---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: demo-tekton
  namespace: demo-tekton
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: pipelineruns-reader
rules:
- apiGroups: ["tekton.dev"]
  resources: ["pipelineruns"]
  verbs: ["get", "list"]
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: read-pipelineruns
subjects:
- kind: ServiceAccount
  name: demo-tekton
  namespace: demo-tekton
  apiGroup: ""
roleRef:
  kind: ClusterRole
  name: pipelineruns-reader
EOF
```

- run `kubectl get -n kubearchive pods` and copy the name of the `kubearchive-sink` pod
- run `kubectl delete pod -n kubearchive <kubearchive-sink pod> (this is needed because the sink won't create new filters when the ConfigMap updates
- in a separate terminal run `kubectl logs -n kubearchive -l app=kubearchive-sink -f
- in the first terminal run `kubectl apply -f https://raw.githubusercontent.com/skoved/kubearchive/first-demo/tekton.yaml`
- run `kubectl apply -f https://raw.githubusercontent.com/skoved/kubearchive/first-demo/tekton.yaml`

### DIFF
--- a/cmd/sink/expr/cel.go
+++ b/cmd/sink/expr/cel.go
@@ -1,0 +1,211 @@
+// Copyright KubeArchive Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package expr
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"os"
+
+	"github.com/google/cel-go/cel"
+	"github.com/google/cel-go/checker/decls"
+	celext "github.com/google/cel-go/ext"
+	"github.com/kubearchive/kubearchive/cmd/operator/api/v1alpha1"
+	"github.com/kubearchive/kubearchive/pkg/files"
+	"gopkg.in/yaml.v3"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+const globalKey = "_global"
+
+type Filters struct {
+	archive map[string]cel.Program
+	delete  map[string]cel.Program
+}
+
+// Returns a Filters struct with an empty archive and delete slice
+func EmptyFilters() *Filters {
+	return &Filters{
+		archive: make(map[string]cel.Program),
+		delete:  make(map[string]cel.Program),
+	}
+}
+
+// Creates a Filters struct from the path to a directory where a ConfigMap was mounted. If path is empty string or path
+// does not exist, it returns a Filters struct with empty archive and delete slices. It will attempt to create all the
+// cel programs that it can from the ConfigMap. Any errors that are encountered are wrapped together and returned. Even
+// if this function returns an error, the Filters struct returned can still be used and will not be nil.
+func NewFilters(path string) (*Filters, error) {
+	var errList []error
+	filters := &Filters{
+		archive: make(map[string]cel.Program),
+		delete:  make(map[string]cel.Program),
+	}
+	exists, err := files.PathExists(path)
+	if err != nil {
+		return filters, fmt.Errorf("cannot determine if ConfigMap is mounted at path %s: %s", path, err)
+	}
+	if path == "" || !exists {
+		return filters, fmt.Errorf("cannot create Filters. ConfigMap is not mounted at path %s", path)
+	}
+	filterFiles, err := files.DirectoryFiles(path)
+	if err != nil {
+		return filters, fmt.Errorf(
+			"cannot create Filters. Could not read files created from ConfigMap mounted at path %s: %s",
+			path,
+			err,
+		)
+	}
+	for namespace, filePath := range filterFiles {
+		resourcesBytes, err := os.ReadFile(filePath)
+		if err != nil {
+			errList = append(errList, err)
+			continue
+		}
+		resources := []v1alpha1.KubeArchiveConfigResource{}
+		err = yaml.Unmarshal(resourcesBytes, &resources)
+		if err != nil {
+			errList = append(errList, err)
+			continue
+		}
+		for _, resource := range resources {
+			namespaceGvk := NamespaceGroupVersionKind(namespace, resource)
+			archiveExpr, err := CreateCelExpr(resource.ArchiveWhen)
+			if err != nil {
+				errList = append(errList, err)
+			} else {
+				filters.archive[namespaceGvk] = *archiveExpr
+			}
+			deleteExpr, err := CreateCelExpr(resource.DeleteWhen)
+			if err != nil {
+				errList = append(errList, err)
+			} else {
+				filters.delete[namespaceGvk] = *deleteExpr
+			}
+		}
+	}
+
+	return filters, errors.Join(errList...)
+
+}
+
+// Returns whether obj needs to be archived. Obj needs to be archived if any of the cel programs in archive return true
+// or if obj needs to be deleted. If obj is nil, it returns false
+func (f *Filters) MustArchive(ctx context.Context, obj *unstructured.Unstructured) bool {
+	if obj == nil {
+		return false
+	}
+	mustArchive := f.MustDelete(ctx, obj)
+	if mustArchive {
+		return mustArchive
+	}
+	globalGvk := GlobalGroupVersionKind(obj)
+	namespaceGvk := NamespaceGroupVersionKindFromObject(obj)
+	program, exists := f.archive[globalGvk]
+	if exists {
+		mustArchive = executeCel(ctx, program, obj)
+	}
+	if mustArchive {
+		return mustArchive
+	}
+	program, exists = f.archive[namespaceGvk]
+	if exists {
+		mustArchive = executeCel(ctx, program, obj)
+	}
+	return mustArchive
+}
+
+// Returns whether obj needs to be deleted. Obj needs to be deleted if the cel program in delete that matches the
+// NamespaceGroupVersionKind of obj or the cel program for the GlobalGroupVersionKind in delete for obj returns true. If
+// obj is nil, it returns false
+func (f *Filters) MustDelete(ctx context.Context, obj *unstructured.Unstructured) bool {
+	if obj == nil {
+		return false
+	}
+	mustDelete := false
+	globalGvk := GlobalGroupVersionKind(obj)
+	namespaceGvk := NamespaceGroupVersionKindFromObject(obj)
+	program, exists := f.delete[globalGvk]
+	if exists {
+		mustDelete = executeCel(ctx, program, obj)
+	}
+	if mustDelete {
+		return mustDelete
+	}
+	program, exists = f.delete[namespaceGvk]
+	if exists {
+		mustDelete = executeCel(ctx, program, obj)
+	}
+	return mustDelete
+}
+
+// Returns a string that is the form <globalKey>:GroupVersionKind
+func GlobalGroupVersionKind(obj *unstructured.Unstructured) string {
+	if obj == nil {
+		return ":"
+	}
+	return fmt.Sprintf("%s:%s", globalKey, obj.GetObjectKind().GroupVersionKind())
+}
+
+// Returns a string that is the form <namespace>:GroupVersionKind
+func NamespaceGroupVersionKindFromObject(obj *unstructured.Unstructured) string {
+	if obj == nil {
+		return ":"
+	}
+	return fmt.Sprintf("%s:%s", obj.GetNamespace(), obj.GetObjectKind().GroupVersionKind())
+}
+
+// Returns a string that is the form <namespace>:GroupVersionKind
+func NamespaceGroupVersionKind(namespace string, resource v1alpha1.KubeArchiveConfigResource) string {
+	gvk := schema.FromAPIVersionAndKind(resource.Selector.APIVersion, resource.Selector.Kind)
+	return fmt.Sprintf("%s:%s", namespace, gvk)
+}
+
+func CreateCelExpr(expr string) (*cel.Program, error) {
+	mapStrDyn := decls.NewMapType(decls.String, decls.Dyn)
+	env, err := cel.NewEnv(
+		celext.Strings(),
+		celext.Encoders(),
+		celext.Sets(),
+		celext.Lists(),
+		celext.Math(),
+		cel.Declarations(
+			decls.NewVar("metadata", mapStrDyn),
+			decls.NewVar("spec", mapStrDyn),
+			decls.NewVar("status", mapStrDyn),
+		),
+	)
+	if err != nil {
+		return nil, err
+	}
+	parsed, issues := env.Parse(expr)
+	if issues != nil && issues.Err() != nil {
+		return nil, issues.Err()
+	}
+	checked, issues := env.Check(parsed)
+	if issues != nil && issues.Err() != nil {
+		return nil, issues.Err()
+	}
+	program, err := env.Program(checked)
+	if err != nil {
+		return nil, err
+	}
+	return &program, err
+}
+
+// executes the cel program with obj as the input. If the program returns and error or if the program returns a value
+// that cannot be converted to bool, false is returned. Otherwise the value returned by the cel program is returned
+func executeCel(ctx context.Context, program cel.Program, obj *unstructured.Unstructured) bool {
+	val, _, err := program.ContextEval(ctx, obj.Object)
+	if err != nil {
+		return false
+	}
+	boolVal, ok := val.Value().(bool)
+	if !ok {
+		return false
+	}
+	return boolVal
+}
--- a/cmd/sink/main.go
+++ b/cmd/sink/main.go
@@ -13,7 +13,7 @@ import (
 	ceOtelObs "github.com/cloudevents/sdk-go/observability/opentelemetry/v2/client"
 	cloudevents "github.com/cloudevents/sdk-go/v2"
 	ceClient "github.com/cloudevents/sdk-go/v2/client"
-	jsonpath "github.com/kubearchive/kubearchive/cmd/sink/jsonPath"
+	"github.com/kubearchive/kubearchive/cmd/sink/expr"
 	"github.com/kubearchive/kubearchive/pkg/database"
 	"github.com/kubearchive/kubearchive/pkg/models"
 	kaObservability "github.com/kubearchive/kubearchive/pkg/observability"
@@ -25,24 +25,28 @@ import (
 )
 
 const (
-	DeleteWhenEnvVar = "DELETE_WHEN"
+	mountPathEnvVar = "MOUNT_PATH"
 )
 
 type Sink struct {
 	Db          database.DBInterface
-	DeleteWhen  string
 	EventClient ceClient.Client
+	Filters     *expr.Filters
 	K8sClient   *dynamic.DynamicClient
 	Logger      *log.Logger
 }
 
-func NewSink(db database.DBInterface, logger *log.Logger) *Sink {
+func NewSink(db database.DBInterface, logger *log.Logger, filters *expr.Filters) *Sink {
 	if logger == nil {
 		logger = log.New(os.Stderr, "", log.LstdFlags|log.Lmicroseconds|log.LUTC)
-		logger.Println("sink was provided a nil logger, falling back to defualt logger")
+		logger.Println("sink was provided a nil logger, falling back to default logger")
 	}
 	if db == nil {
 		logger.Fatalln("cannot start sink when db connection is nil")
+	}
+	if filters == nil {
+		logger.Println("sink was provided nil filters. Archive and delete operations will not happen until new filters are provided")
+		filters = expr.EmptyFilters()
 	}
 
 	httpClient, err := cloudevents.NewHTTP(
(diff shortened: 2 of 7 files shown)


cmd/sink/expr/cel.go
```diff
@@ -0,0 +1,211 @@
...
+				errList = append(errList, err)
+			} else {
+				filters.delete[namespaceGvk] = *deleteExpr
+			}
+		}
+	}
+
+	return filters, errors.Join(errList...)
+
+}
```
beerparty, line 93: You probably should not have lines in a method after the `return`, even if they are empty.

```suggestion
	return filters, errors.Join(errList...)
}
```

cmd/sink/expr/cel.go
```diff
@@ -0,0 +1,211 @@
...
+		mustArchive = executeCel(ctx, program, obj)
+	}
+	if mustArchive {
+		return mustArchive
+	}
+	program, exists = f.archive[namespaceGvk]
+	if exists {
+		mustArchive = executeCel(ctx, program, obj)
+	}
+	return mustArchive
```
ggallen, line 118: Why are you evaluating the global archive and the namespace archive expressions separately?  You can just `or` them and do one evaluation.

cmd/sink/expr/cel.go
```diff
@@ -0,0 +1,211 @@
...
+	}
+	globalGvk := GlobalGroupVersionKind(obj)
+	namespaceGvk := NamespaceGroupVersionKindFromObject(obj)
+	program, exists := f.archive[globalGvk]
+	if exists {
+		mustArchive = executeCel(ctx, program, obj)
+	}
+	if mustArchive {
+		return mustArchive
+	}
```
beerparty, line 113: Since the second conditional is only relevant if the first conditional is true, it should be in the block:

```suggestion
	if exists {
		mustArchive = executeCel(ctx, program, obj)
	    if mustArchive {
		    return mustArchive
	    }
	}
```

cmd/sink/expr/cel.go
beerparty, line 118: This does not match the description. The description says "If obj is nil, it returns false" - but here you will return `nil` if `mustArchive` does not get set by `executeCel`. You need another conditional:

```suggestion
	if exists {
		mustArchive = executeCel(ctx, program, obj)
	    if mustArchive {
		    return mustArchive
	    }
	}
	return false
```

beerparty (changes requested):

cmd/sink/expr/cel.go
```diff
@@ -0,0 +1,211 @@
...
+	return fmt.Sprintf("%s:%s", globalKey, obj.GetObjectKind().GroupVersionKind())
+}
+
+// Returns a string that is the form <namespace>:GroupVersionKind
+func NamespaceGroupVersionKindFromObject(obj *unstructured.Unstructured) string {
+	if obj == nil {
+		return ":"
+	}
+	return fmt.Sprintf("%s:%s", obj.GetNamespace(), obj.GetObjectKind().GroupVersionKind())
+}
```
ggallen, line 159: These two methods should be combined into one common method like this:
```
func NamespaceGroupVersionKindFromObject(namespace string, obj *unstructured.Unstructured) string {
	if obj == nil {
		return ":"
	}
	return fmt.Sprintf("%s:%s", namespace, obj.GetObjectKind().GroupVersionKind())
}
```
On second thought, why are they even different?  Just make the global unstructured object have the namespace name in it and use the second method.

cmd/sink/main.go
```diff
@@ -87,38 +85,48 @@ func (sink *Sink) Receive(ctx context.Context, event cloudevents.Event) {
...
-			sink.DeleteWhen,
+			"failed to write object %s from cloudevent %s to the database: %s\n",
+			k8sObj.GetUID(),
 			event.ID(),
 			err,
 		)
+		return
 	}
-	if mustDelete {
+	sink.Logger.Printf("successfully wrote object %s cloudevent %s to the database\n", k8sObj.GetUID(), event.ID())
```
beerparty, line 110: Missing "from"

```suggestion
	sink.Logger.Printf("successfully wrote object %s from cloudevent %s to the database\n", k8sObj.GetUID(), event.ID())
```

cmd/sink/expr/cel.go
ggallen: This would all be much cleaner if you only evaluate one expression for both archive and delete. What needs to happen is that the `global` expressions for archive and delete need to be `or`'d to each of the namespace expressions. Then evaluate those.

cmd/sink/main.go
```diff
@@ -132,7 +140,14 @@ func main() {
 	if err != nil {
 		logger.Fatalf("Could not connect to the database: %s\n", err)
 	}
-	sink := NewSink(db, logger)
+	filters, err := expr.NewFilters(os.Getenv(mountPathEnvVar))
+	if err != nil {
+		log.Printf(
+			"Not all filters could be created from the ConfigMap. Some archive and delete operations will not execute until the errors are resolved: %s\n",
```
beerparty, line 146: Shouldn't this be "logger" instead of "log"?

All the other logging messages start with a lower case word, why is "Not" capitalized here?

ggallen: I was going to ask the opposite. Why aren't we using normal english, and capitalizing the start of messages?

cmd/sink/expr/cel.go
skoved: there is no "global" unstructured object. The difference between `GlobalGroupVersionKind` and `NamespaceGroupVersionKindFromObject` is that `GlobalGroupVersionKind` will return `_global:<GroupVersionKind of obj>` where `NamespaceGroupVersionKindFromObject` will use the namespace of the object instead (`<namespace of obj>:<GroupVersionKind of obj>`)

cmd/sink/expr/cel.go
skoved: After check if the global CEL expression and the namespaced CEL expression exist I could do

```
return executeCel(ctx, globalProgram obj) || executeCel(ctx, namespacedProgram, obj)
```

If you instead mean, taking the string representations of the CEL expressions and doing something like

```
fmt.Sprintf("%s || %s", namespacedCelExprStr, globalCelExprStr)
```

and then compiling that, is it makes updates more difficult. When you mount a configmap every key under `Data` becomes a file and value for that key is the content of that file. 

```
...
Data:
  _global: status.completedTimestamp != ''
```
becomes a file named `_global` and the content of that file is `status.completedTimestamp != ''`

With having the global expressions separate I can treat every file the same. If it is updated, I read in its content and do `yaml.Unmarshal` to convert it into `[]KubeArchiveConfigResource`. If `or` the global expressions of a specific `Kind` to every namespaced expression of that `Kind` I now need to

1. treat `_global` differently than every other file, which will make the live updating logic even more complicated than it already is
2. find every namespace expression that matches the `Kinds` found in `_global` and recompile **every single one of those CEL expressions** which will pause the execution of all cloudevents the sink receives while this is happening

cmd/sink/main.go
skoved: yes, that should be `logger`. I fixed that. I also capitalized all of the log messages

cmd/sink/main.go
skoved: fixed

cmd/sink/expr/cel.go
ggallen: The `_global` namespace should look exactly like every other namespace in the `ConfigMap`.  When it's read, its contents should be passed to `yaml.Unmarshal` and converted to a `[]KubebeArchiveConfigResource`.

Here are the steps you can go through:

1. Convert the global structure to a map of `Kind` to expression.  Or, if possible, convert it and add it to the filter, as the filter is already a map of `Kind` to CEL program.
1. Process all the other namespaces.  For each `Kind` value in the `[]KubebeArchiveConfigResource`, see if there is a match in `_global`. If so, `or` it in right then.  What you or it with depends on what works. If you need the string, then create the map described in the previous step.  If you can `or` another CEL program, then use the converted value already in the filter.

> With having the global expressions separate I can treat every file the same. 

This makes no sense to me.  I think you are already treating `_global` differently (and it's one of those files) if you aren't using `yaml.Unmarshall` on it and converting it to a filter.

cmd/sink/expr/cel.go
ggallen: > With having the global expressions separate I can treat every file the same. 

Why not? This should be *identical* to every other namespace in the `ConfigMap`. You should be able to read it in and run `yaml.Unmarshal` on it, and there you go.

cmd/sink/main.go
ggallen: > I fixed that. I also capitalized all of the log messages

We should standardize on this if that's what we want to do.

cmd/sink/expr/cel.go
skoved: The global stuff is going away

cmd/sink/expr/cel.go
skoved: This was more straight forward to code so I did it this way to see if this all would work. I'm working on formatting the expression like this

if the `_global` contains a`KubeArchiveConfigResource` has a kind that matches the kind of a namespaced KubeArchiveConfigResource they will be `or`ed in the following way (making sure that `deleteWhen`s only get `or`ed with `deleteWhen`s. The same with `archiveWhen`s)

```bash
( <global cel expression> ) || ( <namespaced cel expression> )

File: cmd/sink/expr/cel.go (hunk 1 above)
Comment (skoved) [reply cmd/sink/expr/cel.go]: fixed

File: cmd/sink/expr/cel.go (hunk 3 above)
Comment (skoved) [reply cmd/sink/expr/cel.go]: I changed the logic for creating filters which allowed me to remove this logic

File: cmd/sink/expr/cel.go (hunk 2 above)
Comment (skoved) [reply cmd/sink/expr/cel.go]: This logic has been implemented. PTAL

File: cmd/sink/main.go [hunk 7]
```diff
@@ -132,9 +140,16 @@ func main() {
 	if err != nil {
 		logger.Fatalf("Could not connect to the database: %s\n", err)
 	}
-	sink := NewSink(db, logger)
+	filters, err := expr.NewFilters(os.Getenv(mountPathEnvVar))
+	if err != nil {
+		logger.Printf(
```
rh-hemartin, line 145: If there are errors while loading our main feature, why we should load at all? If there are errors upon reading a change on the ConfigMap, then its another story.

cmd/sink/main.go
```diff
@@ -25,24 +25,28 @@ import (
...
-		logger.Println("sink was provided a nil logger, falling back to defualt logger")
+		logger.Println("Sink was provided a nil logger, falling back to default logger")
 	}
 	if db == nil {
-		logger.Fatalln("cannot start sink when db connection is nil")
+		logger.Fatalln("Cannot start sink when db connection is nil")
+	}
+	if filters == nil {
+		logger.Println("Sink was provided nil filters. Archive and delete operations will not happen until new filters are provided")
+		filters = expr.EmptyFilters()
```
rh-hemartin, line 49: This won't be true because the expr.NewFilters always return a filters object, even if there are errors.

cmd/sink/expr/cel.go
```diff
@@ -0,0 +1,312 @@
...
+// does not exist, it returns a Filters struct with empty archive and delete slices. It will attempt to create all the
+// cel programs that it can from the ConfigMap. Any errors that are encountered are wrapped together and returned. Even
+// if this function returns an error, the Filters struct returned can still be used and will not be nil.
+func NewFilters(path string) (*Filters, error) {
+	var errList []error
+	filters := &Filters{
+		archive: make(map[string]cel.Program),
+		delete:  make(map[string]cel.Program),
+		path:    path,
+	}
```
rh-hemartin, line 57: I think you should use `EmptyFilters` here

cmd/sink/expr/cel.go
```diff
@@ -0,0 +1,312 @@
...
+var ErrNoGlobal = errors.New("no global expressions exist")
+
+type Filters struct {
+	*sync.RWMutex
+	archive map[string]cel.Program
+	delete  map[string]cel.Program
+	path    string
+}
+
+// Returns a Filters struct with an empty archive and delete slice
```
maruiz93, line 38: IINM, conventionally the go doc comments good practices suggest to use the name of the exported function in the beginning of the doc:
```suggestion
// EmptyFilters returns a Filters struct with an empty archive and delete slice
```
See examples in the [Go Doc Comments Guide](https://go.dev/doc/comment#func)

cmd/sink/main.go
skoved: I agree, but this is more of a defensive thing. Since the method takes in a `*Filter` it is possible a bug is introduced that causes filters to be `nil` here. Having the sink log this instead of crashing will make discovering the bug easier

cmd/sink/main.go
skoved: I disagree. Since the ConfigMap doesn't get created until users create a `KubeArchiveConfig`, it might be the case that this function returns an error because the ConfigMap has not been mounted when the sink starts. I'm working on logic right for `filters` to be able to update itself when files in the directory where the `ConfigMap` is mounted change. The filtering logic can handle the sink having no filters. It will log why it doesn't have any filters and it will log when it receives cloudevents but nothing will get archived or deleted

cmd/sink/main.go
rh-hemartin: If its important enough to create a check here, I think its important enough to have it tested. If we test that the method does not return a `nil` no matter what, we don't need this check. If this does not convince you, then keep it as is.

cmd/sink/expr/cel.go
```diff
@@ -0,0 +1,312 @@
...
+	if obj == nil {
+		return false
+	}
+	mustDelete := false
+	gvk := NamespaceGroupVersionKindFromObject(obj)
+	program, exists := f.delete[gvk]
+	if exists {
+		mustDelete = executeCel(ctx, program, obj)
+	}
+	return mustDelete
```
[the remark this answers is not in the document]
skoved: I think you're right

cmd/sink/expr/cel.go
```diff
@@ -0,0 +1,312 @@
...
+	mustArchive := f.mustDelete(ctx, obj)
+	if mustArchive {
+		return mustArchive
+	}
+	gvk := NamespaceGroupVersionKindFromObject(obj)
+	program, exists := f.archive[gvk]
+	if exists {
+		mustArchive = executeCel(ctx, program, obj)
+	}
+	return mustArchive
```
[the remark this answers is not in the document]
skoved: I think you're right. I didn't think to do something like this because the trade off being made is compactness of the code for ease of following the boolean logic. If you don't think this is too complicated I can update this one too

cmd/sink/main.go
ggallen: The operator should create the `ConfigMap` if it doesn't exist as part of reconciliation.

But we should create the `ConfigMap` to make sure it's there, empty, at the start. I'll create an issue for this and add it to the operator.

The issue is https://github.com/kubearchive/kubearchive/issues/300.

cmd/sink/expr/cel.go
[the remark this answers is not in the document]
ggallen: I do not. To me it's much cleaner.

cmd/sink/expr/cel.go
```diff
@@ -0,0 +1,312 @@
...
+func (f *Filters) createFilters(namespace, file string, globalArchive, globalDelete map[string]string) error {
+	resourcesBytes, err := os.ReadFile(file)
+	if err != nil {
+		return err
+	}
+	resources := []v1alpha1.KubeArchiveConfigResource{}
+	err = yaml.Unmarshal(resourcesBytes, &resources)
+	if err != nil {
+		return err
+	}
```
ggallen, line 127: This code to read the file and also similar code in `getGlobalCelExprs` can be changed to call a common function to do it:
```
func readResources(filename string) ([]v1alpha1.KubeArchiveConfigResource, error) {
        resources := []v1alpha1.KubeArchiveConfigResource{}
        resourcesBytes, err := os.ReadFile(filename)
        if err != nil {
                return resources, err
        }
        err = yaml.Unmarshal(resourcesBytes, &resources)
        if err != nil {
                return resources, err
        }
        return resources, nil
}       
```
The block of code above could then be changed to simply be:
```
        readResources(file)
```
And similar changes can be made in `getGlobalCelExprs`.

ggallen (changes requested):

cmd/sink/expr/cel.go
skoved: done

cmd/sink/expr/cel.go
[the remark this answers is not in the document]
skoved: done

cmd/sink/expr/cel.go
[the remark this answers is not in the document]
skoved: done

cmd/sink/expr/cel.go
skoved: done

cmd/sink/expr/cel.go
skoved: done. I created `KACResourceSliceFromFile` and put it in `cmd/sink/k8s/kinds.go`

beerparty (approved):

cmd/sink/expr/filters.go
```diff
@@ -0,0 +1,331 @@
...
+			}
+			logger.Println("Encountered an error while watching for changes to filters:", err)
+		}
+	}
+}
+
+// handleFsEvent handles the logic for updating filters when an fsnotify.Event is received.
+func (f *Filters) handleFsEvent(logger *log.Logger, event fsnotify.Event) {
+	switch event.Op {
+	case fsnotify.Create:
```
ggallen, line 204: Why aren't `fsnotify.Create` and `fsnotify.Write` the same?  I don't think there should be a difference in what is done in both cases.

There may some code from delete that could also be shared.

ggallen (changes requested):

cmd/sink/expr/filters.go
skoved: `fsnotify.Create` and `fsnotify.Write` are the if the global filters change.

However, they are different if namespaced filters change. This is because if we receive `fsnotify.Create`, we know that there are not any filters for that namespace so we can just create filters for that namespace and add the them to `filters.archive` and `filters.delete`. If we receive `fsnotify.Write` that means we already have filters for that namespace in `filters.archive` and `filters.delete` therefore, we have to remove the old filters from both maps before we add the new ones. This ensures that only filters that are currently defined in the ConfigMap are used by the sink. I thought about combining them, but unfortunately the logic have to use for deleting the entries requires looping over every key in both `f.archive` and `f.delete`. To speed up the runtime for creating filters for a new namespace, I kept the logic separate

skoved: You're right, the logic is the same, I'll make a new method that they both call

skoved: merged the cases for `fsnotify.Create` and `fsnotify.Write`

cmd/sink/expr/cel.go
```diff
@@ -0,0 +1,81 @@
...
+	val, _, err := program.ContextEval(ctx, obj.Object)
+	if err != nil {
+		return false
+	}
+	boolVal, ok := val.Value().(bool)
+	if !ok {
+		return false
+	}
+	return boolVal
+}
```
ggallen, line 81: Couldn't this be simplified to:
```
        return ok && boolVal
```

ggallen (approved):

cmd/sink/expr/cel.go
skoved: done

PR #302, opened 2024-07-31
Operator needs to provide an empty sink-filters ConfigMap.

### LINKED ISSUE #300: Operator needs to provide an empty sink-filters ConfigMap
The operator needs to include an empty sink-filters `ConfigMap`.  It will create one during reconciliation if it it doesn't exist, but that won't happen until a `KubeArchiveConfig` resource is created and reconciled.  The sink needs one to mount.

ggallen:

Resolves #300

## Release Note

```release-note
NONE
```

### DIFF
--- a/charts/kubearchive/templates/operator/sink_filters.yaml
+++ b/charts/kubearchive/templates/operator/sink_filters.yaml
@@ -1,0 +1,5 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: sink-filters


charts/kubearchive/templates/operator/sink_filters.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
```
beerparty, line 1: You are missing the YAML header.

```suggestion
---
apiVersion: v1
```

charts/kubearchive/templates/operator/sink_filters.yaml
beerparty, line 1: Should this file (and the other templates) have the copyright and license notice?

charts/kubearchive/templates/operator/sink_filters.yaml
ggallen: Fixed.

charts/kubearchive/templates/operator/sink_filters.yaml
ggallen: That I don't know.  But I don't think any have them, and I think if we want to add them, then that is a different issue.

skoved (approved):

charts/kubearchive/templates/operator/sink_filters.yaml
beerparty: We should discuss with the group. I will resolve this since it is a bigger issue.

beerparty (approved):

Issue #307, opened 2024-08-01
Add OpenTelemetry instrumentation to the operator-sdk


ggallen:
There is no OpenTelemetry instrumented library for the operator-sdk. The basics for instrumenting the operator are in https://github.com/kubearchive/kubearchive/issues/236. This issue is to complete that work and instrument the operator-sdk ourselves.

rh-hemartin:
We couldn't go with the autoinstrumentation because it needs privileged containers. While this may be achieved under some circumstances, I do not think this would be allowed on many clusters thus limiting the usage of this feature or directly limiting the adoption of KubeArchive. Instead we used the same approach we use in the sink: adding the otelhttp custom transport in the chain of the http client used by the operator.

rh-hemartin:
We are currently instrumenting the operator as we do with the other components, I'm closing this.

PR #309, opened 2024-08-01
Implement OpenTelemetry in the operator.

### LINKED ISSUE #236: Implement OpenTelemetry in the operator.
OpenTelemetry needs to be added to the operator as is done with the other KubeArchive components.

ggallen:

Resolves #236

## Release Note

```release-note
NONE
```

## Notes for Reviewers

Note that this PR only includes changes to add instrumentation to the the operator.  The operator-sdk itself has no instrumented library, so this PR just lays the groundwork for work to be done in https://github.com/kubearchive/kubearchive/issues/307.

As a result, the changes here are minimal.

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -6,6 +6,7 @@ package main
 import (
 	"crypto/tls"
 	"flag"
+	"log"
 	"os"
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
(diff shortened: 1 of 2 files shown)


cmd/operator/main.go
```diff
@@ -42,6 +48,11 @@ func init() {
 }
 
 func main() {
+	err := observability.Start(otelServiceName)
+	if err != nil {
+		log.Printf("Could not start OpenTelemetry: %s", err)
```
beerparty, line 53: "Printf" needs a return at the end of the line:

```suggestion
		log.Printf("Could not start OpenTelemetry: %s\n", err)
```

beerparty (changes requested):

cmd/operator/main.go
```diff
@@ -23,11 +24,16 @@ import (
...
 	"github.com/kubearchive/kubearchive/cmd/operator/internal/controller"
+	"github.com/kubearchive/kubearchive/pkg/observability"
 
 	sourcesv1 "knative.dev/eventing/pkg/apis/sources/v1"
 	//+kubebuilder:scaffold:imports
 )
 
+const (
+	otelServiceName = "kubearchive.operator"
+)
```
maruiz93, line 35: As there is only one constant maybe you can put it like:

```suggestion
const otelServiceName = "kubearchive.operator"
```

maruiz93 (approved):

skoved (approved):

cmd/operator/main.go
ggallen: Fixed.

cmd/operator/main.go
ggallen: Fixed.

PR #321, opened 2024-08-12
add retries to the database connection

### LINKED ISSUE #242: Add retries for all database operations
Instead of failing if a database operator could not be performed successfully, it should be retried.

There are go modules that provide retry functionality that we can use. See #241 for the list

mafh314:

Fixes #242 

## Release Note

```release-note
NONE
```

## Notes for Reviewers

- Added retries to the database connection using the library: [retry-go](https://github.com/avast/retry-go)
- The library uses, by default, exponential back-off.
- Added an integration test to check the retry mechanism.
- Manually tested deploying kuberarchive from scratch (no restart in the sink and neither in the API)
- Manually tested write in the database after retries + connection.

### DIFF
--- a/pkg/database/database.go
+++ b/pkg/database/database.go
@@ -7,8 +7,11 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
+	"log"
+	"time"
 
 	"github.com/XSAM/otelsql"
+	"github.com/avast/retry-go/v4"
 	"github.com/kubearchive/kubearchive/pkg/models"
 	_ "github.com/lib/pq"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -37,13 +40,28 @@ func NewDatabase() (*Database, error) {
 	if err != nil {
 		return nil, err
 	}
-	db, err := otelsql.Open("postgres", dataSource)
-	if err != nil {
-		return nil, err
+	var db *sql.DB
+	configs := []retry.Option{
+		retry.Attempts(10),
+		retry.OnRetry(func(n uint, err error) {
+			log.Printf("Retry request %d, get error: %v", n+1, err)
+		}),
+		retry.Delay(time.Second),
 	}
-	if err = db.Ping(); err != nil {
-		return nil, err
+
+	errRetry := retry.Do(
+		func() error {
+			db, err = otelsql.Open("postgres", dataSource)
+			if err != nil {
+				return err
+			}
+			return db.Ping()
+		},
+		configs...)
+	if errRetry != nil {
+		return nil, errRetry
 	}
+	log.Println("Successfully connected to the database")
 	return &Database{db, resourceTableName}, nil
 }
 
--- a/go.mod
+++ b/go.mod
@@ -8,6 +8,7 @@ require (
 	github.com/DATA-DOG/go-sqlmock v1.5.2
 	github.com/Masterminds/semver/v3 v3.2.1
 	github.com/XSAM/otelsql v0.32.0
+	github.com/avast/retry-go/v4 v4.6.0
 	github.com/cloudevents/sdk-go/observability/opentelemetry/v2 v2.15.2
 	github.com/cloudevents/sdk-go/v2 v2.15.2
 	github.com/fsnotify/fsnotify v1.7.0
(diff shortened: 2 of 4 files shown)


pkg/database/database.go
```diff
@@ -32,18 +35,33 @@ type Database struct {
...
-		return nil, err
+
+	errRetry := retry.Do(
+		func() error {
+			db, err = otelsql.Open("postgres", dataSource)
+			if err = db.Ping(); err != nil {
+				return err
+			}
+			return err
+		},
```
beerparty, line 59: Having both returns the same (`return err`) is confusing, I suggest simplifying it to something like this:

```suggestion
			db, err = otelsql.Open("postgres", dataSource)
			if err != nil {
				return db.Ping()
			}
			return nil
		},
```

beerparty (commented):
One suggestion to improve the readability of the code.

pkg/database/database.go
```diff
@@ -32,18 +35,33 @@ type Database struct {
...
-	db, err := otelsql.Open("postgres", dataSource)
-	if err != nil {
-		return nil, err
+	var db *sql.DB
+	configs := []retry.Option{
+		retry.Attempts(10),
+		retry.OnRetry(func(n uint, err error) {
+			log.Printf("Retry request %d, get error: %v", n+1, err)
+		}),
+		retry.Delay(time.Second),
```
beerparty, line 49: The delay should be longer than a single second. I would suggest at least 5 seconds between each retry. I don't know if it is supported in the library you are using, but I think an exponential back-off would be appropriate here.

mafh314: I'm sorry, I should add to the PR notes that, by default, the library uses exponential back-off.
These are some traces from the integration-test
`2024/08/12 15:17:46 2024/08/12 **13:17:03** Retry request 1, get error: dial tcp 10.96.117.213:5432: connect: connection timed out
2024/08/12 **13:17:04** Retry request 2, get error: dial tcp 10.96.117.213:5432: connect: connection refused
2024/08/12 **13:17:06** Retry request 3, get error: dial tcp 10.96.117.213:5432: connect: connection refused
2024/08/12 **13:17:10** Retry request 4, get error: dial tcp 10.96.117.213:5432: connect: connection refused
2024/08/12 **13:17:18** Successfully connected to the database
`

pkg/database/database.go
ggallen: That  confuses me too. Are you saying, @beerparty , that if the open fails you then want to do a ping?  I would assume you would only ping if the open was successful.

Why ping if the open was successful?

mafh314: The code that you mentioned, it was copied directly from the [go documentation](https://go.dev/doc/database/open-handle#confirm_connection). I've tried an approach as you mentioned but the [RetryableFunc](https://pkg.go.dev/github.com/avast/retry-go?utm_source=godoc#RetryableFunc) that I'm using has to return an error.
But, I think a I can do something with the errors that I'm returning. what do you think?
```
func() error {
    db, err = otelsql.Open("postgres", dataSource)
    if errPing := db.Ping(); errPing != nil {
        return errPing
    }
    return err
}
```

rh-hemartin: Regardless of what the documentation says... shouldn't we error on `otelsql.Open` first, and then try to ping and then error on the ping? It makes sense to me.

Also, from the documentation:

> Your code should check for an error from attempting to create a handle, such as with sql.Open. This won’t be a connection error. Instead, you’ll get an error if sql.Open was unable to initialize the handle. This could happen, for example, if it’s unable to parse the DSN you specified.

maruiz93: according to the previous comments maybe something like this it's more clear and should do the same:
```suggestion
			db, err = otelsql.Open("postgres", dataSource)
			if err != nil {
				return err
			}
			return db.Ping()
		},
```

pkg/database/database.go
```diff
@@ -32,18 +35,33 @@ type Database struct {
 	resourceTableName string
 }
 
-func NewDatabase() (*Database, error) {
+func GetDatabaseConnection() (*Database, error) {
```
maruiz93, line 38: I am not sure about this change of name, I've usually seen that in go the "constructors" are called New<StructName> in this case, this function instantiates a new instance of the struct Database declared above, so, I would keep the name `NewDatabase()` instead of `GetDatabaseConnection()`

Plus, the name `GetDatabaseConnection` suggests me that we are retrieving a DB connection that has been already established, and that is not the case, we intend to create the sql.DB instance once (as it is a connection pool) and keep it until the program ends.

maruiz93 (changes requested):

pkg/database/database.go
beerparty: You don't need a comment on the PR. If Delay does exponential back-off by default that is fine.

pkg/database/database.go
ggallen: It's still not clear to me why we need to call `db.Ping` if the open is successful.  That sort of goes against any pattern I have seen, ever.

ggallen: Never mind, I finally read the referenced documentation and it explains why.

mafh314: @maruiz93 I'm gonna test your suggestion :) Thanks!
@ggallen I understood that the ping is a way to confirm that you can connect to the db when you will not use the connection immediately. 
@rh-hemartin you're right.

mafh314: Changed and tested.

beerparty (approved):

pkg/database/database.go
mafh314: I saw  our code and we are using the same convention in all the places. So, I will change then the name of this function to the old one.

skoved (approved):

Issue #271, opened 2024-07-29
Create database schema SQL file for release and other changes


ggallen:
The database should be creating using an SQL file that a user can simply run using `psql`. This SQL script should do the following:

1. Create the `kubearchive` database.
2. Create all the tables, indexes and any other constraints in the `kubearchive` database.
3. Create a `kubearchive` user who is the owner of the `kubearchive` database. This user should be used for all KubeArchive database connections.
4. Set the password for the `kubearchive` user to some default value. This value will be recommended to be changed by the user as installation time.

The best way to create this script is to create the database as desired, then dump it using the `pg_dump` utility. This also sets a number of other options on the database by default.  

When changes are needed in the future, simply use `psql` to create the database using the SQL script. Then make the changes to the database. Next, use `pg_dump` to generate a new SQL script. Finally, compare this new script to the committed version, and if the changes are as expected, commit it.

Part of this work should also extract the default password and create a secret for it in the `kubearchive` namespace. This secret should be mounted by whatever processes need to access the database (sink, etc). At KubeArchive installation time, the user should provide the password for the `kubearchive` database user so the installation process can create the needed secret correctly.

Additionally, the database connection information (URL, username, etc) should be stored in a `ConfigMap` and read by database users such as the sink.

maruiz93:
In PostgreSQL there is also the `schema` (different concept from `database`, a `database` can allocate several schemas, more information [here](https://www.sprinkledata.com/blogs/postgresql-schema-vs-database-unveiling-the-distinctions-for-effective-data-management#:~:text=PostgreSQL%20Schema%20vs.-,Database%3A%20Key%20Differences,and%20RENAME%20to%20modify%20data.)).

 I included  a `kubearchive` schema in the PR https://github.com/kubearchive/kubearchive/pull/269/files based on the issue https://github.com/kubearchive/kubearchive/issues/258.
 
 You didn't mention the `schema` but are you ok with creating both database AND schema?

skoved:
> Additionally, the database connection information (URL, username, etc) should be stored in a ConfigMap and read by database users such as the sink.

Right now, the helm chart creates the username, url, and password as a secret and the sink and api server use those secrets to create environment variables

PR #328, opened 2024-08-14
Release as single YAML

### LINKED ISSUE #277: Release should publish a single YAML for installation
Currently we are using Helm to install and we agreed to use a single YAML for installation instead. The release process needs to change so it does not try to publish the Helm chart, but it generates the single YAML using `helm template` and then publishes it with the `gh create release` command.

Note: the single YAML needs to be processed by `ko` so it publishes and replaces the image references, so probably the final comand looks like `helm template | ko resolve -f - > kubearchive.yaml`

### LINKED ISSUE #278: Release should publish the DDL
Currently we are extracting the PostgreSQL definitions from the `database/init_db.go` script to a `ddl.sql` file (#258 ). When that happens, that file should be published on the release and users will be directed to it in the installation process.

rh-hemartin:

Resolves #277, resolves #278, may be related to #271

## Release Note

```release-note
NONE
```

## Notes for Reviewers

* Enabled the database deployment by default for development, as the Helm chart is now development oriented I think it makes sense.
* Moved POSTGRES_URL to the secret, should I move the port too?
* Moved the testSA to its own file on the `test` folder so its easier to toggle.
* Removed braces rules: they are not needed anymore and they introduce artifacts in the resulting release YAML.
* Introduced the creation of a namespace for production, so we don't need to ask users to `kubectl create ns kubearchive`

### DIFF
--- a/charts/kubearchive/templates/_helpers.tpl
+++ b/charts/kubearchive/templates/_helpers.tpl
@@ -19,3 +19,19 @@ SPDX-License-Identifier: Apache-2.0
 - name: OTEL_EXPORTER_OTLP_ENDPOINT
   value: "{{ $endpoint }}"
 {{- end -}}
+
+{{/* Override database information when database is deployed */}}
+{{- define "kubearchive.v1.database.env" -}}
+{{- $enabled := .Values.database.enabled -}}
+{{- if $enabled -}}
+POSTGRES_URL: {{ .Values.database.url | b64enc | quote }}
+POSTGRES_DB: {{ .Values.database.postgresData.dbName | b64enc | quote }}
+POSTGRES_USER: {{ .Values.database.postgresData.dbUser | b64enc | quote }}
+POSTGRES_PASSWORD: {{ .Values.database.postgresData.dbPassword | b64enc | quote }}
+{{- else -}}
+POSTGRES_URL: ""
+POSTGRES_DB: ""
+POSTGRES_USER: ""
+POSTGRES_PASSWORD: ""
+{{- end -}}
+{{- end -}}
--- a/charts/kubearchive/values.yaml
+++ b/charts/kubearchive/values.yaml
@@ -3,6 +3,11 @@
 # Declare variables to be passed into your templates.
 
 ---
+global:
+  # If 'true' it excludes resources such as the `test` namespace or the test service account
+  # used for testing API endpoints.
+  production: false
+
 # values used to create the api-server
 apiServer:
   name: "{{ .Release.Name }}-api-server"
@@ -60,12 +65,14 @@ operator:
 database:
   # Controls if the database is deployed or not. The secret with the credentials
   # is deployed always, as the sink and the api-server need to connect somewhere.
-  # Provide the proper `url` and credentials if you provide your database
-  # TODO: set 'enabled' to 'false' when we start distributing the chart for production installation
+  # If 'enabled: true' the secret credentials are automatically set, if you want to
+  # provide your own database, set 'enabled: false' and modify the secret after deploying
   enabled: true
 
   name: "{{ .Release.Name }}-database"
+
   url: "{{ tpl .Values.database.name . }}.{{ .Release.Namespace }}.svc.cluster.local"
+
   # number of database pods that should be deployed
   replicaCount: 1
   # database container image
@@ -86,11 +93,12 @@ database:
   # database basic configuration
   postgresData:
     # database name
-    dbName: "postgresdb"
+    dbName: "kubearchive"
     # database user
-    dbUser: "ps_user"
+    dbUser: "kubearchive"
     # database password
     dbPassword: "P0stgr3sdbP@ssword"  # notsecret
+
 
 integrations:
   observability:
--- a/database/init_db.go
+++ b/database/init_db.go
@@ -19,8 +19,8 @@ type flags struct {
 }
 
 var defaultValues = &flags{
-	DatabaseName:     "postgresdb",
-	DatabaseUser:     "ps_user",
+	DatabaseName:     "kubearchive",
+	DatabaseUser:     "kubearchive",
 	DatabasePassword: "P0stgr3sdbP@ssword", // notsecret
 }
 
--- a/hack/install-database.sh
+++ b/hack/install-database.sh
@@ -1,0 +1,109 @@
+#!/bin/bash
+# Copyright KubeArchive Authors
+# SPDX-License-Identifier: Apache-2.0
+# WARNING: this is intended to run in a single node
+#   Kubernetes environment and just for testing purposes
+
+set -o errexit
+set -o xtrace
+
+kubectl apply -f - <<EOF
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: databases
+---
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  labels:
+    app: postgresql
+  name: postgresql
+  namespace: databases
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 5Gi
+---
+apiVersion: v1
+kind: PersistentVolume
+metadata:
+  labels:
+    type: local
+    app: postgresql
+  name: postgresql
+spec:
+  capacity:
+    storage: 5Gi
+  accessModes:
+    - ReadWriteOnce
+  hostPath:
+    path: /data/postgresql
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    app: postgresql
+  name: postgresql
+  namespace: databases
+spec:
+  selector:
+    matchLabels:
+      app: postgresql
+  template:
+    metadata:
+      labels:
+        app: postgresql
+    spec:
+      containers:
+        - name: postgresql
+          image: postgres:16
+          ports:
+            - containerPort: 5432
+          volumeMounts:
+            - name: postgresql
+              mountPath: /var/lib/postgresql/data
+          env:
+            - name: POSTGRES_DB
+              value: root
+            - name: POSTGRES_USER
+              value: root
+            - name: POSTGRES_PASSWORD
+              value: password  # notsecret
+      volumes:
+        - name: postgresql
+          persistentVolumeClaim:
+            claimName: postgresql
+---
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: postgresql
+  name: postgresql
+  namespace: databases
+spec:
+  type: ClusterIP
+  ports:
+    - port: 5432
+  selector:
+    app: postgresql
+
+EOF
+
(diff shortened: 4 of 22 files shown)


hack/install-database.sh
```diff
@@ -0,0 +1,68 @@
...
+set -o errexit
+set -o xtrace
+
+kubectl apply -f - <<EOF
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: databases
+---
```
beerparty, line 12: Why is this YAML embedded in a bash script instead of being in a separate file?

hack/install-database.sh
```diff
@@ -0,0 +1,68 @@
...
+    app: postgresql
+
+EOF
+
+kubectl rollout status deployment --namespace=databases --timeout=30s
+sleep 15  # Wait for PostgreSQL to be really ready
+
+echo "CREATE USER kubearchive WITH ENCRYPTED PASSWORD 'P0stgr3sdbP@ssword';" | kubectl exec -i -n databases $(kubectl get -n databases pod -o name) -- psql -h localhost
+echo "CREATE DATABASE kubearchive WITH OWNER kubearchive;" | kubectl exec -i -n databases $(kubectl get -n databases pod -o name) -- psql -h localhost
+cat database/ddl.sql | kubectl exec -i -n databases $(kubectl get -n databases pod -o name) -- psql -h localhost --user=kubearchive kubearchive
```
beerparty, line 63: You have three instances here of `$(kubectl get -n databases pod -o name)` - you should consider running it once and saving the result in a variable name to use in these commands.

hack/install-database.sh
```diff
@@ -0,0 +1,68 @@
...
+
+kubectl port-forward -n databases svc/postgresql 5432:5432 &
```
beerparty, line 65: In the previous `kubectl` commands you calculate the name, here it is hard coded as "svc/postgresql". Why?

beerparty (changes requested):

hack/quick-install.sh
```diff
@@ -26,12 +28,21 @@ data:
 EOF
 
 bash cmd/operator/generate.sh
+helm template kubearchive charts/kubearchive -n kubearchive \
+    --include-crds \
+    --set "global.production=true" \
+    --set "database.enabled=false" > /tmp/kubearchive-not-resolved.yaml
+ko resolve -f /tmp/kubearchive-not-resolved.yaml --base-import-paths > kubearchive.yaml
+kubectl apply -n kubearchive -f kubearchive.yaml
```
beerparty, line 36: I noticed that the files you are creating in this script are going into `/tmp` except for `kubearchive.yaml`. Is keeping the `kubearchive.yaml` file in the current directory after running the script intentional? I ask because the `release.sh` script removes the `kubearchive.yaml` file when the script is complete.

hack/install-database.sh
rh-hemartin: Because its specific and simple enough that it can embedded. I could extract it to the `hack` folder, but I think it will clutter the folder.

This exists because in the CI we will be testing the production deployment so the database is provided by the user. This script mimics the user providing the database.

hack/install-database.sh
rh-hemartin: Because the other `kubectl` commands were pointing to a Pod deployed by a Deployment. Those pods get numbers with a short uid (for example `postgresql-57d5dd655f-bxmpz`) so I need to retrieve the pod name dynamically. However I will use a label to filter which pods can be returned as it will better indicate the intention.

In this `kubectl` command I'm pointing to a service that has a fixed name.

hack/quick-install.sh
rh-hemartin: Its not really intentional, as there is no point of keeping it in the current directory. I will send it to `/tmp/` too. If someone wants to inspect it, they can open it directly from `/tmp/`.

hack/install-database.sh
rh-hemartin: Done

beerparty (approved):

mafh314:
@rh-hemartin 
About
> Moved POSTGRES_URL to the secret, should I move the port too?
I searching for references about this, and I've found that it's recommended to have in the secret the DB url, DB user, DB pass and the DB name, but nothing about the port.

hack/install-database.sh
```diff
@@ -0,0 +1,69 @@
+#!/bin/bash
```
mafh314, line 1: I think we need to add to this file the copyright paragraph.

hack/quick-install.sh
```diff
@@ -3,6 +3,8 @@
 set -o errexit
 set -o xtrace
 
+bash hack/install-database.sh
```
mafh314, line 6: Also in this file, we need to add the copyright paragraph.

hack/install-database.sh
```diff
@@ -0,0 +1,69 @@
...
+  type: ClusterIP
+  ports:
+    - port: 5432
+  selector:
+    app: postgresql
+
+EOF
+
+kubectl rollout status deployment --namespace=databases --timeout=30s
+sleep 15  # Wait for PostgreSQL to be really ready
```
mafh314, line 59: I'm not sure, but maybe 15 sec is not enough :(

rh-hemartin: I tested this multiple times and it is, its not enough when you tested this PR?

hack/install-database.sh
```diff
@@ -0,0 +1,69 @@
...
+
+set -o errexit
+set -o xtrace
+
+kubectl apply -f - <<EOF
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: databases
```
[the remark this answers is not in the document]
rh-hemartin: The CI is more production like, so it could make sense that the DB uses a different namespace. We don't need to have the database anywhere special, as long as its accessible.

I noticed that this causes an issue with the `TestDatabaseConnection` test, I'm working on it.

hack/quick-install.sh
rh-hemartin: Added

hack/install-database.sh
mafh314: I've just tested. My comment was based on my previous tests, when I had to wait more than 15 sec to have the DB working. 
My results from testing the script are that the 15 sec are enough. I tested also deploying the DB after delete the postgresql:16 image, to force the pull.

hack/install-database.sh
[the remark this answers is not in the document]
rh-hemartin: After some back and forth with a couple ideas, I just changed the namespace and name of the deployment being manipulated to the one on `databases` and added a PV and PVC to the database deployment, so the initialization (so the user and password created for KubeArchive) is retained.

mafh314 (approved):

charts/kubearchive/values.yaml
```diff
@@ -53,12 +58,14 @@ operator:
...
-  # Provide the proper `url` and credentials if you provide your database
-  # TODO: set 'enabled' to 'false' when we start distributing the chart for production installation
+  # If 'enabled: true' the secret credentials are automatically set, if you want to
+  # provide your own database, set 'enabled: false' and modify the secret after deploying
   enabled: true
 
   name: "{{ .Release.Name }}-database"
-  url: "{{ tpl .Values.database.name . }}.{{ .Release.Namespace }}.svc.cluster.local"
+
+  url: "kubearchive-database.kubearchive.svc.cluster.local"
```
maruiz93, line 67: Why are we including this change?

According to the way that [Kubernetes set up the DNS for the different services](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#services), this name is built through the following template:

<service-name>.<service-namespace>.svc.cluster.local

If we change the namespace provided in the `helm template`/`helm install` command (e.g. if we want to deploy 2 different kubearchive charts for developing/debugging/troubleshooting purposes), we need to remember also updating this line.

database/init_db.go
```diff
@@ -19,8 +19,8 @@ type flags struct {
 }
 
 var defaultValues = &flags{
-	DatabaseName:     "postgresdb",
-	DatabaseUser:     "ps_user",
+	DatabaseName:     "kubearchive",
+	DatabaseUser:     "kubearchive",
```
maruiz93, line 23: These changes won't apply anymore after https://github.com/kubearchive/kubearchive/pull/346/files PR

hack/install-database.sh
```diff
@@ -0,0 +1,107 @@
...
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 5Gi
+---
+apiVersion: v1
+kind: PersistentVolume
+metadata:
+  labels:
+    type: local
```
maruiz93, line 33: The pods that aim to use local PVs need to be allocated where the PV lives.
This works well when this is deployed on a one-worker-node Kubernetes cluster, otherwise it could randomly fail.

I suggest to do one of the following:
* Add a comment stating that this should be only used in single-worker-node kubernetes cluster.
* Add some constraints to make this work for multinode clusters:
  * Create a `StorageClass` with the `volumeBindingMode: WaitForFirstConsumer`
  * Update the `Deployment ` of the database to include all the pods under the same node, something along these lines:
  
```yaml
      spec:
      affinity:
        podAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
          - labelSelector:
              matchExpressions:
              - key: app
                operator: In
                values: 
                - postgresql
            topologyKey: "kubernetes.io/hostname"
```

maruiz93 (changes requested):

database/init_db.go
rh-hemartin: Yes, I am aware. If mine goes first then they are needed, if Andrea's go first this file will disappear when I rebase.

hack/install-database.sh
rh-hemartin: I don't think going the extra mile with the affinity (and its testing) is worth it as this is intended for CI, so I added a comment at the top of the script.

charts/kubearchive/values.yaml
rh-hemartin: Ops, I did oversee this change when reviewing my own changes. I reverted it so there is no change here.

charts/kubearchive/templates/_helpers.tpl
```diff
@@ -19,3 +19,19 @@ SPDX-License-Identifier: Apache-2.0
 - name: OTEL_EXPORTER_OTLP_ENDPOINT
   value: "{{ $endpoint }}"
 {{- end -}}
+
+{{/* Override database information when database is deployed */}}
+{{- define "kubearchive.v1.database.env" -}}
+{{- $enabled := .Values.database.enabled -}}
+{{- if (eq $enabled true) -}}
```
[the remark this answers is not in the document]
rh-hemartin: Right, I changed it.

maruiz93 (approved):

skoved (approved):

ggallen (approved):

Issue #294, opened 2024-07-31
Add tracing to the DEVELOPMENT.md


rh-hemartin:
Now we have a telemetry deployment available for us in development, we need to document it.

The process to enable telemetry is:

1. After installing knative eventing, create the following ConfigMap:
    ```
    kubectl apply -f - <<EOF
    apiVersion: v1
    kind: ConfigMap
    metadata:
      name: config-tracing
      namespace: knative-eventing
    data:
      backend: "zipkin"
      zipkin-endpoint: "http://traces-collector.kubearchive.svc.cluster.local:9411/api/v2/spans"
      sample-rate: "1"
    EOF
    ```
1. Pass `--set "integrations.observability.enabled=true"` to helm's install command.
1. Port forward the Jaeger port: `kubectl port-forward -n kubearchive svc/traces-collector 16686:16886 &`

After generating traffic in the sink or API we should see traces there. The traces from the sink are wrapped into apiserversource traces, as we receive a POST call and then we react to it.

PR #329, opened 2024-08-14
How to enable traces on development

### LINKED ISSUE #295: Add Zipkin endpoint to the Jaeger deployment
Currently we are deploying a Jaeger all-in-one pod to collect traces. We need Knative Eventing to emit traces so the sink also emit traces (see #233 for more information on this). Knative only supports zipkin as an endpoint to emit traces, but jaeger is compatible with that. However that endpoint is not currently enabled in our jaeger deployment, so we need to enable it.

See the Getting Started Jaeger documentation. There are examples of deployments with the Zipkin endpoint enabled: https://www.jaegertracing.io/docs/1.6/getting-started/

rh-hemartin:

Resolves #295 and #294 

## Release Note

```release-note
NONE
```

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -215,6 +215,39 @@ to start a debugger to which attach from your IDE.
    https://localhost:8081/apis/batch/v1/jobs | jq
    ```
 
+## Enabling Telemetry
+
+We use Jaeger for tracing on development. As some dependencies use the Zipkin format
+to send traces, we are using the zipkin-compatible endpoint of Jaeger.
+
+1. Create the following ConfigMap to enable Telemetry for Knative Eventing.
+    ```bash
+    kubectl apply -f - <<EOF
+    apiVersion: v1
+    kind: ConfigMap
+    metadata:
+      name: config-tracing
+      namespace: knative-eventing
+    data:
+      backend: "zipkin"
+      zipkin-endpoint: "http://traces-collector.kubearchive.svc.cluster.local:9411/api/v2/spans"
+      sample-rate: "1"
+    EOF
+    ```
+    **Note**: Knative's APIServerSource created before applying this change do not emit traces. Recreate
+    the KubeArchiveConfig to trigger the recreation of the APIServerSource.
+1. Install or upgrade KubeArchive adding the following option to the Helm command:
+    ```bash
+    --set "integrations.observability.enabled=true"
+    ```
+    This extra flag makes Helm deploy a traces collector and sets the appropiate environment
+    variables for the KubeArchive deployments.
+1. Forward the traces collector port to localhost:
+    ```bash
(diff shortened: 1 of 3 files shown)


DEVELOPMENT.md
```diff
@@ -215,6 +215,36 @@ to start a debugger to which attach from your IDE.
...
+    ```bash
+    --set "integrations.observability.enabled=true"
+    ```
+    This extra flag makes Helm deploy a traces collector and sets the appropiate environment
+    variables for the KubeArchive deployments.
+1. Forward the traces collector port to localhost:
+    ```bash
+    kubectl port-forward -n kubearchive svc/traces-collector 16686:16686 &
+    ```
+1. Open [http://localhost:16686](http://localhost:16686) in your browser.
```
beerparty, line 230: What does this do? Is it to login or is there some other purpose to connecting to this URL? I ask because I wonder if this step is just for viewing the traces. If it is just for viewing the traces, this step should be after the step for enabling Telemetry for Knative Eventing.

DEVELOPMENT.md
```diff
@@ -215,6 +215,36 @@ to start a debugger to which attach from your IDE.
    https://localhost:8081/apis/batch/v1/jobs | jq
    ```
 
+## Enabling Telemetry
+
+1. Install or upgrade KubeArchive adding the following option to the Helm command:
```
beerparty, line 220: It looks like this is the first introduction to "zipkin". Could you include a sentence or two here to explain what it is?

```suggestion
## Enabling Telemetry
Zipkin is an open-source unified tracing system. It is used here so each development instance
can gather tracing data without having to set up a centralized data collection system.

1. Install or upgrade KubeArchive adding the following option to the Helm command:
```

beerparty (changes requested):

DEVELOPMENT.md
rh-hemartin: Its just to open the Jaeger UI to search for traces. I can move it, no problems.

DEVELOPMENT.md
rh-hemartin: We are not using Zipkin (the tool), we are using Jaeger. Knative just emits traces using the zipkin protocol, and we use the zipkin compatible endpoint of Jaeger to receive the traces. I can add a generic line about tracing and why we are using tracing if you want.

beerparty: Could you include your explanation in the comment? I think it would be helpful to explain why we are using "zipkin" in the telemetry configuration.

Something like:

"The development instance uses Jaeger to gather telemetry traces. The code uses the zipkin compatible endpoint of Jaeger to receive the traces."

rh-hemartin: Done, check it out.

beerparty (approved):

PR #331, opened 2024-08-14
Refactor CEL code to the operator.

### LINKED ISSUE #330: Refactor CEL code to the operator.
The operator needs to be able to compile the CEL code as part of validating the `KubeArchiveConfig` resource (see https://github.com/kubearchive/kubearchive/issues/311). Currently the CEL code resides in the sink, and the sink depends on the operator for the type definition of the expressions used for filters.

When trying to add the call to validate the CEL expressions in the operator, and import cycle is created.  The operator tries to call the CEL code in the sink, and then the sink turns around and tries to access the operator code to get the needed type definition.

This can be resolved by refactoring the sink code and moving the CEL expression code into the operator code.

ggallen:

Resolves #330

## Release Note

```release-note
NONE
```


cmd/operator/cel/cel.go
skoved, line 1: I understand why `cel.go` needs to move out of `cmd/sink/filters`, but I don't agree with moving it to `cmd/operator`. I think the correct place for it is `pkg/cel` since both the sink and the operator depend on this code. Like how the sink and the api share `pkg/database` for doing database operations.

skoved (changes requested):

cmd/operator/cel/cel.go
ggallen: Fair point. I've moved it.

cmd/sink/filters/filters.go
```diff
@@ -15,7 +15,8 @@ import (
 
 	"github.com/fsnotify/fsnotify"
 	"github.com/google/cel-go/cel"
-	"github.com/kubearchive/kubearchive/cmd/sink/k8s"
+	kubearchivev1alpha1 "github.com/kubearchive/kubearchive/cmd/operator/api/v1alpha1"
```
beerparty, line 18: This label includes the version "v1alpha1", are you going to change this name, and all the references, to something else when we release "beta" and "v2" code? It seems like it would make more sense to call this "kubearchiveapi", "operatorapi", or just "oapi" so when we change the operator version we only have to change one line in this file.

ggallen: Good point. I've done that.

rh-hemartin (approved):

beerparty (approved):

PR #340, opened 2024-08-20
Cache AuthN and AuthZ requests

### LINKED ISSUE #194: Cache TokenReview and SubjectAccessReview queries
Currently the REST API Server is querying k8s API in each call once for the TokenReview and once for the SubjectAccessReview.

We should include a memory cache to prevent hammering the k8s API.

rh-hemartin:

Resolves #194 

## Release Note

```release-note
NONE
```

### DIFF
--- a/charts/kubearchive/templates/api_server/api_server.yaml
+++ b/charts/kubearchive/templates/api_server/api_server.yaml
@@ -49,6 +49,10 @@ spec:
             - name: POSTGRES_PORT
               value: "{{ required "A Postgres port must be specified" .Values.database.service.port }}"
             {{- include "kubearchive.v1.otel.env" . | nindent 12 }}
+            - name: CACHE_EXPIRATION_AUTHENTICATED
+              value: "{{ .Values.apiServer.cache.expirationAuthenticated }}"
+            - name: CACHE_EXPIRATION_UNAUTHENTICATED
+              value: "{{ .Values.apiServer.cache.expirationUnauthenticated }}"
 ---
 kind: Service
 apiVersion: v1
--- a/cmd/api/main.go
+++ b/cmd/api/main.go
@@ -7,11 +7,14 @@ import (
 	"fmt"
 	"log"
 	"net/http"
+	"os"
+	"time"
 
 	"github.com/gin-gonic/gin"
 	"github.com/kubearchive/kubearchive/cmd/api/auth"
 	"github.com/kubearchive/kubearchive/cmd/api/discovery"
 	"github.com/kubearchive/kubearchive/cmd/api/routers"
+	"github.com/kubearchive/kubearchive/pkg/cache"
 	"github.com/kubearchive/kubearchive/pkg/database"
 	"github.com/kubearchive/kubearchive/pkg/observability"
 	"go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin"
@@ -20,17 +23,26 @@ import (
 	"k8s.io/client-go/rest"
 )
 
+const (
+	otelServiceName                   = "kubearchive.api"
+	cacheExpirationAuthorizedEnvVar   = "CACHE_EXPIRATION_AUTHORIZED"
+	cacheExpirationUnauthorizedEnvVar = "CACHE_EXPIRATION_UNAUTHORIZED"
+)
+
 var (
 	version = "main"
 	commit  = ""
 	date    = ""
 )
 
-const otelServiceName = "kubearchive.api"
-
 type Server struct {
 	k8sClient kubernetes.Interface
 	router    *gin.Engine
+}
+
+type cacheExpirations struct {
+	Authorized   time.Duration
+	Unauthorized time.Duration
 }
 
 func getKubernetesClient() *kubernetes.Clientset {
@@ -47,11 +59,11 @@ func getKubernetesClient() *kubernetes.Clientset {
 	return client
 }
 
-func NewServer(k8sClient kubernetes.Interface, controller routers.Controller) *Server {
+func NewServer(k8sClient kubernetes.Interface, controller routers.Controller, cache *cache.Cache, cacheExpirations *cacheExpirations) *Server {
 	router := gin.Default()
 	router.Use(otelgin.Middleware("")) // Empty string so the library sets the proper server
-	router.Use(auth.Authentication(k8sClient.AuthenticationV1().TokenReviews()))
-	router.Use(auth.RBACAuthorization(k8sClient.AuthorizationV1().SubjectAccessReviews()))
+	router.Use(auth.Authentication(k8sClient.AuthenticationV1().TokenReviews(), cache, cacheExpirations.Authorized, cacheExpirations.Unauthorized))
+	router.Use(auth.RBACAuthorization(k8sClient.AuthorizationV1().SubjectAccessReviews(), cache, cacheExpirations.Authorized, cacheExpirations.Unauthorized))
 	// TODO - Probably want to use cache for the discovery client
 	// See https://pkg.go.dev/k8s.io/client-go/discovery/cached/disk#NewCachedDiscoveryClientForConfig
 	router.Use(discovery.GetAPIResource(k8sClient.Discovery().RESTClient()))
@@ -70,14 +82,49 @@ func main() {
 	if err != nil {
 		log.Printf("Could not start opentelemetry: %s", err)
 	}
+
 	db, err := database.NewDatabase()
 	if err != nil {
 		log.Fatalf("Could not connect to database: %s", err)
 	}
 	controller := routers.Controller{Database: db}
-	server := NewServer(getKubernetesClient(), controller)
+
+	cacheExpirations, err := getCacheExpirations()
+	if err != nil {
+		log.Fatal(err)
+	}
+	cache := cache.New()
+
+	server := NewServer(getKubernetesClient(), controller, cache, cacheExpirations)
 	err = server.router.RunTLS("localhost:8081", "/etc/kubearchive/ssl/tls.crt", "/etc/kubearchive/ssl/tls.key")
 	if err != nil {
 		log.Printf("Could not run server on localhost: %s", err)
 	}
+}
+
+func getCacheExpirations() (*cacheExpirations, error) {
+	expirationAuthorizedString := os.Getenv(cacheExpirationAuthorizedEnvVar)
+	if expirationAuthorizedString == "" {
+		return nil, fmt.Errorf("The environment variable '%s' should be set.", cacheExpirationAuthorizedEnvVar)
+	}
+
+	expirationAuthorized, err := time.ParseDuration(expirationAuthorizedString)
+	if err != nil {
+		return nil, fmt.Errorf("'%s': '%s' could not be parsed into a duration: %s", cacheExpirationAuthorizedEnvVar, expirationAuthorizedString, err)
+	}
+
+	expirationUnauthorizedString := os.Getenv(cacheExpirationUnauthorizedEnvVar)
+	if expirationUnauthorizedString == "" {
+		return nil, fmt.Errorf("The environment variable '%s' should be set.", cacheExpirationUnauthorizedEnvVar)
+	}
+
+	expirationUnauthorized, err := time.ParseDuration(expirationUnauthorizedString)
+	if err != nil {
+		return nil, fmt.Errorf("'%s': '%s' could not be parsed into a duration: %s", cacheExpirationUnauthorizedEnvVar, expirationUnauthorizedString, err)
+	}
+
+	return &cacheExpirations{
+		Authorized:   expirationAuthorized,
+		Unauthorized: expirationUnauthorized,
+	}, nil
 }
--- a/docs/modules/ROOT/pages/configuration/cache-expiration-time.adoc
+++ b/docs/modules/ROOT/pages/configuration/cache-expiration-time.adoc
@@ -1,0 +1,57 @@
+= Cache Expiration Time
+
+KubeArchive delegates authentication and authorization to Kubernetes. To
+avoid pressure on the Kubernetes API, KubeArchive caches the authentication
+and authorization answers.
+
+There are two different environment variables that control cache expiration:
+
+* `CACHE_EXPIRATION_AUTHORIZED`: controls the expiration time for authorized/authenticated requests.
+    If not set, the default value is **10 minutes**.
+* `CACHE_EXPIRATION_UNAUTHORIZED`: controls the expiration time for unauthorized/unauthenticated requests.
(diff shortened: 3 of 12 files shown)


beerparty (changes requested):

cmd/api/main.go
```diff
@@ -63,14 +66,31 @@ func main() {
...
+	authCacheExpiration := getCacheExpirationDuration()
+	cache, err := cache.New(authCacheExpiration)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	server := NewServer(getKubernetesClient(), controller, cache)
 	err = server.router.RunTLS("localhost:8081", "/etc/kubearchive/ssl/tls.crt", "/etc/kubearchive/ssl/tls.key")
 	if err != nil {
 		log.Printf("Could not run server on localhost: %s", err)
```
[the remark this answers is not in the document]
rh-hemartin: Yes, it needs to be a Fatalf. I think that if this fails right now the Deployment enters into a infinite loop spinning the pod, ending it with exit 0 and spinning another one again. However its not a problem currently.

cmd/api/main.go
```diff
@@ -63,14 +66,31 @@ func main() {
...
 	}
 }
+
+func getCacheExpirationDuration() string {
+	expirationString := os.Getenv(authCacheExpirationKey)
+	if expirationString == "" {
+		log.Fatalf("The environment variable 'AUTHENTICATION_CACHE_EXPIRATION' needs to be set.")
```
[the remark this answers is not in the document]
rh-hemartin: My bad, changed.

docs/modules/ROOT/pages/configuration/cache-expiration-time.adoc
```diff
@@ -0,0 +1,40 @@
+= Cache Expiration Time
+
+KubeArchive delegates authentication and authorization to Kubernetes. To
+avoid pressure on the Kubernetes API, KubeArchive caches the authentication
+and authorization answers. You can modify the expiration time of the cache.
+By default is set to 10 minutes.
```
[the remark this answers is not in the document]
rh-hemartin: Right! Changed.

beerparty (approved):

docs/modules/ROOT/pages/configuration/cache-expiration-time.adoc
```diff
@@ -0,0 +1,40 @@
...
+kubectl patch -n kubearchive deployment kubearchive-api-server --patch-file path/to/patch.yaml
+----
+
+[IMPORTANT]
+.Expiring the cache
+====
+To expire the current cache restart the `kubearchive-api-server` deployment with
+the command
+[command]`kubectl rollout -n kubearchive restart deployment kubearchive-api-server`.
+====
```
maruiz93, line 40: There is a decision being made here. I don't have strong feelings against this and we could go with what you are proposing, but I would want to raise other options in here.

This PR is currently enforcing all authZ and authN calls to be cached (success and failures) and not providing a way to enforce token re-auth.
Pros:
* Harder to get slow responses due to the requests to the k8s API
Cons:
* If the permissions change, either forbidding access or allowing it we may need to wait the expiration caching time to be able to see the changes.

There is also the possibility of only caching success requests. Then if the client was unauthorized and ask for his permissions to be changed by a cluster admin, he can automatically see the changes there. The drawback here is that it potentially increases the slowdown of the API if there are too many failed unauthenticated / unauthorized requests.

maruiz93 (approved):
I am OK with the PR, just added some comments to see if we explicitly want to make the current design decisions for now.

cmd/api/main.go
```diff
@@ -63,14 +66,31 @@ func main() {
...
 	err = server.router.RunTLS("localhost:8081", "/etc/kubearchive/ssl/tls.crt", "/etc/kubearchive/ssl/tls.key")
 	if err != nil {
 		log.Printf("Could not run server on localhost: %s", err)
 	}
 }
+
+func getCacheExpirationDuration() string {
+	expirationString := os.Getenv(authCacheExpirationKey)
+	if expirationString == "" {
+		log.Fatal("The environment variable 'AUTHENTICATION_CACHE_EXPIRATION' needs to be set.")
```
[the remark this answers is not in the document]
rh-hemartin: The default value is on the Deployment environment. We have a couple of levels and I am not sure where to put it. However I know I don't want to specify a default in multiple levels. Moving the default to code makes a typo in the environment unseen, as it won't fail. However if there is a typo, then the container fails to start.

docs/modules/ROOT/pages/configuration/cache-expiration-time.adoc
rh-hemartin: I see what you mean... we can also have different cache expiration for both but that would complicate the implementation... let me explore a little the implementations and then I will be back with a more complete answer.

rh-hemartin:
@beerparty and @maruiz93 I introduced significant changes: now there are two expiration times, one for "positive" answers and one for "negative" answers to give more control to the user. Please take a look.

cmd/api/main.go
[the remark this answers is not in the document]
rh-hemartin: I introduced two environment variables, the code is outdated, but your point still stands.

docs/modules/ROOT/pages/configuration/cache-expiration-time.adoc
rh-hemartin: I introduced one variable to control the authorized/authenticated expiration and one to control the unauthorized/unauthenticated expiration. Its not that much complex in code and mentally, so I went for it.

cmd/api/main.go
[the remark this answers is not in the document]
maruiz93: I personally prefer to have the default values in code and then be able to override them with environment variables. 
Let me complete my suggestion with another comment on the api-server YAML template

charts/kubearchive/templates/api_server/api_server.yaml
```diff
@@ -49,6 +49,10 @@ spec:
             - name: POSTGRES_PORT
               value: "{{ required "A Postgres port must be specified" .Values.database.service.port }}"
             {{- include "kubearchive.v1.otel.env" . | nindent 12 }}
+            - name: CACHE_EXPIRATION_AUTHORIZED
+              value: "10m"
+            - name: CACHE_EXPIRATION_UNAUTHORIZED
+              value: "1m"
```
maruiz93, line 55: Although this is perfectly valid, I would go with not setting the env vars by default, probably something like this:

```suggestion
           {{- if .Values.apiServer.expirationAuth }}
            - name: CACHE_EXPIRATION_AUTHORIZED
              value: "10m"
            {{- end }}
            {{- if .Values.apiServer.cache.expirationUnauth }}
            - name: CACHE_EXPIRATION_UNAUTHORIZED
              value: "1m"
            {{-end}}
```

And add a couple of comments in the values.yaml with an example of how to override the default values stating which is the default when unset.

maruiz93 (changes requested):

cmd/api/main.go
```diff
@@ -63,14 +66,56 @@ func main() {
...
+	}
+
+	expirationAuthorized, err := time.ParseDuration(expirationAuthorizedString)
+	if err != nil {
+		log.Fatalf("'%s': '%s' could not be parsed into a duration: %s", cacheExpirationAuthorizedKey, expirationAuthorizedString, err)
+	}
+
+	expirationUnauthorizedString := os.Getenv(cacheExpirationUnauthorizedKey)
+	if expirationUnauthorizedString == "" {
+		log.Fatalf("The environment variable '%s' needs to be set.", cacheExpirationAuthorizedKey)
```
maruiz93, line 109: There is a typo here:
```suggestion
		log.Fatalf("The environment variable '%s' needs to be set.", cacheExpirationUnauthorizedKey)
```

docs/modules/ROOT/pages/configuration/cache-expiration-time.adoc
maruiz93: I like your approach. I made a couple of comments of where I would put the default values, but for the rest I am ok with the changes

maruiz93 (changes requested):

charts/kubearchive/templates/api_server/api_server.yaml
rh-hemartin: I don't think that moving the default value to the Values file provides value. I think the difference between production and development comes into play here. As this is something configurable by the user in production, it makes sense that we configure it in the same way in development.

cmd/api/main.go
rh-hemartin: Good catch!

charts/kubearchive/templates/api_server/api_server.yaml
maruiz93: Let me rephrase this: 

I think it makes more sense having the default cache expiration times set in code so when no environment variables are provided, these values are in place.

Now, for how to better communicate the developers/users that this values can be overwritten:
* For the users, as you already did, having instructions for how to introduce this env vars in the Deployment
* For the developers, as we have the helm chart, I think that if we centralize this kind of stuff in the `Values.yaml` instead of having to open the different templates provides value.

rh-hemartin: I'm OK with moving the default values to the code.

I don't agree with having the option to set the values in development using the Values.yaml. I do think it contributes to distance the development experience from the production experience, which is something I think we need to keep to a minimum. However I does not bother me that the option is available to whoever wants to use it.

maruiz93: the reason I proposed to use the helm templated values is because we have helm in place

if we want to narrow the gap between experience of the user and the developer by not using helm templated values, does it makes sense to get rid of helm at some point?

rh-hemartin: I think that ideally we should be deploying in development as we deploy in production, but the world is not ideal. So I understand that helm might be useful to auto-deploy some large stuff that is not deployed on production, in our case the database deployment or a telemetry deployment. For configuration that is offered to the user I find more useful to use the production process so I can feel something similar to what they are feeling and it is what I do.

So while I think it makes sense to get rid of Helm, I don't think we want to get rid of it ever, as it is helpful to us. And given that, and given that I personally can deploy without using Helm, its fine for me to add more stuff into Helm.

cmd/api/main.go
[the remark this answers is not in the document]
rh-hemartin: Done! :+1:

charts/kubearchive/templates/api_server/api_server.yaml
maruiz93: your statement makes a lot of sense, and it actually could be a design decision that we are explicitly making for "How a user should deploy KubeArchive and why do we include Helm charts just for development"

maruiz93 (approved):

charts/kubearchive/templates/api_server/api_server.yaml
ggallen: > I think it makes more sense having the default cache expiration times set in code so when no environment variables are provided, these values are in place.

Why?  Will the container ever be run without the deployment? It seems reasonable to me, and common practice, to have environment variables with defaults.  

I don't think *everything* has to go into values.yaml.

maruiz93: > Why? Will the container ever be run without the deployment? It seems reasonable to me, and common practice, to have environment variables with defaults.

I am more comfortable with the idea of not needing environment variables and then, if you look into customization of the api, override the default values with environment variables rather than modifying them.

If I thought that the cache expiration time is going to be something that each user is going to change, I would prefer it to be more explicit, and to fail if no env variable is provided, but I think that's not the case and it makes sense to don't care about those parameters in general.

With this, the YAML provided to be deployed is shorter and cleaner.

> I don't think everything has to go into values.yaml.
I don't think that everything has to be in the `values.yaml` either, but the things that could be changed yes, and for development purposes I think that I may want to be able to change the cache expiration time, and everything I want to tweak regarding the kubearchive configuration is in one place, I like it better than scattered in different files

beerparty (approved):
LGTM

charts/kubearchive/templates/api_server/api_server.yaml
ggallen: > but the things that could be changed yes

This is where I think we disagree. Just the fact that something *could* be changed doesn't mean it has to go into the `values.yaml` file.  There are a million things that *could* be changed.

But if something is not likely to change, I don't see any value in putting it in `values.yaml`.  And I don't think these are likely to change.

IMO, this will also complicate debugging things in production.  A user would then have to go find the variable names and add them to the deployment, hopefully getting the names correct and doing it right. If they're there, he just has to change the value. You don't really need documentation for that.

maruiz93: > This is where I think we disagree. Just the fact that something could be changed doesn't mean it has to go into the values.yaml file. There are a million things that could be changed.

You are right, and as you say, it makes more sense to say that we should add to the `values.yaml` the parameters that are likely to change by the developers (as we are using helm charts for development)

But my point still stands because as I mentioned earlier, for development purposes I believe I may want to change those parameters.

> IMO, this will also complicate debugging things in production. A user would then have to go find the variable names and add them to the deployment, hopefully getting the names correct and doing it right. If they're there, he just has to change the value. You don't really need documentation for that.

About this, they are different perspectives, I prefer a good documentation an smaller kubernetes resources for things that are optional when we talk about providing things to the user, you prefer having every option explicitly set so the user won't need to look for the docs to customize optional stuff.

If you are worried about having to look for the environment variable name, we could also add a comment in the deployment with the line that should be uncommented and edited to add the env var.

```yaml
           {{- if .Values.apiServer.expirationAuth }}
            - name: CACHE_EXPIRATION_AUTHORIZED
              value: "10m"
            {{- else }}
            # Uncomment the following lines updating the value to override the default value of 10 minutes
            # Check the docs for more information
            # - name: CACHE_EXPIRATION_AUTHORIZED
            #   value: "10m"
            {{- end }}
            {{- if .Values.apiServer.cache.expirationUnauth }}
            - name: CACHE_EXPIRATION_UNAUTHORIZED
              value: "1m"
            {{- else }}
            # Uncomment the following lines updating the value to override the default value of 1 minute. 
            # Check the docs for more information
            # - name: CACHE_EXPIRATION_UNAUTHORIZED
            #   value: "1m"
            {{- end }}
```

rh-hemartin:
@maruiz93 @beerparty another more or less big change after today's meeting:

* defaults are in the values.yml
* environment variables are always populated
* api fails if not set env variables not set or they fail to parse
* introduced test

maruiz93 (approved):
LGTM

PR #341, opened 2024-08-20
Add version and commit vars to components

### LINKED ISSUE #290: Components have a variable that indicate the version
Its a good practice to include a variable that stores the version on the go binaries. At build time for release the process can pass a build flag that sets it. See https://www.digitalocean.com/community/tutorials/using-ldflags-to-set-version-information-for-go-applications and https://ko.build/configuration/#overriding-go-build-settings. Then the components will print and/or include that information in logs/traces.

rh-hemartin:

Resolves #290 

## Release Note

```release-note
NONE
```

### DIFF
--- a/.ko.yaml
+++ b/.ko.yaml
@@ -1,0 +1,5 @@
+---
+defaultLdflags:
+  - -X main.version={{ .Git.Tag }}
+  - -X main.commit={{ .Git.FullCommit }}
+  - -X main.date={{ .Date }}
--- a/cmd/api/main.go
+++ b/cmd/api/main.go
@@ -20,6 +20,12 @@ import (
 	"k8s.io/client-go/rest"
 )
 
+var (
+	version = "main"
+	commit  = ""
+	date    = ""
+)
+
 const otelServiceName = "kubearchive.api"
 
 type Server struct {
@@ -59,6 +65,7 @@ func NewServer(k8sClient kubernetes.Interface, controller routers.Controller) *S
 }
 
 func main() {
+	log.Printf("Starting KubeArchive API with version '%s', commit SHA '%s', built '%s'", version, commit, date)
 	err := observability.Start(otelServiceName)
 	if err != nil {
 		log.Printf("Could not start opentelemetry: %s", err)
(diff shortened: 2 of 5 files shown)


skoved (approved):

cmd/api/main.go
```diff
@@ -59,6 +65,7 @@ func NewServer(k8sClient kubernetes.Interface, controller routers.Controller) *S
 }
 
 func main() {
+	log.Printf("Starting KubeArchive API with version '%s', commit SHA '%s', built at '%s'", version, commit, date)
```
beerparty, line 68: "built at" would indicate a location, for a date I think this should be just "built".

```suggestion
	log.Printf("Starting KubeArchive API with version '%s', commit SHA '%s', built '%s'", version, commit, date)
```

cmd/sink/main.go
```diff
@@ -141,6 +147,8 @@ func (sink *Sink) Receive(ctx context.Context, event cloudevents.Event) {
 
 func main() {
 	logger := log.New(os.Stderr, "", log.LstdFlags|log.Lmicroseconds|log.LUTC)
+	logger.Printf("Starting KubeArchive Sink with version '%s', commit SHA '%s', built at '%s'", version, commit, date)
```
beerparty, line 150: "built at" would indicate a location, for a date I think this should be just "built".

```suggestion
	logger.Printf("Starting KubeArchive Sink with version '%s', commit SHA '%s', built '%s'", version, commit, date)
```

rh-hemartin: Changed!

.ko.yaml
```diff
@@ -0,0 +1,5 @@
+---
+defaultLdflags:
+  - -X main.version={{ .Git.Tag }}
+  - -X main.commit={{ .Git.FullCommit }}
+  - -X main.date={{ .Date }}
```
maruiz93, line 5: I like how this is done really much.
I will point to the docs in case someone else is interested on this part: https://ko.build/configuration/

rh-hemartin: Sorry I didn't wait for your review, I missed you added yourself.

PR #342, opened 2024-08-21
Change to use single ApiServerSource.

### LINKED ISSUE #338: Using single ApiServerSource for KubeArchive
KubeArchive currently creates and `ApiServerSource` in each namespace where a `KubeArchiveConfig` resource exists. This was done initially for simplicity and to allow the `ApiServerSource` to be very specific about what messages get sent to the sink, and the sink was to do no filtering.

The sink is now doing filtering via the CEL expressions that it uses to determine when to archive (`ArchiveWhen`) or delete `DeleteWhen`).   Adding global expressions means that the operator would have to update the `ApiServerSource` role in every namespace when those expressions change. This is complex and error prone.

Since the sink is already filtering messages, and taking into account the added complexity detailed above, it now makes sense to go to a single `ApiServerSource` instance in the `KubeArchive` installation namespace.  The operator can then update the configuration information for this `ApiServerSource` instance whenever a `KubeArchiveConfig` changes.

Note that all settings are additive.  So if namespace A adds settings for `Event` resources and namespace B adds settings for `PipelineRun` resources, the `ApiServerSource` will be configured to send messages for both resources from both namespaces. The sink should apply the filters correctly so that only the appropriate messages are processed for each namespace.

ggallen:

Resolves #338

## Release Note

```release-note
NONE
```


cmd/operator/internal/controller/kubearchiveconfig_controller.go
```diff
@@ -33,7 +34,11 @@ import (
...
 )
 
-const SinkFilterConfigMapName = "sink-filters"
+const (
+	SinkFilterConfigMapName   = "sink-filters"
+	ApiServerSourceLabelName  = "kubearchive.org/enabled"
+	ApiServerSourceLabelValue = "true"
+)
 
 var k9eNs = os.Getenv("KUBEARCHIVE_NAMESPACE")
```
maruiz93, line 43: Are we officially using `k9e` as an abbreviation of KubeArchive in the code?

ggallen: > Are we officially using k9e as an abbreviation of KubeArchive in the code?

I don't think there's anything official, but I think it makes this code clearer.  It was done as part of a change that was merged over a month ago, and is documented at the top of the file.

>  I think "ka" makes much more sense.

You say "toe-may-toe" and I  say "toe-ma-toe".

rh-hemartin:
Could you delete the integration test `TestKubeArchiveOperator` and the function `IsObjectReady`, also from the integration test file? That test is supposed to check that an APIServerSource is deployed on the namespace where a KubeArchiveConfig is deployed. So it should be failing for this PR. However is not working properly.

cmd/operator/internal/controller/kubearchiveconfig_controller.go
```diff
@@ -362,20 +434,48 @@ func (r *KubeArchiveConfigReconciler) reconcileA14e(ctx context.Context, kaconfi
...
+	sort.Strings(resourceKeys)
+
+	resources := make([]sourcesv1.APIVersionKindSelector, 0, len(resourceMap))
+	for _, key := range resourceKeys {
+		resources = append(resources, resourceMap[key])
 	}
+	return namespaces, resources
+}
+
+func (r *KubeArchiveConfigReconciler) desiredA14e(ctx context.Context, namespaces []string, resources []sourcesv1.APIVersionKindSelector) (*sourcesv1.ApiServerSource, error) {
```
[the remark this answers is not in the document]
ggallen: Fixed.

ggallen:
> Could you delete the integration test TestKubeArchiveOperator and the function IsObjectReady

Done.

cmd/operator/internal/controller/kubearchiveconfig_controller.go
```diff
@@ -491,3 +594,48 @@ func (r *KubeArchiveConfigReconciler) removeFilters(ctx context.Context, kaconfi
...
+	err := r.Get(ctx, types.NamespacedName{Name: kaconfig.Namespace}, ns)
+	if err == nil {
+		ns.ObjectMeta.Labels[ApiServerSourceLabelName] = ApiServerSourceLabelValue
+		err = r.Update(ctx, ns)
+		if err != nil {
+			log.Error(err, "Failed to update Namespace "+kaconfig.Namespace)
+			return ns, err
+		}
+	} else {
+		log.Error(err, "Failed to reconcile Namespace "+kaconfig.Namespace)
```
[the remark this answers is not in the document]
ggallen: Fixed.

rh-hemartin (approved):

cmd/operator/internal/controller/kubearchiveconfig_controller.go
```diff
@@ -180,29 +195,30 @@ func (r *KubeArchiveConfigReconciler) desiredA14eServiceAccount(kaconfig *kubear
 	sa := &corev1.ServiceAccount{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      a14eName,
-			Namespace: kaconfig.Namespace,
+			Namespace: k9eNs,
```
rh-hemartin, line 198: With this change, the `kaconfig` parameter is not used.

cmd/operator/internal/controller/kubearchiveconfig_controller.go
```diff
@@ -96,37 +107,43 @@ func (r *KubeArchiveConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		return ctrl.Result{}, nil
 	}
 
-	_, err := r.reconcileA14eServiceAccount(ctx, kaconfig)
+	cm, err := r.reconcileFilterConfigMap(ctx, kaconfig)
```
rh-hemartin, line 110: `kaconfig` is not used.

```suggestion
	cm, err := r.reconcileFilterConfigMap(ctx)
```

cmd/operator/internal/controller/kubearchiveconfig_controller.go
```diff
@@ -96,37 +107,43 @@ func (r *KubeArchiveConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	resources := r.parseConfigMap(ctx, cm)
+
+	_, err = r.reconcileA14eServiceAccount(ctx, kaconfig)
```
rh-hemartin, line 116: `kaconfig` is not used.

```suggestion
	_, err = r.reconcileA14eServiceAccount(ctx)
```

cmd/operator/internal/controller/kubearchiveconfig_controller.go
```diff
@@ -155,7 +170,7 @@ func (r *KubeArchiveConfigReconciler) reconcileA14eServiceAccount(ctx context.Co
 		return sa, err
 	}
 
-	err = r.Get(ctx, types.NamespacedName{Name: a14eName, Namespace: kaconfig.Namespace}, &corev1.ServiceAccount{})
+	err = r.Get(ctx, types.NamespacedName{Name: a14eName, Namespace: k9eNs}, &corev1.ServiceAccount{})
```
rh-hemartin, line 173: With this change, `kaconfig` is not used

cmd/operator/internal/controller/kubearchiveconfig_controller.go
```diff
@@ -229,97 +245,153 @@ func (r *KubeArchiveConfigReconciler) reconcileRole(ctx context.Context, kaconfi
...
+		}
+	} else {
+		log.Error(err, "Failed to reconcile ClusterRole "+roleName)
+		return role, err
+	}
+
+	return role, nil
+}
+
+func (r *KubeArchiveConfigReconciler) desiredClusterRole(ctx context.Context, kaconfig *kubearchivev1alpha1.KubeArchiveConfig, roleName string, rules []rbacv1.PolicyRule) (*rbacv1.ClusterRole, error) {
```
rh-hemartin, line 294: `kaconfig` and `error` are not used.

cmd/operator/internal/controller/kubearchiveconfig_controller.go
```diff
@@ -96,37 +107,43 @@ func (r *KubeArchiveConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+	}
+	resources := r.parseConfigMap(ctx, cm)
+
+	_, err = r.reconcileA14eServiceAccount(ctx, kaconfig)
 	if err != nil {
 		return ctrl.Result{}, err
 	}
 
-	_, err = r.reconcileA14eRole(ctx, kaconfig)
+	clusterrole, err := r.reconcileA14eRole(ctx, kaconfig, resources)
```
rh-hemartin, line 121: `kaconfig` is not used

rh-hemartin (changes requested):
There are some unused parameters, I found them adding "unparam" to my golangci-lint configuration and running `golangci-lint run` at the root of the repository.

maruiz93:
> There are some unused parameters, I found them adding "unparam" to my golangci-lint configuration and running `golangci-lint run` at the root of the repository.

I recommend configuring this locally (e.g. in vscode).
I thought of including this in our CI, but we have unused parameters that I think we should keep. For example, in the API for the testing functions that implement interfaces in a dummy way.

According to the current CI failures shown there are also unordered imports and unused imports, both should be properly updated if you run the `goimports` tool.

This can be integrated with your IDE (e.g. vscode or Goland) to be automatically done on saving.

rh-hemartin:
> > There are some unused parameters, I found them adding "unparam" to my golangci-lint configuration and running `golangci-lint run` at the root of the repository.
> 
> I recommend configuring this locally (e.g. in vscode). I thought of including this in our CI, but we have unused parameters that I think we should keep. For example, in the API for the testing functions that implement interfaces in a dummy way.
> 
> According to the current CI failures shown there are also unordered imports and unused imports, both should be properly updated if you run the `goimports` tool.
> 
> This can be integrated with your IDE (e.g. vscode or Goland) to be automatically done on saving.

I created an issue, please move your comment there so we don't side-track this PR discussion.

cmd/operator/internal/controller/kubearchiveconfig_controller.go
```diff
@@ -96,37 +107,43 @@ func (r *KubeArchiveConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
 	if err != nil {
 		return ctrl.Result{}, err
 	}
 
-	_, err = r.reconcileA14eRoleBinding(ctx, kaconfig)
+	_, err = r.reconcileA14eRoleBinding(ctx, kaconfig, clusterrole)
```
maruiz93, line 126: Just wondering and a bit offtopic for the MR
Considering A14e for "ApiServerSource", but "piServerSourc" are 13 letters, so, shouldn't be A13e?

cmd/operator/internal/controller/kubearchiveconfig_controller.go
```diff
@@ -491,3 +587,48 @@ func (r *KubeArchiveConfigReconciler) removeFilters(ctx context.Context, kaconfi
...
+
+func (r *KubeArchiveConfigReconciler) removeNamespaceLabel(ctx context.Context, kaconfig *kubearchivev1alpha1.KubeArchiveConfig) error {
+	log := log.FromContext(ctx)
+
+	log.Info("in removeNamespaceLabel")
+
+	ns := &corev1.Namespace{}
+	err := r.Get(ctx, types.NamespacedName{Name: kaconfig.Namespace}, ns)
+	if err != nil {
+		log.Error(err, "Unable to get Namespace "+kaconfig.Namespace)
```
beerparty, line 621: All other error messages of this type use "Failed to..." in the description. You should do the same here to keep the code consistent.

```suggestion
		log.Error(err, "Failed to get Namespace "+kaconfig.Namespace)
```

cmd/operator/internal/controller/kubearchiveconfig_controller.go
```diff
@@ -491,3 +587,48 @@ func (r *KubeArchiveConfigReconciler) removeFilters(ctx context.Context, kaconfi
...
+		ns.ObjectMeta.Labels[ApiServerSourceLabelName] = ApiServerSourceLabelValue
+		err = r.Update(ctx, ns)
+		if err != nil {
+			log.Error(err, "Failed to update Namespace "+kaconfig.Namespace)
+			return ns, err
+		}
+	} else {
+		log.Error(err, "Failed to get Namespace "+kaconfig.Namespace)
+		return ns, err
+	}
```
beerparty, line 608: This is inconsistent with the error handling in the rest of the code. In the rest of the code, you check for a non-nil error return. Following that coding convention here would also allow you to reduce the depth of the conditional code:

```suggestion
	if err != nil {
		log.Error(err, "Failed to get Namespace "+kaconfig.Namespace)
		return ns, err
	}
	
	ns.ObjectMeta.Labels[ApiServerSourceLabelName] = ApiServerSourceLabelValue
	err = r.Update(ctx, ns)
	if err != nil {
		log.Error(err, "Failed to update Namespace "+kaconfig.Namespace)
		return ns, err
	}
```

cmd/operator/internal/controller/kubearchiveconfig_controller.go
ggallen: Fixed.

cmd/operator/internal/controller/kubearchiveconfig_controller.go
ggallen: Fixed.

cmd/operator/internal/controller/kubearchiveconfig_controller.go

cmd/operator/internal/controller/kubearchiveconfig_controller.go

cmd/operator/internal/controller/kubearchiveconfig_controller.go

cmd/operator/internal/controller/kubearchiveconfig_controller.go

cmd/operator/internal/controller/kubearchiveconfig_controller.go
ggallen: Good catch. I've changed to A13e.  I must have mis-counted.

rh-hemartin (approved):

maruiz93 (approved):

PR #346, opened 2024-08-22
Create SQL script to initialize the database

### LINKED ISSUE #271: Create database schema SQL file for release and other changes

mafh314:

Fixes #271 

## Release Note

```release-note
NONE
```

## Notes for Reviewers

The `database/init_db.go` is replaced by `database/ddl-resource.sql`. 

I've added to the DEVELOPMENT.MD instructions to load the schema using `psql`

Tests:
* loaded the script using DBeaver.
* loaded the script using psql 
* created cloud events and stored them after the creation of the DB.
* tested the hack/quick-install.sh


DEVELOPMENT.md
```diff
@@ -113,9 +113,13 @@ helm uninstall -n kubearchive kubearchive
    ```bash
    kubectl port-forward -n kubearchive svc/kubearchive-database 5432:5432
    ```
-2. Populate the database with test objects.
+1. Run the sql script to create the `resource` table.
    ```bash
-   go run database/init_db.go
+   psql -h localhost -U kubearchive -p 5432 -d kubearchive --password -a -q -f database/init_db.sql
```
maruiz93, line 118: About the name of the file `init_db.sql`.
If we keep `dml-example.sql` I would suggest to keep `ddl.sql` for naming.

I prefer DDL, that stands for "Data Definition Language" as it's more accurate than "Initialize Database".
The reason is that "Initialize Database" could literally mean to run the [CREATE DATABASAE command](https://www.postgresql.org/docs/current/sql-createdatabase.html) that we are actually not running (we are actually expecting to have the database already created, the one that is called `kubearchive` to which we are connecting in this line.

If we change the name to `ddl.sql` this other comment won't longer be applicable but I suggest to stick with "hyphens" for naming files instead of underscores.
We are using `dml-example.sql` so I think we should have `init-db.sql` instead of `init_db.sql`

maruiz93 (changes requested):

DEVELOPMENT.md
beerparty: I second the request to use "hyphens" in the name. I also agree that the file should not be named "init-db", since there can be other files that will be run to initialize aspects of the database. My suggested name will be added as a comment on the file itself.

database/init_db.sql
beerparty, line 1: I think this file should be called `ddl-resource.sql`, since it is DDL for creating the resource table.

beerparty (changes requested):

DEVELOPMENT.md
mafh314: Hey! I renamed the sql script. Following your suggestions @maruiz93 @beerparty Thanks!

database/init_db.sql
mafh314: Changed!

DEVELOPMENT.md
```diff
@@ -115,9 +115,13 @@ kubectl delete ns kubearchive
...
    ```
-2. Populate the database with test objects.
+1. Run the sql script to create the `resource` table.
    ```bash
-   go run database/init_db.go
+   psql -h localhost -U kubearchive -p 5432 -d kubearchive --password -a -q -f database/ddl-resource.sql
+   ```
+1. Optional: insert test data in the `resource` table.
+   ```bash
+   psql -h localhost -U kubearchive --password -p 5432 -d kubearchive -a -q -f database/dml-example.sql
```
rh-hemartin, line 124: As we are deploying it on development we don't need to provide password: the container is configured so anything coming from localhost is trusted. As this invovled port-forwarding, for the container the connection is coming from the localhost device.

rh-hemartin (approved):
Small comment, no need to take action, is just a simplification.

maruiz93 (approved):

rh-hemartin (changes requested):
This needs to change the name of `ddl.sql` on the `hack/release.sh` script too.

maruiz93:
> This needs to change the name of `ddl.sql` on the `hack/release.sh` script too.

I see that this change is already included in the PR:
image

rh-hemartin:
> > This needs to change the name of `ddl.sql` on the `hack/release.sh` script too.
> 
> I see that this change is already included in the PR: image

That is the script that installs the database

> > This needs to change the name of `ddl.sql` on the `hack/release.sh` script too.
> 
> I see that this change is already included in the PR: image

That is the `hack/install-database.sh` script

mafh314:
> This needs to change the name of `ddl.sql` on the `hack/release.sh` script too.

Done @rh-hemartin

rh-hemartin (approved):

rh-hemartin:
Looks good to me, I will rebase and merge this.

PR #353, opened 2024-08-27
Add docs and update helm chart to allow  interactively debug operator

### LINKED ISSUE #324: Add instructions to debug the operator with delve
Currently we have instructions in `DEVELOPMENT.md` to enable debug for the api server. Do the same for the operator.

maruiz93:

Resolves #324

## Release Note

```release-note
NONE
```

## Notes for Reviewers

* Included a new parameter in operator values called `debug` to allow the debugging
* Refactored api-server deployment to have a more similar syntax in the args list to the operator deployment
* Moved `name` of the container in the template of the operator deployment to the top

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -10,7 +10,7 @@ Install these tools:
 
 1. [`go`](https://golang.org/doc/install)
 1. [`git`](https://help.github.com/articles/set-up-git/)
-1. [`ko`](https://github.com/google/ko)
+1. [`ko`](https://github.com/google/ko) ( >v0.16 if using `dlv` for interactive debugging )
 1. [`kubectl`](https://kubernetes.io/docs/tasks/tools/install-kubectl/)
 1. [`helm`](https://helm.sh/docs/intro/install/)
 1. [`podman`](https://podman.io/docs/installation)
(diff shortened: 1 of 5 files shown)


rh-hemartin:
`ko` supports `ko build --debug` which install `delve` on the image and does some changes so delve is used. Could you check it out?

DEVELOPMENT.md
```diff
@@ -188,29 +188,58 @@ Use `go test` to run the integration test suite:
...
+
 ## Remote IDE debugging
 
 Use [delve](https://golangforall.com/en/post/go-docker-delve-remote-debug.html)
 to start a debugger to which attach from your IDE.
 
 1. Deploy the chart with `ko` and `helm` in debug mode using an image with `delve`:
+   Specify the [component].debug variable to `true` and update the image accordingly.
+
+   * Deployment to debug the API:
```
rh-hemartin, line 210: What do you think about having one single command to deploy everything in debug mode? That should simplify the process and we would achieve the same. So it would include

```
export KO_DEFAULTBASEIMAGE=gcr.io/k8s-skaffold/skaffold-debug-support/go:latest
export KO_DISABLE_OPTIMIZATIONS=true

helm installl .... \
  --set global.debug=true \
  ...
```

DEVELOPMENT.md
```diff
@@ -188,29 +188,58 @@ Use `go test` to run the integration test suite:
...
+   ```bash
+   kubectl port-forward -n kubearchive svc/kubearchive-operator-webhooks 40000:40000
+   ```
+
 1. Enable breakpoints in your IDE.
 1. Connect to the process using the port 40000:
     * [VSCode instructions](https://golangforall.com/en/post/go-docker-delve-remote-debug.html#visual-studio-code)
     * [Goland instructions](https://golangforall.com/en/post/go-docker-delve-remote-debug.html#goland-ide)
-1. Query the API using `curl` or your browser:
+1. [Only for the API] Query the API using `curl` or your browser:
```
rh-hemartin, line 242: Do we need an instruction about how to generate traffic for the webhooks, the same as we have here for the API?

charts/kubearchive/templates/operator/operator.yaml
```diff
@@ -48,34 +48,59 @@ spec:
...
+          command:
+            - ./go/bin/dlv
+          {{- else }}
+          args:
             - --health-probe-bind-address=:8081
             - --metrics-bind-address=127.0.0.1:8080
             - --leader-elect
           command:
             - /ko-app/operator
+          {{- end}}
```
rh-hemartin, line 74:
```suggestion
          {{- end }}
```

charts/kubearchive/templates/operator/operator.yaml
```diff
@@ -48,34 +48,59 @@ spec:
...
           image: {{ .Values.operator.image }}
+          {{- if not .Values.operator.debug }}
           livenessProbe:
             httpGet:
               path: /healthz
               port: {{ $port }}
             initialDelaySeconds: 15
             periodSeconds: 20
-          name: manager
+          {{- end }}
```
rh-hemartin, line 89: If I understand this correctly it adds the livenessProbe when the operator is not on debug mode. Why we don't want the livenessProbe when debugging? It causes the pod to restart because the debugger is blocking the execution of the probe?

DEVELOPMENT.md
```diff
@@ -188,29 +188,58 @@ Use `go test` to run the integration test suite:
...
+By default `ko` uses a lightweight image to run go programs.
+The image can be [overridden](https://ko.build/configuration/#overriding-base-images) with
+the `defaultBaseImage` parameter of the `.ko.yaml` configuration file or through the
+environment variable  `KO_DEFAULTBASEIMAGE`.
+
+An example of an image with a set of debugging tools already installed that can be used is
+`registry.redhat.io/rhel9/support-tools`
+
+**[NOTE]**: Using this images usually needs the pod property `runAsNonRoot` set to `false`.
+
```
rh-hemartin, line 201: I get the feeling that this block is telling me "if you need tools in your base container for something, run this command" but it does not provide me the command.

Is this section to shed some light in the fact that we are using the `KO_DEFAULTBASEIMAGE` for debugging down below, or is this aimed to help developers that need tooling in their base image?

If it aims to help developers that need tooling I would rename the section to "running with tools in the base image" and provide a clear command. If its to shed some light on `KO_DEFAULTBASEIMAGE` I don't think we need to explain this usage of `ko`, people can search for it.

DEVELOPMENT.md
maruiz93: Good point, with the webhooks would be creating a KubeArchiveConfig resource. I'll take care of it

charts/kubearchive/templates/operator/operator.yaml
maruiz93: Yes, with this change the livenessProbe won't be available in debug mode.

I added this because we could potentially be debugging the code in `main.go` before the manager server starts

DEVELOPMENT.md
maruiz93: This section aims to help developers that need tooling, it isn't related to the usage of `KO_DEFAULTBASEIMAGE` when debugging.

I can add these commands:

```bash
export KO_DEFAULTBASEIMAGE=registry.redhat.io/rhel9/support-tools
```

and 
```
echo "defaultBaseImage: registry.redhat.io/rhel9/support-tools" > .ko.yaml 
```

to share the commands that could be used with the image that it's been recommended

DEVELOPMENT.md
maruiz93: I thought about this but I think that we usually want to debug just one of the components at a time.
Also, I think that this won't be trivial, you need to start one debugger process on the IDE per component and make sure that you use different ports per each.

Another thing that worries me for enabling debugging for all components at the same time is that this process usually consumes much more memory than just running, and that could potentially be an issue

rh-hemartin: Got it!

DEVELOPMENT.md
rh-hemartin: I think the `export` one is better as is temporary. The second one can lead to it being included into a PR and then merged by missing it.

maruiz93:
> `ko` supports `ko build --debug` which install `delve` on the image and does some changes so delve is used. Could you check it out?

I have checked this in the past and it wasn't working but is seems to be available since 'v0.16` (released 3 weeks ago) so I'll update this docs accordingly

DEVELOPMENT.md
maruiz93: On a second thought, we are already creating a KubeArchiveConfig with the test resources, and applying a second resource in a namespace that already has one could be not allowed considering https://github.com/kubearchive/kubearchive/issues/338 so I'll just add a comment without a command

DEVELOPMENT.md
```diff
@@ -188,29 +188,58 @@ Use `go test` to run the integration test suite:
...
    --set-string sink.image=$(ko build github.com/kubearchive/kubearchive/cmd/sink) \
    --set-string operator.image=$(ko build github.com/kubearchive/kubearchive/cmd/operator)
    ```
+   * Deployment to debug the Operator:
+   ```bash
+   helm install kubearchive charts/kubearchive --create-namespace -n kubearchive \
+   --set operator.debug=true \
+   --set-string apiServer.image=$(ko build github.com/kubearchive/kubearchive/cmd/api) \
+   --set-string sink.image=$(ko build github.com/kubearchive/kubearchive/cmd/sink) \
+   --set-string operator.image=$(KO_DEFAULTBASEIMAGE=gcr.io/k8s-skaffold/skaffold-debug-support/go:latest ko build --disable-optimizations github.com/kubearchive/kubearchive/cmd/operator)
```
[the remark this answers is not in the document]
maruiz93: Done

charts/kubearchive/templates/operator/operator.yaml
maruiz93: It doesn't apply anymore now that this is provided by the `--debug` option of the `ko build` command

DEVELOPMENT.md
maruiz93: Added

rh-hemartin (approved):

PR #359, opened 2024-08-28
API returns a List resource instead of JSON list

### LINKED ISSUE #228: KubeArchive API Server retrieves the data in a List resource
KubeArchive API Server currently retrieves a JSON list with the different resources stored in KubeArchive database.

As KubeArchive API Server will follow the same structure as Kubernetes API, we need to wrap the results in a "List" resource type.

This will mean following this example:

```
GET /api/v1/namespaces/test/pods
---
200 OK
Content-Type: application/json

{
  "kind": "PodList",
  "apiVersion": "v1",
  "metadata": {"resourceVersion":"10245"},
  "items": [...]
}
```

More context in [Collections docs](https://kubernetes.io/docs/reference/using-api/api-concepts/#collections)

rh-hemartin:

Resolves #228 

## Release Note

```release-note
NONE
```

### DIFF
--- a/cmd/api/routers/routers.go
+++ b/cmd/api/routers/routers.go
@@ -13,10 +13,23 @@ import (
 	"github.com/gin-gonic/gin"
 	"github.com/kubearchive/kubearchive/pkg/database"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
 
 type Controller struct {
 	Database database.DBInterface
+}
+
+// We roll our own List because we don't want to wrap resources
+// into runtime.RawExtension. We will need more fields for
+// pagination, but they are simple. If they are not, we can
+// start using corev1.List{}. See "kubectl get" code to see
+// how Lists are built.
+type List struct {
+	ApiVersion string                       `json:"apiVersion"`
+	Items      []*unstructured.Unstructured `json:"items"`
+	Kind       string                       `json:"kind"`
+	Metadata   map[string]string            `json:"metadata"`
 }
 
 // GetAllResources responds with the list of resources of a specific type across all namespaces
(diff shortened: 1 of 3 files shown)


cmd/api/routers/routers.go
```diff
@@ -30,7 +43,14 @@ func (c *Controller) GetAllResources(context *gin.Context) {
 		context.JSON(http.StatusInternalServerError, err.Error())
 	}
 
-	context.JSON(http.StatusOK, resources)
+	apiVersion := fmt.Sprintf("%s/%s", group, version)
+	listKind := fmt.Sprintf("%sList", kind)
```
maruiz93, line 47: I have some concerns about  building this variable with just appending the suffix "List" to the kind...

When a CRD is created, you could optionally provide the field `spec.names.listKind`. By default this value will be "[kind]List" but it could potentially be a different one (I have never seen a different one, but the possibility is there). With this command the `kindList` parameter for each CRD can be listed: `kubectl get crd -o jsonpath='{.items[*].spec.names.listKind}'`

At first I thought that this parameter should be provided by the [APIResource struct](https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/types.go#L1141) that we have stored in the context by the GetAPIResources middleware, but AFAICS, it isn't there...

I guess that we could query the CRD and retrieve the `kindList` field but I don't like the idea of a new query for requesting that information... so I prefer to stick with this code, but maybe we could add a comment saying that this could potentially not match the listKind of a CRD

WDYT?

rh-hemartin: Now that you mention it... Then I would return a generic `List` and see what happens, if someone complains, then we need to find a way of retrieving the information. If no one complains, then we don't need to think twice. I don't want to have another query to retrieve this either.

maruiz93: Good idea, I think that approach would be better

maruiz93 (approved):

cmd/api/routers/routers.go
ggallen: This sounds dangerous to me.  Why not just get the real list kind and never have to worry about it again?

beerparty (approved):
LGTM

PR #360, opened 2024-08-28
Add core resources endpoints

### LINKED ISSUE #197: Create an endpoint in the REST API for query core resources cluster-wide
We need an endpoint to query k8s core resources in all namespaces (following k8s REST API convention)

maruiz93:

Resolves #197

## Release Note

```release-note
NONE
```

## Notes for Reviewers

Will add the List changes of the PR https://github.com/kubearchive/kubearchive/pull/359 when it's merged

### DIFF
--- a/cmd/api/routers/routers.go
+++ b/cmd/api/routers/routers.go
@@ -4,6 +4,7 @@
 package routers
 
 import (
+	"errors"
 	"fmt"
 
 	"github.com/kubearchive/kubearchive/cmd/api/abort"
@@ -36,11 +37,17 @@ type List struct {
 func (c *Controller) GetAllResources(context *gin.Context) {
 	group := context.Param("group")
 	version := context.Param("version")
-	kind := getAPIResourceKind(context)
-	resources, err := c.Database.QueryResources(context.Request.Context(), kind, group, version)
 
+	kind, err := getAPIResourceKind(context)
 	if err != nil {
-		context.JSON(http.StatusInternalServerError, err.Error())
+		abort.Abort(context, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	resources, err := c.Database.QueryResources(context.Request.Context(), kind, group, version)
+	if err != nil {
+		abort.Abort(context, err.Error(), http.StatusInternalServerError)
+		return
 	}
 
 	context.JSON(http.StatusOK, NewList(resources))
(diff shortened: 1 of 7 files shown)


pkg/database/fake/database_generated_test.go
```diff
@@ -144,3 +144,92 @@ func TestQueryNamespacedResources(t *testing.T) {
...
+			name:     "No matching resources by version",
+			kind:     existingKind,
+			version:  "v2",
+			expected: nil,
+		},
+		{
+			name:     "Matching resources",
+			kind:     existingKind,
+			version:  existingVersion,
+			expected: testResources[1:],
```
rh-hemartin, line 175: This makes it error prone if we add a new element to the list. What about recreating the list? The other option is

```suggestion
			expected: []*unstructured.Unstructured{testResources[1]},
```


```suggestion
			expected: testResources[1:1],
```

cmd/api/routers/routers_test.go
```diff
@@ -18,28 +18,43 @@ import (
...
 	ctrl := Controller{Database: fake.NewFakeDatabase(testResources)}
 	router.Use(func(c *gin.Context) {
-		c.Set("apiResource", testAPIResource)
-		c.Next()
+		if core {
+			c.Set("apiResource", testCoreAPIResource)
+		} else {
+			c.Set("apiResource", testNonCoreAPIResource)
+			c.Next()
+		}
```
rh-hemartin, line 47:
```suggestion
		if core {
			c.Set("apiResource", testCoreAPIResource)
		} else {
			c.Set("apiResource", testNonCoreAPIResource)
		}
		c.Next()
```

rh-hemartin: From the Next docs:

```
// Next should be used only inside middleware.
// It executes the pending handlers in the chain inside the calling handler.
// See example in GitHub.
```

And from this example https://github.com/gin-gonic/gin/blob/master/logger.go#L249, is useful only if called before some code in the middleware. Is not useful to call it at the end. See also this issue: https://github.com/gin-gonic/gin/issues/287. So I'm editing my suggestion and we should review our current middleware at some point to make proper use. Its not that is not working, but it leads to some confusion, at least for me.

maruiz93: Good research! 
I created https://github.com/kubearchive/kubearchive/issues/363 for addressing that thing in the middleware functions and I will remove the `next()` call in this test

maruiz93: addressed

pkg/database/fake/database_generated_test.go
maruiz93: Addressed, but note that I've used `testResources[1:2]` instead of `testResources[1:1]` The last one doesn't do what we expect

cmd/api/routers/routers.go
```diff
@@ -47,6 +47,30 @@ func (c *Controller) GetNamespacedResources(context *gin.Context) {
...
 
+func (c *Controller) GetAllCoreResources(context *gin.Context) {
+	version := context.Param("version")
+	kind := getAPIResourceKind(context)
+	resources, err := c.Database.QueryCoreResources(context.Request.Context(), kind, version)
+
+	if err != nil {
+		context.JSON(http.StatusInternalServerError, err.Error())
+	}
+	context.JSON(http.StatusOK, resources)
```
beerparty, line 58: Where can I find documentation about this usage? Specifically, why the second call to `context.JSON` is not in an `else`.

rh-hemartin: In their docs: https://pkg.go.dev/github.com/gin-gonic/gin#Context.JSON

And I think you are right,  that context.JSON needs to be `abort.JSON` if it needs to exit. We have a small utility function for that. And this sounds like we are not covering the return code in a test, or just the OK case.

maruiz93: I've updated this to the abort so that we return after setting the JSON output to an error

maruiz93: We needed also the return, if not, several JSONs are appended to the response.

I've also added some tests for checking the output on failure

rh-hemartin (approved):

skoved (approved):

Issue #261, opened 2024-07-26
Document how to enable tracing in production


rh-hemartin:
We should document how to enable tracing on a production environment. Currently I see two options readily available:

* Redeploy specifying a value in the `observability.endpoint` (after #214 gets completed)
* Go to the deployments and change their environment variables.

Other options that include changing code are:

* Have a ConfigMap with the configuration that containers read to configure themselves (could be hot-reloaded).

rh-hemartin:
As we are moving away from using Helm as production deployment, the process to be used is to modify the different environment variables (unless we change how we configure tracing). See #328 .

PR #367, opened 2024-09-03
Add metrics and Grafana stack

### LINKED ISSUE #356: Expose metrics
We do need to expose metrics so we can be monitored for performance, problems, etc.

Our first customer (Konflux-CI) uses Prometheus for metrics, however its also an industry standard so exposing a Prometheus endpoint for scrapping sounds like a sensible option. Prometheus also supports Remote Writes and it has a OTLP endpoint, however those options are more modern and probably not widely supported.

I see two different approaches:

* Expose endpoints with Prometheus metrics on our three Deployments
* Send OpenTelemetry metrics to a centralized OpenTelemetry collector (so we need another Deployment) and expose the metrics there

Both approaches are simple to test for one single component, so we may explore the topic a little bit before going with any of those.

rh-hemartin:

Resolves #356 

## Release Note

```release-note
KubeArchive emits metrics in OTLP format. See the KubeArchive documentation for  more information.
```

## Notes for Reviewers

* Changed Jaeger for the Grafana Stack. Interesting for us are: Prometheus to store metrics, Grafana to visualize metrics and traces, Tempo to store traces.
* Added an intermediate OpenTelemetry collector that aggregates everything from the KubeArchive components. In production, operators are supposed to provide the correct variables to our deployments so we send wherever they want: maybe they have a centralized OpenTelemetry Collector, maybe they create an OpenTelemetry Collector for KubeArchive, maybe they point directly to a Prometheus instance, etc. Instructions to enable tracing in production at #261


beerparty (approved):

skoved (approved):

charts/kubearchive/templates/observability/grafana-lgtm-stack.yaml
```diff
@@ -0,0 +1,105 @@
...
+          exporters: [otlphttp/traces]
+        metrics:
+          receivers: [otlp,prometheus/collector]
+          processors: [batch]
+          exporters: [otlphttp/metrics]
+        logs:
+          receivers: [otlp]
+          processors: [batch]
+          exporters: [otlphttp/logs]
+
```
maruiz93, line 104: I would like to propose extracting the content of `otel-config.yaml` to a file.
I believe that having this kind of content directly embedded in the definition of the YAML file makes the chart and the configuration less readable (and there is no syntax highlight either).

The steps for extracting the file could be:

1.  Create a `grafana-otelcol-config.yaml` under `charts/kubearchive/conf/` with the contents of the `otelcol-config.yaml` in the configmap.
2.  Apply the following change on the configmap to make helm embed the content of the file again:
```suggestion
  otelcol-config.yaml: |-
    {{- tpl ($.Files.Get "conf/grafana-otelcol-config.yaml") . | nindent 4 }}
```

More information of this helm templating feature in [here](https://helm.sh/docs/chart_template_guide/accessing_files/#basic-example)

Using `{{ (.Files.Glob "conf/grafana-*").AsConfig | indent 2 }}` could work too, but then the file name has to match key in the ConfigMap

charts/kubearchive/templates/observability/otel-collector.yaml
```diff
@@ -0,0 +1,97 @@
...
+          processors: [batch]
+          exporters: [debug, otlphttp]
+        metrics:
+          receivers: [otlp, prometheus]
+          processors: [batch]
+          exporters: [debug, otlphttp]
+        logs:
+          receivers: [otlp]
+          processors: [batch]
+          exporters: [debug]
```
maruiz93, line 46: Following the previous suggestion, I would also extract this to `charts/kubearchive/conf/otel-otelcol-config.yaml` and do a change like this:

```suggestion
  otel-collector-config: |-
    {{- tpl ($.Files.Get "conf/otel-otelcol-config.yaml") . | nindent 4 }}
```

maruiz93 (changes requested):

charts/kubearchive/templates/observability/grafana-lgtm-stack.yaml
rh-hemartin: Done

charts/kubearchive/templates/observability/otel-collector.yaml
rh-hemartin: Done

maruiz93 (approved):

Issue #305, opened 2024-08-01
Create health check endpoint for API Server


maruiz93:
The API Server should expose an endpoint for doing a health check in which the connectivity to both k8s API and database should be tested.

PR #369, opened 2024-09-03
Fix envvar for cache expiration

rh-hemartin:

Solves a bug introduced in #340 

## Release Note

```release-note
NONE
```

## Notes for Reviewers

* The first test that makes sure that every deployment was OK was too fast, as well as the `rollout status` that happens on the `hack/quick-install.sh` script. This is why this could pass CI.
* I put the cache environment variables get operator before the database, so it crashes before it starts with the retries that connect to the database. I think the retries to the database make the Deployment seem OK, then crash when it connected and got to the cache configuration.
* Also introduced a check for the retry on the integration test, because the general error was not being checked.
* I think these problems will be mitigated by #305

### DIFF
--- a/test/integration/integration_test.go
+++ b/test/integration/integration_test.go
@@ -33,7 +33,7 @@ func TestKubeArchiveDeployments(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	retry.Do(func() error {
+	retryErr := retry.Do(func() error {
 		deployments, err := client.AppsV1().Deployments("kubearchive").List(context.Background(), metav1.ListOptions{})
 		if err != nil {
 			return fmt.Errorf("Failed to get Deployments from the 'kubearchive' namespace: %w", err)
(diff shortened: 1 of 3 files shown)


test/integration/integration_test.go
```diff
@@ -56,6 +56,10 @@ func TestKubeArchiveDeployments(t *testing.T) {
 
 		return errors.New("Timed out while waiting deployments to be ready.")
```
beerparty, line 57: Need a "for"

```suggestion
		return errors.New("Timed out while waiting for deployments to be ready.")
```

test/integration/integration_test.go
```diff
@@ -33,7 +33,7 @@ func TestKubeArchiveDeployments(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	retry.Do(func() error {
+	err = retry.Do(func() error {
```
beerparty, line 36: You should probably give this `err` variable a different name to distinguish it from the `err` used inside the loop. While the two won't collide programmatically, it is confusing for a reviewer. Especially since it is not referenced again until line 60, well below the function definition (so the conditional seems misplaced).

```suggestion
	retryErr = retry.Do(func() error {
```

test/integration/integration_test.go
rh-hemartin: Fixed!

test/integration/integration_test.go
rh-hemartin: Changed!

beerparty (approved):
LGTM

skoved (approved):

mafh314 (approved):

PR #374, opened 2024-09-04
Add observability documentation

### LINKED ISSUE #261: Document how to enable tracing in production

rh-hemartin:

Resolves #261 

## Release Note

```release-note
NONE
```

## Notes for Reviewers

* Called "observability" instead of "tracing" because now we support metrics too.


beerparty (approved):
LGTM

PR #376, opened 2024-09-04
Fix GetAPIResources case with core resources

### LINKED ISSUE #375: Bug with `GetAPIResource` and core resources
Currently `GetAPIResource` builds a URL to retrieve names from the k8s API. This URL is built with `/apis` as base and has a comment that it needs to be changed when core resources are supported. They are now as #360 is merged. This `/apis` root of the URL causes a problem. It needs to be fixed.

**Acceptance Criteria**

* `GetAPIResource` works for core resources too.

rh-hemartin:

Resolves #375 

## Release Note

```release-note
NONE
```

## Notes for Reviewers

* Extracted URL creation to a method to test just that logic easier.


cmd/api/discovery/apiresources.go
```diff
@@ -53,3 +47,17 @@ func GetAPIResource(client rest.Interface) gin.HandlerFunc {
...
+func getDiscoveryURL(group, version string) string {
+	groupVersion := version // Base case: core resource, just version
+	if group != "" {
+		groupVersion = fmt.Sprintf("%s/%s", group, version)
+	}
+
+	url := fmt.Sprintf("/api/%s", groupVersion) // Base case: core resource, /api as root
+	if group != "" {
+		url = fmt.Sprintf("/apis/%s", groupVersion)
+	}
```
beerparty, line 60: You have duplicate conditionals here - you should consolidate the logic and remove the intermediate `groupVersion`:

```suggestion
	url := fmt.Sprintf("/api/%s", version) // Base case: core resource, just version, /api as root
	
	if group != "" {
		url = fmt.Sprintf("/apis/%s/%s", group, version) // Extended case: group and version, /apis as root
	}
```

beerparty (changes requested):

cmd/api/discovery/apiresources.go
rh-hemartin: Good catch Chris! This is much simpler.

mafh314 (approved):

beerparty (approved):
LGTM

PR #378, opened 2024-09-05
Add installation docs

### LINKED ISSUE #252: Installation documentation

rh-hemartin:

Resolves #252 

## Release Note

```release-note
NONE
```

## Notes for Reviewers

* The `kubectl rollout` feels a little weird, but I don't want to change that right now. We can followup in a different issue.
* If you reapply the `kubearchive.yaml` file, it overwrites the database secret. We may want to rethink how we want users to deal with that. We can followup in a different issue.
* I introduced `GIN_MODE` as it was pretty straightforward to do.
* I did a little conditional in the page to use the latest download link if the user is looking at the `main` version, otherwise it uses the versioned link. Its not ideal, but if we keep it small, its not that bad.

### DIFF
--- a/docs/modules/ROOT/pages/getting-started/installation.adoc
+++ b/docs/modules/ROOT/pages/getting-started/installation.adoc
@@ -1,0 +1,62 @@
+= Installation
+
+ifeval::["{page-component-display-version}" == "main"]
+:download-path: https://github.com/kubearchive/kubearchive/releases/latest/download
+endif::[]
+ifeval::["{page-component-display-version}" != "main"]
+:download-path: https://github.com/kubearchive/kubearchive/releases/download/{page-component-display-version}
+endif::[]
+
+This document explains how to install and configure KubeArchive in your Kubernetes cluster.
+
+== Prerequisites
+
+* A PostgreSQL instance
+* CertManager is installed on the Kubernetes cluster (+v1.9.1)
+* Knative Eventing is installed on the Kubernetes cluster (+v1.15.0)
+
+== Prepare the Database
+
+. Create ada user and a password for KubeArchive to use.
+. Create a database for KubeArchive, with admin permissions granted to the user you just created.
+. As the user, execute the `ddl-resource.sql` we provide for initializing the database, located at:
++
+[source,bash,subs="attributes"]
+----
+wget {download-path}/ddl-resource.sql
+psql -h database.example.com -U username ... -f ddl-resource.sql
+----
++
+This file creates a PostgreSQL schema, a table for KubeArchive and other stuff KubeArchive
+requires.
+
+== Install and Configure KubeArchive
+
+. Apply the release YAML to install KubeArchive:
++
+[source,bash,subs="attributes"]
+----
+kubectl apply -n kubearchive -f {download-path}/kubearchive.yaml
+----
+
+. Provide the credentials for the database in the `kubearchive-database-credentials` secret:
++
+[source,bash]
+----
+cat <<EOF > /tmp/patch.yaml
+stringData:
+  POSTGRES_URL: <your-postgresql-url>
+  POSTGRES_USER: <kubearchive-username>
+  POSTGRES_DB: <kubearchive-database>
+  POSTGRES_PASSWORD: '<kubearchive-password>'
+EOF
+kubectl patch -n kubearchive secret kubearchive-database-credentials --patch-file /tmp/patch.yaml
+----
+
+. Restart the `kubearchive-api-server` and `kubearchive-sink` Deployments to update
+the database credentials secret:
++
+[source,bash]
+----
+kubectl rollout -n kubearchive restart deployment kubearchive-sink kubearchive-api-server
+----
--- a/charts/kubearchive/templates/api_server/api_server.yaml
+++ b/charts/kubearchive/templates/api_server/api_server.yaml
@@ -25,6 +25,8 @@ spec:
               readOnly: true
               mountPath: /etc/kubearchive/ssl/
           env:
+            - name: GIN_MODE
+              value: {{ ternary "release" "debug" .Values.global.production }}
             - name: POSTGRES_DB
               valueFrom:
                 secretKeyRef:
--- a/docs/modules/ROOT/nav.adoc
+++ b/docs/modules/ROOT/nav.adoc
@@ -1,3 +1,5 @@
+* Getting Started
+** xref:getting-started/installation.adoc[]
 
 * Configuration
 ** xref:configuration/kubearchiveconfig.adoc[KubeArchiveConfig CRD]


docs/modules/ROOT/pages/getting-started/installation.adoc
```diff
@@ -0,0 +1,64 @@
...
+ifeval::["{page-component-display-version}" == "main"]
+https://github.com/kubearchive/kubearchive/releases/latest/download/ddl.sql
+endif::[]
+ifeval::["{page-component-display-version}" != "main"]
+https://github.com/kubearchive/kubearchive/releases/download/{page-component-display-version}/ddl.sql
+endif::[]
+----
++
+This file creates a PostgreSQL schema, a table for KubeArchive and other stuff KubeArchive
+requires.
```
maruiz93, line 28: Maybe we can also point the user to a way to apply directly through `psql` CLI

```suggestion
This file creates a PostgreSQL schema, a table for KubeArchive and other stuff KubeArchive
requires. It can be applied with `-f` option of `psql` CLI. See
link:https://www.postgresql.org/docs/current/app-psql.html#APP-PSQL-OPTION-FILE[docs].
```

docs/modules/ROOT/pages/getting-started/installation.adoc
```diff
@@ -0,0 +1,64 @@
...
+. Provide the credentials for the database in the `kubearchive-database-credentials` secret:
++
+[source,bash]
+----
+cat <<EOF > /tmp/patch.yaml
+stringData:
+  POSTGRES_URL: <your-postgresql-url>
+  POSTGRES_USER: <kubearchive-username>
+  POSTGRES_DB: <kubearchive-database>
+  POSTGRES_PASSWORD: '<kubearchive-password>'  # notsecret
```
maruiz93, line 53: Is the `#notsecret` directive needed here?

docs/modules/ROOT/pages/getting-started/installation.adoc
```diff
@@ -0,0 +1,64 @@
+= Installation
```
maruiz93, line 1: To keep the doc more readable I'd move the logic of creating the download path to the top of the doc and extract it to a variable:

```suggestion
ifeval::["{page-component-display-version}" == "main"]
:releases-download-path: https://github.com/kubearchive/kubearchive/releases/latest/download
endif::[]
ifeval::["{page-component-display-version}" != "main"]
:releases-download-path: https://github.com/kubearchive/kubearchive/releases/download/{page-component-display-version}
endif::[]

= Installation
```

docs/modules/ROOT/pages/getting-started/installation.adoc
```diff
@@ -0,0 +1,64 @@
...
++
+[source,bash,subs="attributes"]
+----
+ifeval::["{page-component-display-version}" == "main"]
+https://github.com/kubearchive/kubearchive/releases/latest/download/ddl.sql
+endif::[]
+ifeval::["{page-component-display-version}" != "main"]
+https://github.com/kubearchive/kubearchive/releases/download/{page-component-display-version}/ddl.sql
+endif::[]
+----
```
maruiz93, line 25: Following the suggestion of extracting the release download path to a variable this can be simpler:

```suggestion
----
{releases-download-path}/ddl.sql
----
```

maruiz93: also, as this is just a link maybe makes more sense to have it outside the bash section to be just an external link

docs/modules/ROOT/pages/getting-started/installation.adoc
```diff
@@ -0,0 +1,64 @@
...
++
+[source,bash,subs="attributes"]
+----
+ifeval::["{page-component-display-version}" == "main"]
+kubectl apply -f https://github.com/kubearchive/kubearchive/releases/latest/download/kubearchive.yaml
+endif::[]
+ifeval::["{page-component-display-version}" != "main"]
+kubectl apply -f https://github.com/kubearchive/kubearchive/releases/download/{page-component-display-version}/kubearchive.yaml
+endif::[]
+----
```
maruiz93, line 42: Following the suggestion of extracting the release download path to a variable, this can be simpler:

```suggestion
----
kubectl apply -f {release-download-path}/kubearchive.yaml
----
```

maruiz93 (changes requested):

docs/modules/ROOT/pages/getting-started/installation.adoc
```diff
@@ -0,0 +1,64 @@
...
+
+* A PostgreSQL instance
+* CertManager is installed on the Kubernetes cluster (+v1.9.1)
+* Knative Eventing is installed on the Kubernetes cluster (+v1.15.0)
+
+== Prepare the Database
+
+. Create a user and a password for KubeArchive to use.
+. Create a database for the user to use.
+. As the user, execute the `ddl.sql` we provide in the database, located at:
```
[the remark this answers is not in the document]
rh-hemartin: Right, I'll leave this open in case that PR goes first.

docs/modules/ROOT/pages/getting-started/installation.adoc
rh-hemartin: I will remove it and let's see, maybe the secrets scanner does not notice.

docs/modules/ROOT/pages/getting-started/installation.adoc
rh-hemartin: Ah, very nice. Done.

docs/modules/ROOT/pages/getting-started/installation.adoc
rh-hemartin: I changed the block slightly, now its `wget` + `psql` with example parameters.

docs/modules/ROOT/pages/getting-started/installation.adoc
rh-hemartin: I did some changes.

maruiz93:
> The kubectl rollout feels a little weird, but I don't want to change that right now. We can followup in a different issue.

I think it's ok having it like you proposed. 
If we end up opening an issue to address a change there I will post there too but, I think is worth noticing that there is an issue (https://github.com/kubernetes/kubernetes/issues/22368) opened in the Kubernetes project to get rollouts triggered on secret/configmap changes, but the issue is still unaddressed after 6 years...

maruiz93 (approved):

docs/modules/ROOT/pages/getting-started/installation.adoc
```diff
@@ -0,0 +1,62 @@
...
++
+[source,bash,subs="attributes"]
+----
+wget {download-path}/ddl.sql
```
beerparty: The file `ddl.sql` is in the `database` directory. Doesn't that need to be part of the path for `wget`?

docs/modules/ROOT/pages/getting-started/installation.adoc
```diff
@@ -0,0 +1,62 @@
...
+
+* A PostgreSQL instance
+* CertManager is installed on the Kubernetes cluster (+v1.9.1)
+* Knative Eventing is installed on the Kubernetes cluster (+v1.15.0)
+
+== Prepare the Database
+
+. Create a user and a password for KubeArchive to use.
+. Create a database for the user to use.
+. As the user, execute the `ddl.sql` we provide in the database, located at:
```
beerparty, line 22: I think this should either be "we provide for initializing the database" or "we provide in the `database` directory", depending on your intention.

docs/modules/ROOT/pages/getting-started/installation.adoc
```diff
@@ -0,0 +1,62 @@
...
+== Prerequisites
+
+* A PostgreSQL instance
+* CertManager is installed on the Kubernetes cluster (+v1.9.1)
+* Knative Eventing is installed on the Kubernetes cluster (+v1.15.0)
+
+== Prepare the Database
+
+. Create a user and a password for KubeArchive to use.
+. Create a database for the user to use.
```
beerparty, line 21: Since the database is for KubeArchive, could you reword this? Something like "Create a database for KubeArchive,  with admin permissions granted to the user you just created."

docs/modules/ROOT/pages/getting-started/installation.adoc
```diff
@@ -0,0 +1,62 @@
...
+  POSTGRES_URL: <your-postgresql-url>
+  POSTGRES_USER: <kubearchive-username>
+  POSTGRES_DB: <kubearchive-database>
+  POSTGRES_PASSWORD: '<kubearchive-password>'
+EOF
+kubectl patch -n kubearchive secret kubearchive-database-credentials --patch-file /tmp/patch.yaml
+----
+
+. Restart the `kubearchive-api-server` and `kubearchive-sink` Deployments so they get the
+changes on the database credentials secret:
```
beerparty, line 57: Should be "...so they get the changes **to** the database credentials secret" or "...to update the database credentials secret"

```suggestion
. Restart the `kubearchive-api-server` and `kubearchive-sink` Deployments to update
the database credentials secret:
```

docs/modules/ROOT/pages/getting-started/installation.adoc
rh-hemartin: No, the `wget` points to a release file. See the different files related to a release in knative eventing: https://github.com/knative/eventing/releases

docs/modules/ROOT/pages/getting-started/installation.adoc
rh-hemartin: Done

docs/modules/ROOT/pages/getting-started/installation.adoc
rh-hemartin: Done

docs/modules/ROOT/pages/getting-started/installation.adoc
rh-hemartin: Done

beerparty (approved):

docs/modules/ROOT/pages/getting-started/installation.adoc
```diff
@@ -0,0 +1,62 @@
...
+
+* A PostgreSQL instance
+* CertManager is installed on the Kubernetes cluster (+v1.9.1)
+* Knative Eventing is installed on the Kubernetes cluster (+v1.15.0)
+
+== Prepare the Database
+
+. Create ada user and a password for KubeArchive to use.
+. Create a database for KubeArchive, with admin permissions granted to the user you just created.
+. As the user, execute the `ddl-resource.sql` we provide for initializing the database, located at:
```
ggallen, line 22: This should all be *in* the SQL file. There's no reason for the user to do this.

> Create ada user

I also think there's a typo here.

docs/modules/ROOT/pages/getting-started/installation.adoc
```diff
@@ -0,0 +1,62 @@
...
+. Provide the credentials for the database in the `kubearchive-database-credentials` secret:
++
+[source,bash]
+----
+cat <<EOF > /tmp/patch.yaml
+stringData:
+  POSTGRES_URL: <your-postgresql-url>
+  POSTGRES_USER: <kubearchive-username>
+  POSTGRES_DB: <kubearchive-database>
+  POSTGRES_PASSWORD: '<kubearchive-password>'
```
ggallen, line 51: We can have defaults for all of these with the exception of POSTGRES_URL and put them in the SQL file.  Then if the user wants to change them, we can just note how to do that.

docs/modules/ROOT/pages/getting-started/installation.adoc
```diff
@@ -0,0 +1,62 @@
...
+EOF
+kubectl patch -n kubearchive secret kubearchive-database-credentials --patch-file /tmp/patch.yaml
+----
+
+. Restart the `kubearchive-api-server` and `kubearchive-sink` Deployments to update
+the database credentials secret:
++
```
ggallen, line 58: If the database is configured ahead of time, there is no need to restart these services.  

Why not just have the users download the file, edit it and add the POSTGRES_URL value, and apply it.  It seems simpler than letting the install happen and then patching it.  It also allows a user to change the other values as well if they have changed them from the default when creating the database.

rh-hemartin: I will create a issue to review how we handle this.

docs/modules/ROOT/pages/getting-started/installation.adoc
rh-hemartin: Fixed the typo. I created #381 to discuss this.

PR #387, opened 2024-09-06
Fix documentation workflow to publish tags

### LINKED ISSUE #383: Documentation does not publish versioned pages
Currently the documentation is not publishing versions depending on tags. That is happening locally, so I think its because it does not grab the tags properly. The workflow needs to have this step:

```
# Checkout 'fetch-tags' does not work properly, see https://github.com/actions/checkout/issues/701
      - name: Fecth tags
        run: git fetch --tags origin
```

The same as the release workflow has.

rh-hemartin:

Resolves #383 

## Release Note

```release-note
NONE
```

## Notes for Reviewers

* Added steps in the documentation build so we can check from PRs if the thing is generating correctly, set the retention to 1 to avoid flooding limits.

### DIFF
--- a/.github/workflows/documentation-build.yaml
+++ b/.github/workflows/documentation-build.yaml
@@ -12,6 +12,9 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
+      # Checkout 'fetch-tags' does not work properly, see https://github.com/actions/checkout/issues/701
+      - name: Fecth tags
+        run: git fetch --tags origin
       - name: Install Node.js
         uses: actions/setup-node@v4
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/documentation-build.yaml
```diff
@@ -22,3 +25,8 @@ jobs:
       - name: Generate Site
         working-directory: ./docs
         run: npm run build
+      - name: Upload Artifacts
+        uses: actions/upload-artifact@v4
+        with:
+          path: docs/build/site
+          retention-days: 1
```
maruiz93, line 32: Why do we need this step? Debugging purposes?

rh-hemartin: Yes

maruiz93 (approved):

Issue #381, opened 2024-09-06
Include user creation in ddl-resource.sql


rh-hemartin:
In #378 there was a changed proposed to the installation flow. Instead of having the user provide the user and database for KubeArchive, the `ddl-resource.sql` should create the user and database instead.

rh-hemartin:
I think a problem with this approach is that it assumes that the database user our user is provided has the permissions to create users and databases. I don't think that would be the case. However we can go with this approach and see if we need to change it later.

maruiz93:
> a problem with this approach is that it assumes that the database user our user is provided has the permissions to create users and databases

I am not following here, so we are saying that we the KubeArchive, instead of:
1. Create a DB user for kubearchive
2. Create a database for kubearchive inside of the postgres
3. Apply the given DDL

Just do:
1. Apply the given DDL

The one that applies the DDL would be the same that creates a DB user and the DB, so I think it's a good idea to merge all the steps in just one.

I do not see the problem though (maybe I get lost with the different users of the sentence) can you rephrase it?

maruiz93:
The only problem I see here is that the default password of the kubearchive user will be in the DDL too so the user will potentially need to change it in several places (secret and DDL)

rh-hemartin:
> I do not see the problem though (maybe I get lost with the different users of the sentence) can you rephrase it?

The problem I see is that the database user that executes the SQL may not have permissions to create databases and other users.

> The one that applies the DDL would be the same that creates a DB user and the DB

I think this is incorrect: the one that needs to apply the DDL should be the KubeArchive Database User. One of the statements inside the DDL fail when applied by the admin user. See #384 .

> so we are saying that we the KubeArchive, instead of:

Just to be a little bit more specific, we are not executing the SQL. This is a SQL script meant to be run by the one who installs KubeArchive.

maruiz93:
> I think this is incorrect: the one that needs to apply the DDL should be the KubeArchive Database User. One of the statements inside the DDL fail when applied by the admin user. See https://github.com/kubearchive/kubearchive/issues/384 .

I've reopened that issue and added a comment, PTAL.

> Just to be a little bit more specific, we are not executing the SQL. This is a SQL script meant to be run by the one who installs KubeArchive.

Yes we were on the same page here but I wasn't very clear so appreaciated the puntualization

ggallen:
> I think this is incorrect: the one that needs to apply the DDL should be the KubeArchive Database User.

[...]

PR #394, opened 2024-09-09
Unify DB Deployment

### LINKED ISSUE #390: Unify the deployment of the database for testing and development purposes
## Context

KubeArchive in meant to be used along a Postgres Database provided by the user.

For development and testing purposes we deploy our own database.

### Development

We provide the option `database.enabled` in KubeArchive's Helm chart to deploy a database along the other components of KubeArchive.

The database resources are in [charts/kubearchive/templates/database](https://github.com/kubearchive/kubearchive/tree/main/charts/kubearchive/templates/database)

### Testing

For the [integration tests](https://github.com/kubearchive/kubearchive/blob/main/.github/workflows/go_test_integration.yml) we use the [hack/quick-install.sh script](https://github.com/kubearchive/kubearchive/blob/main/hack/quick-install.sh) that deploys a Database and configure it in the way that we expect the user to do:
1. Create a kubearchive user
2. Create a kubearchive database
3. Apply the DDL of the release

## Problem

The code for both deployments is really similar and the aim of this issue is to unify them to prevent duplication and keep things short while keeping the integration test steps as similar as we could to the steps that will be reproduced by a Production user.

## Proposal

* Adjust the Helm Chart DB template with the following:
  * Create a specific namespace for the database
  * Create a ConfigMap with the contents of `ddl-resource.sql`
  * Move the database to its own namespace
  * Deploy the database with credentials different from the ones that kubearchive will use (e.g, `postgres` or `root`)
  * Mount the ConfigMap with the `ddl-resource.sql` in the Deployment, in different locations (to be automatically executed or not) based on the chart value `global.production`
* In `hack/quick-install.sh` replace the call of `hack/database-install.sh` for the following:

  ```bash
  kubectl apply -f - <<< $(helm template kubearchive charts/kubearchive \
    -s templates/database/database.yaml \
    --set "database.enabled=true"  --set "global.production=true")

  kubectl rollout status deployment --namespace=databases --timeout=30s
  sleep 15  # Wait for PostgreSQL to be really ready

  echo "CREATE USER kubearchive WITH ENCRYPTED PASSWORD 'P0stgr3sdbP@ssword';" | kubectl exec -i -n databases deploy/kubearchive-database -- psql -h localhost
  echo "CREATE DATABASE kubearchive WITH OWNER kubearchive;" | kubectl exec -i -n databases deploy/kubearchive-database -- psql -h localhost
  kubectl exec -i -n databases deploy/kubearchive-database -- psql -h localhost --user=kubearchive kubearchive -f /ddl.sql
  ```

* Remove `hack/database-install.sh` file as the functionality is replaced

### Other tweaks that could be also addressed in this or in other issues to improve the workflow:

* Move the creation of the user and database to the DDL (addressed in https://github.com/kubearchive/kubearchive/issues/381)
* Add a readiness probe on the DB deployment to remove the `sleep`

maruiz93:

Resolves #390

## Release Note

```release-note
NONE
```

## Notes for Reviewers

The main purpose of this PR is to show my proposal of unifying the DB deployment for development and testing
so we avoid the duplication we currently have of the DB resources inside the helm chart and in the `hack/install-database.sh` script.

In this PR I am moving the development closer to the production way and not the other way around:
- The database deployment is being deployed in a namespace different from `kubearchive`
- The database deployment is not having the kubearchive credentials when instantiated

I copied the file `ddl-resources.sql` into the helm chart so now is duplicated.
We shouldn't have duplication but it's tentative while in draft mode to show how I envision this change.
I think that we probably should take advantage of sym links, maybe have the real file in the chart and the sym link under `database/ddl-resource.sql`.

Also, this won't work as it is for development, before, we should address #381 to have the user and database creation applied from the DDL.


hack/quick-install.sh
```diff
@@ -5,7 +5,16 @@
 set -o errexit
 set -o xtrace
 
-bash hack/install-database.sh
+kubectl apply -f - <<< $(helm template kubearchive charts/kubearchive \
+  -s templates/database/database.yaml \
+  --set "database.enabled=true"  --set "global.production=true")
+
+kubectl rollout status deployment --namespace=databases --timeout=30s
+sleep 15  # Wait for PostgreSQL to be really ready
```
beerparty, line 13: If you are going to have a pause like this, it is a good idea to put info on the screen so a user watching the progress knows why the script is pausing here.

```suggestion
echo "Waiting 15 seconds for the PostgreSQL database to finish initialization..."
sleep 15
```

beerparty (changes requested):

hack/quick-install.sh
```diff
@@ -5,7 +5,16 @@
 set -o errexit
 set -o xtrace
 
-bash hack/install-database.sh
+kubectl apply -f - <<< $(helm template kubearchive charts/kubearchive \
+  -s templates/database/database.yaml \
+  --set "database.enabled=true"  --set "global.production=true")
```
rh-hemartin, line 10: This will fail to install, and we would be installing KubeArchive two times. I don't think we need to align testing with development as well. With some of the changes proposed here we are already approaching development to testing.

maruiz93: Note that with `-s templates/database/database.yaml` it should only render that specific file, so we don't need `ko` to be applied there because we are using the `postgres` image.

I tried the script `hack/quick-install.sh` with these changes and it actually works, but for this specific line you could just try locally:

```bash
 helm template kubearchive charts/kubearchive \
  -s templates/database/database.yaml \
  --set "database.enabled=true"  --set "global.production=true"
  ```

Comment (maruiz93) [reply hack/quick-install.sh:10]: > I don't think we need to align testing with development as well. With some of the changes proposed here we are already approaching development to testing.

About this, I think that the sentence I wrote in the Notes for reviewers ("My intention is to move the development closer to the production way and not the other way around.") is a little missleading.

I mean that with this PR, as I am "merging" the helm templates of the database and the script for the database into one, as they are not identical I chose the especifics of the testing environment (script) instead of the development environment (chart templates).

One of the examples of this is that I chose to use the `root` credentials instead of the kubearchive credentials in the `Deployment`

So, I don't think that there is extra work to be done related to the testing/development alignment apart from what I am covering in the current PR

Comment (rh-hemartin) [reply hack/quick-install.sh:10]: > Note that with -s templates/database/database.yaml it should only render that specific file, so we don't need ko to be applied there because we are using the postgres image.

Ah, I need to read better. Sorry about that.

File: charts/kubearchive/values.yaml [hunk 3]
```diff
@@ -71,7 +71,7 @@ database:
 
   name: "{{ .Release.Name }}-database"
 
-  url: "{{ tpl .Values.database.name . }}.{{ .Release.Namespace }}.svc.cluster.local"
+  url: "{{ tpl .Values.database.name . }}.database.svc.cluster.local"
```
rh-hemartin, line 74:
```suggestion
  url: "{{ tpl .Values.database.name . }}.databases.svc.cluster.local"
```

charts/kubearchive/conf/ddl-resource.sql
rh-hemartin, line 1: You mention symlinking from here to `database/ddl-resource.sql` what about the other way around?

maruiz93: Also doable, my preference of having the "real" file inside the chart was because I thought of having a "exportable" helm chart (like if we hypothetically wanted to provide the helm chart to the KubeArchive users) but don't want to do that so there is no actual argument for doing it one way or the other.

If you prefer keeping the real file where it is, I can add the link to that file in the chart.

hack/quick-install.sh
ggallen: No sleep. Why not try to connect to the database in the script?

https://gist.github.com/nicerobot/1136dcfba6ce3da67ce3ded5101a4078 has some examples.

maruiz93: I am thinking of adding a readiness probe in the deployment and remove the sleep, I think it will be easier to understand. For the first approach I used the sleep as it's what we are currently using in `hack/install-database.sh`, see https://github.com/kubearchive/kubearchive/blob/main/hack/install-database.sh#L99

ggallen:
Why use helm for the database at all?  Why don't we just have a script that installs postgres with some default values?

I'm not seeing any benefit for using helm here.

maruiz93:
I created this other PR (https://github.com/kubearchive/kubearchive/pull/401) that solved the issue in a different way, by using just the script instead of the helm chart for deploying the database. Please take a look.

If I get approvals to that PR, I will close this in favor of the other one.

maruiz93:
Closing this PR in favor of https://github.com/kubearchive/kubearchive/pull/401

PR #395, opened 2024-09-09
Introduce telemetry modes

### LINKED ISSUE #233: Observability configuration: enabled, disabled and delegated
From the discussions at #199 and #158 we think that we need to support more than two modes (enabled/disabled) for our telemetry. Given that we receive calls from upstream services (knative eventing apiserversource, Konflux UI in the future) and that those can be instrumented as well, we need a little more complexity. For that we think that we need to support three modes:

* Enabled: regardless of what upstream services do, KubeArchive sends telemetry data to the endpoints specified. In this case it needs to ignore the sampled flag in the traceparent header.
* Disabled: regardless of what upstream services do, KubeArchive does not send telemetry data.
* Delegated: if the upstream services record traces, KubeArchive does as well. If upstream services don't record the traces, KubeArchive do not as well. This means that KubeArchive honors the sampled flag in the traceparent header (which happens automatically as far as we tested).

Other names could be in discussion (in relation to the names exposed above):

* Enabled, disabled and forced (forced == enabled, enabled == delegated).

rh-hemartin:

Resolves #233 

## Release Note

```release-note
Now KubeArchive supports different telemetry modes of operation using `KUBEARCHIVE_OTEL_MODE`. This variable replaced `KUBEARCHIVE_OTEL_ENABLED`.
```

### DIFF
--- a/docs/modules/ROOT/pages/configuration/enable-observability.adoc
+++ b/docs/modules/ROOT/pages/configuration/enable-observability.adoc
@@ -4,8 +4,13 @@ KubeArchive is instrumented to emit observability data, specifically metrics and
 traces. This can be enabled and configured with two environment
 variables present on all of the KubeArchive components:
 
-* `KUBEARCHIVE_OTEL_ENABLED`: one of "true" or "false". Defaults to "false". It
-   controls if observability data is sent to `OTEL_EXPORTER_OTLP_ENDPOINT`.
+* `KUBEARCHIVE_OTEL_MODE`: one of "enabled", "delegated" or "disabled". Defaults to
+an "disabled". It controls when observability data is sent to `OTEL_EXPORTER_OTLP_ENDPOINT`:
+** "disabled": do not send traces or metrics.
+** "enabled": always send traces and metrics. Useful for observability in development.
+** "delegated": always send metrics but send traces only when the incoming request sent its traces (see
+link:https://www.w3.org/TR/trace-context-2/#sampled-flag[W3 Trace Context `sampled` flag]).
+Useful for observability in production.
 * `OTEL_EXPORTER_OTLP_ENDPOINT`: an OTLP compatible endpoint where traces are
     sent. By default it is set to an empty string.
 
@@ -26,8 +31,8 @@ spec:
       containers:
         - name: kubearchive-api-server
           env:
-            - name: KUBEARCHIVE_OTEL_ENABLED
-              value: "true"
+            - name: KUBEARCHIVE_OTEL_MODE
+              value: "delegated"
             - name: OTEL_EXPORTER_OTLP_ENDPOINT
               value: "http://otel-collector.observability.svc.cluster.local:4318"
 ----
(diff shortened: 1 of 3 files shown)


docs/modules/ROOT/pages/configuration/enable-observability.adoc
```diff
@@ -4,8 +4,13 @@ KubeArchive is instrumented to emit observability data, specifically metrics and
 traces. This can be enabled and configured with two environment
 variables present on all of the KubeArchive components:
 
-* `KUBEARCHIVE_OTEL_ENABLED`: one of "true" or "false". Defaults to "false". It
-   controls if observability data is sent to `OTEL_EXPORTER_OTLP_ENDPOINT`.
+* `KUBEARCHIVE_OTEL_MODE`: one of "enabled" or "delegated". Defaults to an empty string (`""`). It
+   controls when observability data is sent to `OTEL_EXPORTER_OTLP_ENDPOINT`:
+** Empty string (`""`): do not send traces or metrics.
```
maruiz93, line 9: What about having a explicit `disabled` instead of an empty value?

maruiz93: Great, I believe it's less error prone to see an explicit value in the configuration :)

pkg/observability/trace.go
```diff
@@ -75,11 +88,11 @@ func Start(serviceName string) error {
 	return nil
 }
 
-// canSkipInit returns a bool representing if OtelStartEnvVar is set to false. This function is a helper for Start.
+// canSkipInit returns a bool representing if OtelStartEnvVar is set empty or not. This function is a helper for Start.
 // Instrumentation should *ONLY* be started if this function returns false
 func canSkipInit() bool {
 	startEnv := os.Getenv(OtelStartEnvVar)
-	return strings.ToLower(startEnv) == "false"
+	return strings.ToLower(startEnv) == "disabled"
```
beerparty, line 95: Don't you need to check for empty and return true in that case?

beerparty (changes requested):

docs/modules/ROOT/pages/configuration/enable-observability.adoc
ggallen: How can you have `disabled` on a boolean value?  Or is this only for `KUBEARCHIVE_OTEL_MODE`?

rh-hemartin: This is only for `KUBEARCHIVE_OTEL_MODE`, which replaces the old (boolean) value.

pkg/observability/trace.go
```diff
@@ -75,11 +88,11 @@ func Start(serviceName string) error {
 	return nil
 }
 
-// canSkipInit returns a bool representing if OtelStartEnvVar is set to false. This function is a helper for Start.
+// canSkipInit returns a bool representing if OtelStartEnvVar is set empty or not. This function is a helper for Start.
```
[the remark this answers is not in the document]
rh-hemartin: Changed!

pkg/observability/trace.go
rh-hemartin: Its just checking "disabled", now the docs match.

beerparty (approved):
Thank you for the clarification in the comments.

Issue #297, opened 2024-07-31
CLI accepts short, normal and extended resource names


rh-hemartin:
Currently the CLI accepts "group/version/resource" (e.g. batch/v1/jobs) as a resource string to query the K8s and K9e APIs. We should accept the same forms as kubectl get to make for a better user experience.

PR #398, opened 2024-09-09
Add namespace support to the CLI

### LINKED ISSUE #296: CLI returns namespaced resources
The CLI accept and uses `--namespace` or `--all-namespaces`, if no one of those have been specified it uses the current namespace on the Kubeconf context as if it was provided on `--namespace`.

For core resources it needs #197 and #198

rh-hemartin:

Resolves #296 

## Release Note

```release-note
NONE
```

## Notes for Reviewers

* The way to specify resources will be addressed at #297

### DIFF
--- a/pkg/cmd/get.go
+++ b/pkg/cmd/get.go
@@ -20,6 +20,7 @@ type GetOptions struct {
 	IsCoreResource bool
 	APIPath        string
 	Resource       string
+	GroupVersion   string
 	Token          string
 
 	KubeArchiveHost string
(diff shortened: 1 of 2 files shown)


pkg/cmd/get.go
```diff
@@ -66,14 +67,20 @@ func NewGetCmd() *cobra.Command {
...
-	if strings.HasPrefix(o.Resource, "v1") {
+	o.GroupVersion = args[0]
+	if strings.HasPrefix(o.GroupVersion, "v1") {
 		o.IsCoreResource = true
 	}
 
-	o.APIPath = fmt.Sprintf("/apis/%s", o.Resource)
+	o.Resource = args[1]
+	APIPathWithoutRoot := fmt.Sprintf("%s/%s", o.GroupVersion, o.Resource)
+	if *o.kubeFlags.Namespace != "" {
```
beerparty, line 77: Is there a way to mock `*o.kubeFlags.Namespace` in the test code so you can check this path? If not, create a one line method (something like `getNamespace` or `getKubeFlagsNamespace`) so you can mock that and test this path of the code.

rh-hemartin: We don't need mocking for this, I just set the value. Changed!

beerparty (approved):
Thanks for adding the new unit test.

PR #400, opened 2024-09-10
Add health probes for API server

### LINKED ISSUE #305: Create health check endpoint for API Server

maruiz93:

Resolves #305

## Release Note

```release-note
NONE
```


cmd/api/main.go
```diff
@@ -59,18 +54,30 @@ func getKubernetesClient() *kubernetes.Clientset {
...
+	}
+
+	router.GET("/livez", controller.Livez)
+	router.GET("/readyz", controller.Readyz)
+
+	apisGroup.GET("/:group/:version/:resourceType", controller.GetAllResources)
+	apisGroup.GET("/:group/:version/namespaces/:namespace/:resourceType", controller.GetNamespacedResources)
+
+	apiGroup.GET("/:version/:resourceType", controller.GetAllCoreResources)
+	apiGroup.GET("/:version/namespaces/:namespace/:resourceType", controller.GetNamespacedCoreResources)
```
rh-hemartin, line 80: Purely visual/story-telling suggestion, feel free to ignore. I'm not sure GET and Use can be flipped:

```suggestion
	router.GET("/livez", controller.Livez)
	router.GET("/readyz", controller.Readyz)

	coreGroup := router.Group("/api")
	coreGroup.GET("/:version/:resourceType", controller.GetAllCoreResources)
	coreGroup.GET("/:version/namespaces/:namespace/:resourceType", controller.GetNamespacedCoreResources)

	nonCoreCoreGroup := router.Group("/apis")
	nonCoreCoreGroup.GET("/:group/:version/:resourceType", controller.GetAllResources)
	nonCoreCoreGroup.GET("/:group/:version/namespaces/:namespace/:resourceType", controller.GetNamespacedResources)

    // setup middleware for resource groups
	for _, group := range [...]*gin.RouterGroup{nonCoreCoreGroup, coreGroup} {
		group.Use(auth.Authentication(k8sClient.AuthenticationV1().TokenReviews(), cache, cacheExpirations.Authorized, cacheExpirations.Unauthorized))
		group.Use(auth.RBACAuthorization(k8sClient.AuthorizationV1().SubjectAccessReviews(), cache, cacheExpirations.Authorized, cacheExpirations.Unauthorized))
		// TODO - Probably want to use cache for the discovery client
		// See https://pkg.go.dev/k8s.io/client-go/discovery/cached/disk#NewCachedDiscoveryClientForConfig
		group.Use(discovery.GetAPIResource(k8sClient.Discovery().RESTClient()))
	}
```

cmd/api/routers/routers.go
```diff
@@ -110,6 +118,34 @@ func (c *Controller) GetNamespacedCoreResources(context *gin.Context) {
 	context.JSON(http.StatusOK, NewList(resources))
 }
 
+// Livez returns current server configuration
```
rh-hemartin, line 121: I would like to document WHY it returns just the configuration. "Isn't this supposed to check if the app is able to operate?". What about:

```suggestion
// Livez returns current server configuration as we don't have a clear deadlock indicator
```

cmd/api/main.go
maruiz93: I will try it and let you know in this thread

maruiz93: Just tried it out. The GET and Use cannot be flipped, that way the middleware don't get applied to the groups.

cmd/api/routers/routers.go
maruiz93: Change applied

cmd/api/routers/routers.go
```diff
@@ -110,6 +118,34 @@ func (c *Controller) GetNamespacedCoreResources(context *gin.Context) {
...
+	})
+}
+
+// Readyz checks Database connection
+func (c *Controller) Readyz(context *gin.Context) {
+	err := c.Database.Ping(context)
+	if err != nil {
+		abort.Abort(context, err.Error(), http.StatusServiceUnavailable)
+		return
+	}
```
beerparty, line 145: Is aborting what the code should be doing here? Can't you just set the context to `http.StatusServiceUnavailable` with an appropriate message?

beerparty (commented):
One question about the new code, other than that it looks good to me.

skoved (approved):

cmd/api/routers/routers.go
maruiz93: I'm calling `Abort()` because it's a utility defined [here](https://github.com/maruiz93/kubearchive/blob/main/cmd/api/abort/abort.go) that is used in other places of the API (endpoints and middleware) to save a couple of lines. It's just a shortcut to do these 3 steps:
* Log the error message
* Set the JSON output with a specific `message` and the desired http status
* abort the context (which I think it's expected based on go context management) 

So I think that we are good with the abort right here.

Issue #380, opened 2024-09-06
Database secret in the installation process


rh-hemartin:
Currently we ask the user to apply our YAML to install and then to patch the secret we add. This was documented in #378 . There were some comments proposing changes to the flow. If I understood correctly, the proposed flow is:

* Have default values on the database secret in the release YAML.
* Have the user download the release YAML, edit it and then apply it.

This looks simpler than patching and avoids the `kubectl rollout restart` after installing.

rh-hemartin:
There are some discussions about being agnostic of the database being used. That work is more than it sounds, as we would need to include more drivers, adapt calls, etc. I'm not sure its worth it.

ggallen:
My suggestion is simple. Just put all our development values in the secret YAML file.

1. Developers don't have to change anything, they just apply it.
2. Production are suggested to change it, so the instructions are the same.

PR #401, opened 2024-09-10
Extract DB deployment from helm chart

### LINKED ISSUE #390: Unify the deployment of the database for testing and development purposes

maruiz93:

Resolves #390

## Release Note

```release-note
NONE
```

## Notes for Reviewers

As discussed in the related issue, I created this PR to show how it would look extracting the DB from the helm chart
(as it isn't part of KubeArchive per se) and provide it just from a script.

I also made some tweaks to the current `hack/install-database.sh` script:

* Added a readinessProbe to prevent having a sleep to wait for the db to be available
* Refactor the usage of a particular pod for the usage of the deployment name that will always be the same.
* Extract the YAML resources to an external file so it's easier to read with syntax highlighting


hack/quick-install.sh
```diff
@@ -32,19 +32,9 @@ EOF
...
 kubectl apply -n kubearchive -f /tmp/kubearchive.yaml
 
-cat <<EOF > /tmp/patch.yaml
-stringData:
-  POSTGRES_URL: postgresql.databases.svc.cluster.local
-  POSTGRES_USER: kubearchive
-  POSTGRES_DB: kubearchive
-  POSTGRES_PASSWORD: 'P0stgr3sdbP@ssword'  # notsecret
-EOF
-kubectl patch -n kubearchive secrets kubearchive-database-credentials --patch-file /tmp/patch.yaml
```
rh-hemartin, line 47: I think we need to keep this. We have #380 to review how we provide the secret, but the state right know is that the user needs to `patch` and `rollout restart` after installing.

rh-hemartin:
If you want to change the DB credentials Secret name from `kubearchive-database-credentials` to `postgresql-credentials` you need to change the deployment instructions.

charts/kubearchive/templates/_helpers.tpl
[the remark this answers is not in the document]
maruiz93: Done

charts/kubearchive/values.yaml
```diff
@@ -61,33 +61,13 @@ operator:
...
-    name: "postgresdata"
-    storage: "5Gi"
-  # database service
-  service:
-    type: NodePort
-    port: 5432
+  name: "postgresql"
+  namespace: "databases"
+  port: 5432
+  url: "{{ tpl .Values.database.name . }}.{{ tpl .Values.database.namespace . }}.svc.cluster.local"
```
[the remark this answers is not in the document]
maruiz93: Done

hack/quick-install.sh
maruiz93: I removed it because the secret already had these values. 
If I'm understanding this correctly I should put empty values in the secret based on a chart value (like we have before for `database.enabled`) so that this patch actually do something, right?

maruiz93:
> If you want to change the DB credentials Secret name from `kubearchive-database-credentials` to `postgresql-credentials` you need to change the deployment instructions.

My bad, I just changed it to `kubearchive-database-credentials` again

hack/quick-install.sh
rh-hemartin: I don't think you need to put empty values based on a chart value, just put empty values. The `patch` will configure it properly.

maruiz93: Then, the usage for "development" won't work.
The instructions we currently have for development are:
1. Run `hack/install-database.sh`
2. Install the helm chart

The patch is being done in the `hack/quick-install.sh` meaning, it won't be done in the development as we have it.

Other alternative is to add one more step to the development procedure to run the patch after installing the chart.

hack/db-resources.yaml
```diff
@@ -0,0 +1,95 @@
...
+          volumeMounts:
+            - name: postgresql
+              mountPath: /var/lib/postgresql/data
+          env:
+            - name: POSTGRES_DB
+              value: root
+            - name: POSTGRES_USER
+              value: root
+            - name: POSTGRES_PASSWORD
+              value: password  # notsecret
```
ggallen, line 70: Why are these not the same values that are in the `values.yaml` file?

hack/quick-install.sh
rh-hemartin: Ah, right. Then maybe recovering the `_helpers.tpl` block and using `.Values.global.production` as conditional may work.

DEVELOPMENT.md
```diff
@@ -72,6 +72,11 @@ Install these tools:
     cmd/operator/generate.sh
     ```
 
+1. Install a database:
+   ```bash
+   hack/install-database.sh
```
ggallen, line 77: Can't we put these in an `integrations/database/postgres` directory?

charts/kubearchive/templates/api_server/api_server.yaml
```diff
@@ -24,31 +24,14 @@ spec:
             - name: tls-secret
               readOnly: true
               mountPath: /etc/kubearchive/ssl/
+          envFrom:
+            - secretRef:
+                name: {{ .Values.database.secret }}
```
ggallen, line 29: Why does the API server need the secret but no other database information?

charts/kubearchive/templates/api_server/api_server.yaml
```diff
@@ -24,31 +24,14 @@ spec:
...
-                  name: {{ tpl .Values.database.secret . }}
-                  key: POSTGRES_PASSWORD
-            - name: POSTGRES_URL
-              valueFrom:
-                secretKeyRef:
-                  name: {{ tpl .Values.database.secret . }}
-                  key: POSTGRES_URL
             - name: POSTGRES_PORT
-              value: "{{ required "A Postgres port must be specified" .Values.database.service.port }}"
+              value: "5432"
```
ggallen, line 34: Why is the postgres port needed here?

charts/kubearchive/templates/api_server/api_server.yaml
maruiz93: it needs all the database information, the api server, the same as the sink has all these variables:
* POSTGRES_USER
* POSTGRES_PASSWORD
* POSTGRES_DATABASE
* POSTGRES_URL
* POSTGRES_PORT

The first four are being mounted through the Kubernetes Secret in this line.
The change I did here is to simplify, when `envFrom.secretRef` is used it mounts all the values from a secret as environment variables so you don't need to specify them one by one.

ggallen:
> > If you want to change the DB credentials Secret name from `kubearchive-database-credentials` to `postgresql-credentials` you need to change the deployment instructions.
> 
> My bad, I just changed it to `kubearchive-database-credentials` again

I think it *should* be `postgres-credentials`.   It's specific to postgres.

charts/kubearchive/templates/api_server/api_server.yaml
ggallen: In that case, I suggest moving POSTGRES_PORT to the secret as well.  It looks out of place and is confusing to only see the port specified explicitly here.

rh-hemartin: `envFrom` unpacks the keys of the secret into the environment. As we were unpacking every key into the environment, this makes it more compact. So it only needs the secret name.

charts/kubearchive/templates/api_server/api_server.yaml
maruiz93: The same as the sink, in [this function](https://github.com/maruiz93/kubearchive/blob/main/pkg/database/connection.go#L36) we establish the postgres db connection string which needs the port. The port is one of the variables and is injected (like the rest) from the env variable.

charts/kubearchive/templates/api_server/api_server.yaml
maruiz93: does it make sense to have the port in a secret with the value encoded? Isn't better to have it in clear text?

I can add a comment to prevent confusion

DEVELOPMENT.md
maruiz93: I don't have any arguments not to move it, @rh-hemartin, do you?

hack/db-resources.yaml
rh-hemartin: These define the superadmin user for postgresql, I don't think that running KubeArchive with the superadmin user is realistic, so I used these for the production-like deployment (ci).

hack/quick-install.sh
maruiz93: I put it directly on the secret, looks good to you? or you still prefer to recover the `_helpers.tpl` block?

charts/kubearchive/templates/api_server/api_server.yaml
ggallen: But how is the port different than the user, URL, and database?  Why are those all encoded?  The only one that really needs encoding is the password.

But I'm all for keeping them together.

This also could help us making the database a true integration.  You'd (potentially) only need to change the secret to support a new database.

rh-hemartin: I think it would make sense to have it as a secret as we already have the URL. They are not really secrets, but obfuscating them is somewhat useful. Same way you would like the SSH port to be secret if you changed it from the default one.

maruiz93:
> > > If you want to change the DB credentials Secret name from `kubearchive-database-credentials` to `postgresql-credentials` you need to change the deployment instructions.
> > 
> > 
> > My bad, I just changed it to `kubearchive-database-credentials` again
> 
> I think it _should_ be `postgres-credentials`. It's specific to postgres.

These credentials are from postgres but are the ones used by kubearchive for the DB integration which is a KubeArchive dependency (no DB no place to archive anything) so I think that `kubearchive-database-credentials` name makes sense

charts/kubearchive/templates/api_server/api_server.yaml
maruiz93: ok, I'll make the change then

charts/kubearchive/templates/database/database_secret.yaml
```diff
@@ -3,8 +3,16 @@ apiVersion: v1
...
+  POSTGRES_URL: ""
+  POSTGRES_DB: ""
+  POSTGRES_USER: ""
+  POSTGRES_PASSWORD: ""
+  {{- else }}
+  POSTGRES_URL: {{ .Values.database.url | b64enc | quote }}
+  POSTGRES_DB: {{ .Values.database.postgresData.dbName | b64enc | quote }}
+  POSTGRES_USER: {{ .Values.database.postgresData.dbUser | b64enc | quote }}
+  POSTGRES_PASSWORD: {{ .Values.database.postgresData.dbPassword | b64enc | quote }}
+  {{- end}}
```
ggallen, line 18: Why not only have one set of values here, and those values are the development values?  Our instructions for production will tell the user to change them, but they don't have to.  And it will always work for development.

hack/quick-install.sh
rh-hemartin: Let's keep it on the secret and see where #380 lead us.

ggallen:
> > > > If you want to change the DB credentials Secret name from `kubearchive-database-credentials` to `postgresql-credentials` you need to change the deployment instructions.
> > > 
> > > 
> > > My bad, I just changed it to `kubearchive-database-credentials` again
> > 
> > 
> > I think it _should_ be `postgres-credentials`. It's specific to postgres.
> 
> These credentials are from postgres but are the ones used by kubearchive for the DB integration which is a KubeArchive dependency (no DB no place to archive anything) so I think that `kubearchive-database-credentials` name makes sense

But what if the next integration doesn't use a secret?  I think we should be explicit so that anyone looking for where we are using postgres can find it easily.

hack/quick-install.sh
ggallen: See my other comment.  Just put all the development values into the secret. I don't think you'll need this patch at all then.

rh-hemartin:
> > > > > If you want to change the DB credentials Secret name from `kubearchive-database-credentials` to `postgresql-credentials` you need to change the deployment instructions.
> > > > 
> > > > 
> > > > My bad, I just changed it to `kubearchive-database-credentials` again
> > > 
> > > 
> > > I think it _should_ be `postgres-credentials`. It's specific to postgres.
> > 
> > 
> > These credentials are from postgres but are the ones used by kubearchive for the DB integration which is a KubeArchive dependency (no DB no place to archive anything) so I think that `kubearchive-database-credentials` name makes sense
> 
> But what if the next integration doesn't use a secret? I think we should be explicit so that anyone looking for where we are using postgres can find it easily.

I think this is implying that changing the database is simple, but we will need changes in the code, so it does not make sense to generalize it now. When we have a series of users require another database, then we can generalize this. Now its just trying to solve what I see as distant-future problems.

charts/kubearchive/templates/database/database_secret.yaml
maruiz93: > Why not only have one set of values here, and those values are the development values? Our instructions for production will tell the user to change them, but they don't have to. And it will always work for development

IINM, We will address this in https://github.com/kubearchive/kubearchive/issues/380.
For now I am just leaving this logic as it was.

mafh314:
Since this script will be here `integrations/database/postgresql/install.sh` does it make sense to move all the [database folder](https://github.com/kubearchive/kubearchive/tree/main/database) too?

maruiz93:
> Since this script will be here `integrations/database/postgresql/install.sh` does it make sense to move all the [database folder](https://github.com/kubearchive/kubearchive/tree/main/database) too? If that makes sense, that could be cover by another issue.

I think it makes sense

hack/quick-install.sh
```diff
@@ -32,19 +32,18 @@ EOF
...
 helm template kubearchive charts/kubearchive -n kubearchive \
     --include-crds \
-    --set "global.production=true" \
-    --set "database.enabled=false" > /tmp/kubearchive-not-resolved.yaml
+    --set "global.production=true" > /tmp/kubearchive-not-resolved.yaml
 ko resolve -f /tmp/kubearchive-not-resolved.yaml --base-import-paths > /tmp/kubearchive.yaml
 kubectl apply -n kubearchive -f /tmp/kubearchive.yaml
 
-cat <<EOF > /tmp/patch.yaml
+cat << EOF > /tmp/patch.yaml
```
beerparty, line 39: For temp file safety in bash, you should use `mktemp` rather than hard coding a temp file name. I suggest something like the below and then use the variable in future references.

```suggestion
PATCH_TMPFILE=$(mktemp --prefix patch --suffix .yaml)
cat << EOF > ${PATCH_TMPFILE}
```

beerparty (changes requested):

hack/quick-install.sh
beerparty: Sorry, I just realized you did not write this code. You were just updating the spacing. I will resolve this comment and submit a separate ticket to fix this.

hack/db-resources.yaml
ggallen: I don't understand.  Why not just put in the real development values?

maruiz93: IINM, what @rh-hemartin tried to do in the original `hack/install-database.sh` script (where this piece comes from) was to replicate the user situation as much as possible.

Before installing KubeArchive, a user will probably have a Kubernetes database deployed somewhere (outside of kubearchive namespace, maybe even outside of the cluster) and with at least a user with admin privileges.

We ask the user to run the DDL and set things up so there is a kubearchive database user with which interact.

So, taking this under consideration, for testing purposes:
* We deploy a db in a namespace different from `kubearchive` (but inside the cluster for convenience) 
* The DB has one user database with admin privileges different from `kubearchive`, `root` in this case.
* This admin is the one that is going to perform the operations of creating a user

That's why the deployment of the datbase doesn't have the "development" values, because they are inserted after the database is deployed in the `install.sh` script.

ggallen: Well, this file appears to be gone now.  I'll drop this for now.

rh-hemartin (approved):

mafh314 (approved):

PR #403, opened 2024-09-10
Add CronJob to generate log files.

### LINKED ISSUE #402: Create a cronjob log generator.
We need an easy way to generate logs for development and testing.

We should create a `CronJob` that runs a job every minute to create job logs.

A similar one can be created for Tekton pipelines, but that's not in the scope here.

ggallen:

Resolves #402

## Release Note

```release-note
NONE
```

### DIFF
--- a/test/log-generators/cronjobs/install.sh
+++ b/test/log-generators/cronjobs/install.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+# Create a CronJob that run every minute, creating an Apache log file 1024 lines long.
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+
+namespace="generate-logs-cronjobs"
+kubectl create ns ${namespace}
+kubectl -n ${namespace} apply -f ${SCRIPT_DIR}
--- a/test/log-generators/cronjobs/cronjob.yaml
+++ b/test/log-generators/cronjobs/cronjob.yaml
@@ -1,0 +1,23 @@
+---
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: generate-log
+spec:
+  schedule: "* * * * *"
+  successfulJobsHistoryLimit: 1000
+  failedJobsHistoryLimit: 50
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          containers:
+            - name: generate
+              image: mingrammer/flog
+              imagePullPolicy: IfNotPresent
+              args:
+                - "-n"
+                - "1024"
+                - "-d"
+                - "20ms"
+          restartPolicy: OnFailure


test/log-generators/install-cronjobs.sh
```diff
@@ -0,0 +1,55 @@
...
+  schedule: "* * * * *"
+  successfulJobsHistoryLimit: 1000
+  failedJobsHistoryLimit: 50
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          containers:
+          - name: generate
+            image: redhat/ubi8
```
maruiz93, line 28: When I was investigating how to approach the logging integration part, I saw that there are some images available to create random logs, I think that we could take advantage of one of those:
* https://github.com/chentex/random-logger -- https://hub.docker.com/r/chentex/random-logger
* https://github.com/mingrammer/flog -- https://hub.docker.com/r/mingrammer/flog

The first one has more pulls in Dockerhub but the second one has more stars and forks, but I think that both of them could fit.

test/log-generators/install-cronjobs.sh
```diff
@@ -0,0 +1,55 @@
...
+# The pods for the last 1000 successful runs and 50 failed runs will be kept on the cluster, any logs
+# over those thresholds are automatically cleaned up by kubernetes.
+#
+# The job sleeps at random intervals for 1-3 seconds to get different timestamps in the log.
+
+namespace="generate-logs-cronjobs"
+
+kubectl create ns ${namespace}
+
+kubectl -n ${namespace} apply -f - << EOF
```
maruiz93, line 13: I would prefer to have the YAML definition of the CronJob in a separate file, specially for resources that are larger than 10 lines (for having a threshold defined).

The main advantage I see here is to take advantage of the syntax hightlighting and, in case of YAML, easier to check potential indentation problems.

maruiz93 (changes requested):

test/log-generators/install-cronjobs.sh
ggallen: I agree with you in most cases, but this is for testing so for me there is a lower threshold.

I think the ability to have this all in a single file outweighs putting the YAML in a second file. I can just copy this file anywhere I want and run it.  I don't have to keep track of two files.

Also, as soon as the YAML is split out, the next ask would be to split out the script into it's own file for the same reasons. Then it gets even more complicated.

I also have a second generator for Tekton pipelines coming. It's the same, all in a single file, and that's even more important for that one.  That requires some substitution to be done on the YAML files before they are applied. That's easily done with the YAML contained in the file.

test/log-generators/install-cronjobs.sh
ggallen: I looked at both of those, and concluded that for what I need now the script is sufficient.  We could of course update this later if we decide we need the additional log generating functionality that these images might provide.

test/log-generators/install-cronjobs.sh
ggallen: Another point is that with additional files we'll need to have each generator in its own sub-directory.  That's not a big deal really, but I think it cuts down on the clutter to have them all self-contained in one file.

maruiz93: > Also, as soon as the YAML is split out, the next ask would be to split out the script into it's own file for the same reasons. Then it gets even more complicated.

You are right about this, I also thought of mentioning this and suggest using Kustomization for that, but, I prefered to suggest using an image that already generate logs for us.

> I also have a second generator for Tekton pipelines coming. It's the same, all in a single file, and that's even more important for that one. That requires some substitution to be done on the YAML files before they are applied. That's easily done with the YAML contained in the file.

I think that with this approach you are prioritizing "easy to run" (for the copy/paste reference that you say) than "easy to read". I prefer the second approach and I think that we usually prefer that for the code that we haven't wrote ourselves (think of the general ask of not putting to much logic on top of the helm chart to be able to have it more readable). 

I haven't seen the Tekton pipeline thing yet so I'm not able to provide an alternative to it.
About this PR, I still think that extracting the YAML shouldn't be a big deal, and we could defer the discussion about the Kustomize thing (we can even create a separate issue+pr for that and see how it looks, the refactor shouldn't take long and we can always close the PR if we see that doesn't make things more readable)

test/log-generators/install-cronjobs.sh
maruiz93: The advantages that I see of using one of those images are:
* Better readability, it's a short script but bash isn't a verbose language so if we stick with the script I would want to add some comments over it to make things clearer
* Get rid of the embedded code in the container command
* Code already being maintained and used by a community
* No hard dependency for us, as it's just an image pulled the same as `redhat/ubi8`
* The change is quickly done even if we want to roll it back in the future --> this could be a point for keeping the script right now also, because as you mention we could update this in the future, but given the previous points I would rather try this out now

test/log-generators/cronjobs/install.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# Create a CronJob that run every minute, creating an Apache log file 1024 lines long.
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
```
maruiz93, line 5: Is this line used to locate the  directory where install.sh is no matter from where this is ran?

Maybe we could add a comment stating that.
I didn't thought about this but we could take advantage of this approach in future  scripts.

maruiz93 (approved):

PR #407, opened 2024-09-11
Create a Tekton Pipeline log generator.

### LINKED ISSUE #406: Create a Tekton Pipeline log generator.
We need an easy way to generate logs for development and testing.

We should create a CronJob that runs a Tekton `Pipeline` every minute to create job logs.

ggallen:

Resolves #406

## Release Note

```release-note
NONE
```

### DIFF
--- a/test/log-generators/pipelines/install.sh
+++ b/test/log-generators/pipelines/install.sh
@@ -1,0 +1,50 @@
+#!/bin/bash
+# Copyright KubeArchive Authors
+# SPDX-License-Identifier: Apache-2.0
+
+# Create a CronJob that runs every minute, submitting a Tekton pipeline that generates an Apache
+# log file 1024 lines in length.
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+
+# Parse command line arguments
+for i in "$@"
+do
+case $i in
+    --namespace=*)
+    NAMESPACE=`echo $i | sed 's/[-a-zA-Z0-9]*=//'`
+    ;;
+    --help)
+    HELP=True
+    ;;
+    *)
+    echo "Unknown option $i" # unknown option
+    HELP=True
+    UNKNOWN=True
+    ;;
+esac
+done
+
+HELP=${HELP:-""}
+UNKNOWN=${UNKNOWN:-""}
+export NAMESPACE=${NAMESPACE:-generate-logs-pipelines}
+
+# Help and usage
+if [ "${HELP}" == "True" ] || [ "${UNKNOWN}" == "True" ]; then
+    echo -e "$0
+
+    --namespace    Namespace to use to log generation.
+                   Default value is ${NAMESPACE}
+
+    "
+    if [ "${UNKNOWN}" == "True" ]; then
+      exit 1;
+    else
+      exit 0;
+    fi
+fi
+
+
+kubectl create ns ${NAMESPACE} --dry-run=client -o yaml | kubectl apply -f -
+cat ${SCRIPT_DIR}/*.yaml | envsubst | kubectl -n ${NAMESPACE} apply -f -
+kubectl rollout status deployment --namespace=${NAMESPACE} --timeout=90s
--- a/test/log-generators/pipelines/cluster-role-binding.yaml
+++ b/test/log-generators/pipelines/cluster-role-binding.yaml
@@ -1,0 +1,13 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: "triggers-example-eventlistener-clusterbinding-${NAMESPACE}"
+subjects:
+  - kind: ServiceAccount
+    name: tekton-triggers-example-sa
+    namespace: "${NAMESPACE}"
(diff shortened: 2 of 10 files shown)


test/log-generators/pipelines/install.sh
```diff
@@ -0,0 +1,114 @@
+#!/bin/bash
+
+# Create a CronJob that run every minute, submitting a Tekton pipeline that generates an Apache
```
beerparty, line 3: Small typo in the comment:

```suggestion
# Create a CronJob that runs every minute, submitting a Tekton pipeline that generates an Apache
```

test/log-generators/pipelines/install.sh
```diff
@@ -0,0 +1,114 @@
...
+# log file 1024 lines in length.
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
```
beerparty, line 6: What is the `SCRIPT_DIR` variable used for? I don't see it used anywhere in the script. If it is used in something called by this code, could you please add a comment with that information?

beerparty (changes requested):

test/log-generators/pipelines/install.sh
ggallen: It's there because I expect @maruiz93 to request other changes that will need it!

beerparty: If you end up not needing it, you should remove it. See also, [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it).

ggallen: YAGNI!  Love it. :-)

test/log-generators/pipelines/install.sh
```diff
@@ -0,0 +1,114 @@
...
+
+namespace="generate-logs-pipelines"
```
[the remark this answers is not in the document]
ggallen: @maruiz93, I think your suggestion works for that first substitution.  But what about this one?
```
apiVersion: batch/v1
kind: CronJob
metadata:
  name: submit-generate-logs
spec:
  schedule: "*/1 * * * *"
  jobTemplate:
    spec:
      template:
        spec:
          containers:
          - name: submit
            image: curlimages/curl
            args: ["curl", "-X", "POST", "--data", "{}", "el-cron-listener.${namespace}.svc.cluster.local:8080"]
          restartPolicy: Never
```

rh-hemartin (approved):

test/log-generators/pipelines/install.sh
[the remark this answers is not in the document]
maruiz93: After trying out my own suggestion I think that using kustomize for this isn't doable.

First of all, the `vars` feature is deprecated in favor of `replacements` ([docs](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/replacements/)) that aren't that versatile.

Second, using `vars` won't work for resources that aren't listed [in here](https://github.com/kubernetes-sigs/kustomize/blob/a280cdf5eeb748f5a72c8d94164ffdd68d03c5ce/api/konfig/builtinpluginconsts/varreference.go) meaning.

So, the functionality is limited, what leads me to think that this isn't the right too for doing this replacement.

Then, I change my suggestion to just extract the YAML files and do a sed of the default namespace `generate-logs-pipelines` and apply it.

This makes the YAML applyable with a `kubectl apply -f` with the default values with the bonus of the highlighting

test/log-generators/pipelines/install.sh
```diff
@@ -0,0 +1,114 @@
+#!/bin/bash
+
```
mafh314, line 2: Hey @ggallen could you add the license lines here, please?
`# Copyright KubeArchive Authors

# SPDX-License-Identifier: Apache-2.0
`

test/log-generators/pipelines/install.sh
ggallen: Fixed.

test/log-generators/pipelines/install.sh
ggallen: Fixed.!

test/log-generators/pipelines/install.sh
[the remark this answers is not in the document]
ggallen: @maruiz93 , see my latest changes which use `envsubst` to do this.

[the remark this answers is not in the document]
maruiz93: I like the way it looks now.

BTW, I see that you separated the k8s resources into it's own file. For me it's ok to have all the YAMLs together in the same file separated by `---`, specially if they are intended to be always applied together. The important thing for me was to not have mixed bash and YAML syntax in the same file :)

mafh314:
@ggallen could you add some information about the log generator in the test [README.md](https://github.com/kubearchive/kubearchive/blob/main/test/README.md), please?

maruiz93 (approved):

test/log-generators/pipelines/install.sh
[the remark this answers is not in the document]
ggallen: I separated them, and will leave them like that for now.

beerparty (approved):

PR #417, opened 2024-09-14
Install test database in development environment.

### LINKED ISSUE #380: Database secret in the installation process

ggallen:

Resolves #380, #381

## Release Note

```release-note
NONE
```

## Notes for Reviewers

The PR installs a database for KubeAchive using the cloudnative-pg operator.

It requires `kubectl` 1.31 or later.


database/postgres/install.sh
```diff
@@ -0,0 +1,35 @@
...
+
+# Wait for $LOCAL_PORT to become available.
+while ! nc -vz localhost ${LOCAL_PORT} > /dev/null 2>&1 ; do
+    #echo sleeping
+    sleep 0.1
+done
+psql -h localhost -U postgres -p ${LOCAL_PORT} -f kubearchive.sql
+
+# Kill all background jobs, including the port-forward.
+trap 'kill $(jobs -p)' EXIT
```
beerparty, line 35: Can you add a comment about **why** you have to kill all the background jobs? This seems like an odd thing to do as part of an install script.

database/postgres/install.sh
```diff
@@ -0,0 +1,35 @@
...
+LOCAL_PORT=5433
+export PGPASSWORD=$(kubectl -n ${NAMESPACE} get secret kubearchive-superuser -o jsonpath='{.data.password}' | base64 --decode)
+kubectl -n ${NAMESPACE} port-forward service/kubearchive-rw ${LOCAL_PORT}:5432 >& /dev/null &
+pid=$!
+
+# Wait for $LOCAL_PORT to become available.
+while ! nc -vz localhost ${LOCAL_PORT} > /dev/null 2>&1 ; do
+    #echo sleeping
+    sleep 0.1
+done
```
beerparty, line 31: Since this is an install script, people will likely be watching it to make sure it completes successfully. I would suggest that you add an echo before the wait loop explaining what is happening so they don't have to guess.

Also, if you want to indicate progress, the nerdy way to do that is with a line of dots, created by `echo -n .` inside the loop with a one second wait between attempts. Then a final `echo .` to force a CRLF at the end.

```suggestion
echo Waiting for database port $LOCAL_PORT to become available.
while ! nc -vz localhost ${LOCAL_PORT} > /dev/null 2>&1 ; do
    echo -n .
    sleep 1
done
echo .
```

rh-hemartin:
I want this PR to modify also `hack/quick-install.sh` which is used to get things ready on the CI to run integration testing.

database/postgres/install.sh
ggallen: I made the comment a little more explicit.

database/postgres/install.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+
+
```
rh-hemartin, line 3: Add this option so it fails when something where fails. I see some logs in the CI that says that the kubectl wait timed out while waiting for condition:

```suggestion
set -o errexit
```

database/postgres/install.sh
```diff
@@ -0,0 +1,37 @@
...
+VERSION="1.24.0"
+NAMESPACE="postgres"
+
+# Install cloudnative-pg operator.
+kubectl apply --server-side -f \
+  https://raw.githubusercontent.com/cloudnative-pg/cloudnative-pg/main/releases/cnpg-${VERSION}.yaml
+kubectl rollout status deployment --namespace=cnpg-system --timeout=90s
+
+# Create the postgres database server.
+kubectl create ns ${NAMESPACE}
```
rh-hemartin, line 16: Running `kubectl create ns` fails if the namespace already exists, use this other command to make the operation idempotent.

```suggestion
kubectl create ns ${NAMESPACE} --dry-run=client -o yaml | kubectl apply -f -
```

database/postgres/install.sh
```diff
@@ -0,0 +1,37 @@
...
+kubectl -n ${NAMESPACE} apply -f .
+kubectl -n ${NAMESPACE} wait pod/kubearchive-1 --for=create --timeout=60s
```
rh-hemartin, line 18: This yields the following error:

```
+ kubectl -n postgres wait pod/kubearchive-1 --for=create --timeout=60s
error: unrecognized condition: "create"
```

Looks it should be `--for=condition=ready`. However that may not work if the command is issued before the Pod starts creating. Maybe you can wait for the `Cluster` object to be ready instead.

database/postgres/install.sh
```diff
@@ -0,0 +1,37 @@
...
+kubectl -n ${NAMESPACE} wait pod --all --for=condition=ready --timeout=30s
+
+# Create the kubearchive database
+LOCAL_PORT=5433
+echo Forwarding port ${LOCAL_PORT} to service/kubearchive-wr:5432.
+export PGPASSWORD=$(kubectl -n ${NAMESPACE} get secret kubearchive-superuser -o jsonpath='{.data.password}' | base64 --decode)
+kubectl -n ${NAMESPACE} port-forward service/kubearchive-rw ${LOCAL_PORT}:5432 >& /dev/null &
+pid=$!
```
rh-hemartin, line 26: You don't need to register the `pid` if you then kill all background jobs.

database/postgres/install.sh
```diff
@@ -0,0 +1,37 @@
...
+
+echo Waiting for port $LOCAL_PORT to become available.
+while ! nc -vz localhost ${LOCAL_PORT} > /dev/null 2>&1 ; do
+    echo -n .
+    sleep 0.5
+done
+echo .
+psql -h localhost -U postgres -p ${LOCAL_PORT} -f kubearchive.sql
```
rh-hemartin, line 34: This will require `psql` to be installed on the CI. I don't recommend to go that way, this is why I did the thing with the pod and `cat` into the database pod.

database/postgres/kubearchive.sql
rh-hemartin, line 1: If this is replacing the `database/ddl-resource.sql` that file needs to go away and you need to change the `release.sh` so we release this file instead.

rh-hemartin:
Reviewing this a second time, I think you are doing too many changes at once. That is not a problem on itself, but It will take a while to get this ready to release.

test/log-generators/cronjobs/kaconf.yaml
```diff
@@ -0,0 +1,12 @@
...
+kind: KubeArchiveConfig
+metadata:
+  name: kubearchive
+spec:
+  resources:
+    - selector:
+        apiVersion: batch/v1
+        kind: Job
+      archiveWhen: has(status.startTime)
+      deleteWhen: has(status.completionTime)
```
maruiz93, line 12: Should this be part of this PR?

charts/kubearchive/values.yaml
```diff
@@ -69,9 +69,9 @@ database:
   # provide your own database, set 'enabled: false' and modify the secret after deploying
   enabled: true
 
-  name: "{{ .Release.Name }}-database"
+  name: "{{ .Release.Name }}-rw"
```
maruiz93, line 72: why `rw`? I guess it stands for `readwrite` but I don't see how this name is better than database

database/postgres/install.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
```
maruiz93, line 1: In the PR https://github.com/kubearchive/kubearchive/pull/401/files#diff-0d8d9d8bb96db992801d14fd0d9ef717539dd570c17eafda6ac8c3d6d339df42 I proposed locating these files under: `integrations/database/postgresql/` instead of `database/postgresql`

maruiz93 (changes requested):
I am ok in general terms with the idea of using an operator to deploy the Postgres DB for testing and development, but related to this specific PR, I see that we are mixing  goals here and in https://github.com/kubearchive/kubearchive/pull/401

The PR I mention is mainly for extracting the DB from the Helm chart, and this part is not addressed here (which is ok, it means that we need both) but there is some overlapping between both.

* In both PRs we touch the `hack/quick-install.sh` script, I guess that if we agree of going with the operator db, we should keep this one and I should remove that part from https://github.com/kubearchive/kubearchive/pull/401
* We both touch the `values.yaml` but I guess that should be done in the PR where the DB deployment is extracted from the helm chart.
* We both modify the script for installing the database, I guess I should remove this from https://github.com/kubearchive/kubearchive/pull/401
* We both moved where the postgresql database resources are located, I think that should be done in this PR, but I liked what we talked in the other PR of considering the database an integration and consequently, put its resources under `/integrations/database/postgresql/`

database/postgres/install.sh
ggallen: No. I added a comment in the "Notes for Reviewers" section:
>It requires kubectl 1.31 or later.

database/postgres/install.sh
ggallen: Well, `psql` is already there and available for the CI. It's running now.

Either way, `psql` is part of the general PostgreSQL installation.  It's the main way to interact with the database.  It seems silly to go through extra hoops to do this when there's a command line tool that will do it for you.

And database admin are going to use `psql`, so it's more like the user experience.

rh-hemartin: I saw it running but I didn't process that it was already installed, I tend to think about CI as some minimal environment but no, we are running with an image full of tools:

https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md

database/postgres/install.sh
ggallen: Fixed.

database/postgres/install.sh
ggallen: Fixed.

test/log-generators/cronjobs/kaconf.yaml
ggallen: Removed and now included in https://github.com/kubearchive/kubearchive/pull/431.

ggallen:
> Also, you are overlapping a lot with other PRs. In case they go first, you will need to reconciliate changes.

I would say that this PR should replace all of those other PRs, with the exception of https://github.com/kubearchive/kubearchive/pull/401, which is about to be committed.

database/postgres/install.sh
ggallen: Fixed.

charts/kubearchive/values.yaml
maruiz93: I see, it's explained in the [architecture section](https://cloudnative-pg.io/documentation/1.15/architecture/) of the cnpg docs

database/postgres/kubearchive.sql
ggallen: Done.

database/postgres/install.sh
ggallen: Moved everything to `integrations/database/postgresql`.

ggallen:
> You need to modify the `docs/modules/ROOT/pages/getting-started/installation.adoc` because now the file is not called (and uploaded) as `ddl-resource.sql`, but `kubearchive.sql` instead.

Fixed.

rh-hemartin (approved):

hack/quick-install.sh
```diff
@@ -15,35 +15,28 @@ kubectl apply -f https://github.com/knative/eventing/releases/download/knative-$
...
+# Make sure webhooks are up and running.
+LOCAL_PORT=8443
+kubectl -n kubearchive port-forward service/kubearchive-operator-webhooks ${LOCAL_PORT}:443 >& /dev/null &
+
+# Wait for $LOCAL_PORT to become available.
+while ! nc -vz localhost ${LOCAL_PORT} > /dev/null 2>&1 ; do
+    echo -n .
+    sleep 0.5
+done
+echo .
```
maruiz93, line 37: This checks made me think that we probably need to better configure the `/readyz` endpoint of the operator to do that, so the rollout and the wait made this work for us as the deployment "should know" when it is ready to provide service.

IINM, we have currently a [dummy function](https://github.com/maruiz93/kubearchive/blob/main/cmd/operator/main.go#L147) `healthz.Ping` that always returns true, but I guess we could use a more sophisticated check if we need.

Of course this is out of the scope of the PR, but just checking if we need to create a new issue to address this.

pkg/database/database.go
```diff
@@ -18,7 +18,7 @@ import (
 )
 
 const (
-	resourceTableName        = "kubearchive.resource"
+	resourceTableName        = "resource"
```
mafh314, line 21: @ggallen is this change to use the public schema?

ggallen: Yes.

hack/quick-install.sh
```diff
@@ -15,35 +15,28 @@ kubectl apply -f https://github.com/knative/eventing/releases/download/knative-$
...
-cat << EOF > /tmp/patch.yaml
-stringData:
-  POSTGRES_PORT: "5432"
-  POSTGRES_URL: postgresql.databases.svc.cluster.local
-  POSTGRES_USER: kubearchive
-  POSTGRES_DB: kubearchive
-  POSTGRES_PASSWORD: 'P0stgr3sdbP@ssword'  # notsecret
-EOF
-kubectl patch -n kubearchive secrets kubearchive-database-credentials --patch-file /tmp/patch.yaml
-kubectl rollout restart deployment --namespace=kubearchive kubearchive-sink kubearchive-api-server
```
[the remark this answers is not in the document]
ggallen: I think this should change and we should leave this out.  We should use our development credentials for everything, so things work right out-of-the-box.  And we should tell users what to change to make it production ready.

We can have a test case that does this, but I don't see any advantage to having settings that don't work at all.

hack/quick-install.sh
ggallen: @maruiz93 , create an issue for this for the operator, and I'll take a look.  I only added this because I know that the webhooks have to be responding before we can apply a `KubeArchveConfig` and have things work.  If it's not things don't work.

But other components may have similar endpoints that need this check too.  Or the probe.

integrations/database/postgresql/kubearchive-user.yaml
```diff
@@ -0,0 +1,11 @@
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: kubearchive-user
+type: kubernetes.io/basic-auth
+data:
+  host: a3ViZWFyY2hpdmUtcnc=
+  password: UDBzdGdyM3NkYlBAc3N3b3Jk
```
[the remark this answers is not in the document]
ggallen: Fixed.

pkg/database/database.go
mafh314: I that case, I thought that we should have `public.resource`, or will it be public if we don't specify the schema used?

ggallen: I believe `public` is the default.

hack/quick-install.sh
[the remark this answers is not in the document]
maruiz93: > We can have a test case that does this, but I don't see any advantage to having settings that don't work at all.

How can the settings not work?
IIUC, we can do one of the following:

* Follow the `DEVELOPMENT.md` instructions to set up the development environment, i.e., run the `install.sh` for deploying the database and install the kubearchive helm chart in non-production mode. This will work as intended.

* Run the `quick-install.sh` script to deploy everything from scratch. When this script install the helm chart it's done with the setting `global.production=true`. With the current chart, this means the secret will be deployed without credentials because it expects the user to provide them with a custom one, then, in the script, we set this credentials as we expect the user to do.

So, both of them, either running quick-install or going through the development process will end up with settings that are expected to work.

If we test the "production-like" deployment here by changing the db credentials on kubearchive here, I think we should then add a specific integration test for doing it.

hack/quick-install.sh
maruiz93: Done https://github.com/kubearchive/kubearchive/issues/439

maruiz93 (approved):

mafh314:
@ggallen I think we should also change this part of the [DEVELOPMENT.md](https://github.com/kubearchive/kubearchive/blob/main/DEVELOPMENT.md#initialize-the-database)

hack/quick-install.sh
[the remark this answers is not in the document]
ggallen: > How can the settings not work?

The initial production settings do not work, they are all empty strings.  The user must change them first before they can do anything.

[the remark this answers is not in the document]
ggallen: Based on our discussion today I'm going to resolve this conversation.

I've created https://github.com/kubearchive/kubearchive/issues/442 to update the documentation.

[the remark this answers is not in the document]
maruiz93: After the last talk, I am ok with going with your suggestion, specially considering that we will change the user instructions to modify the kubearchive YAML *before* applying it and not afterwards in https://github.com/kubearchive/kubearchive/issues/380.

Nevertheless, we could have an integration test for testing changes in the database credentials (like password rotation) that will need to have kubearchive application rolled out. I created an issue to address this: https://github.com/kubearchive/kubearchive/issues/444

PR #426, opened 2024-09-16
sink can archive objects when they are deleted

### LINKED ISSUE #337: Add "ArchiveOnDelete" functionality
It would be useful for KubeArchive to allow users to designate resource to be archived when deleted outside of KubeArchive.

There are other tools out there that handle deleting resources.  Giving users the ability to archive these resource before deletion could prove valuable.

This could easily be done by adding an optional `archiveOnDelete` key to each resource in the `KubeArchiveConfig` resource. It would default to `false`.  If this flag is `true`, then sink should receive the delete message and archive the resource.

The sink needs to differentiate between a delete it initiated and one initiated by another means, and if the sink initiated the delete then the `archiveOnDelete` flag should be ignored. `archiveOnDelete` only applies to delete messages.

The sink would also need to include `archiveOnDelete` in it's CEL expressions.

skoved:

Resolves #337 

## Release Note

```release-note
NONE
```

## Notes for Reviewers

The sink now checks if the received cloudevent's type is 'delete'. If it is, the sink checks if the object contained in the cloudevent meets the criteria for archiveOnDelete. If the object meets that criteria, the object is written to the database. If the object does not meet the criteria, the sink ignores the message. If the cloudevent's type is not 'delete', the sink handles the cloudevent normally.

### DIFF
--- a/cmd/sink/filters/filters.go
+++ b/cmd/sink/filters/filters.go
@@ -31,25 +31,28 @@ var ErrNoGlobal = errors.New("no global expressions exist")
 
 type Filters struct {
 	*sync.RWMutex
-	archive map[NamespaceGroupVersionKind]cel.Program
-	delete  map[NamespaceGroupVersionKind]cel.Program
-	path    string
+	archive         map[NamespaceGroupVersionKind]cel.Program
+	delete          map[NamespaceGroupVersionKind]cel.Program
+	archiveOnDelete map[NamespaceGroupVersionKind]cel.Program
+	path            string
 }
 
-// EmptyFilters returns a Filters struct with an empty archive and delete slice.
+// EmptyFilters returns a Filters struct with an empty archive, delete, and archiveOnDelete slices.
 func EmptyFilters() *Filters {
 	return &Filters{
-		RWMutex: &sync.RWMutex{},
-		archive: make(map[NamespaceGroupVersionKind]cel.Program),
-		delete:  make(map[NamespaceGroupVersionKind]cel.Program),
-		path:    os.Getenv(mountPathEnvVar),
+		RWMutex:         &sync.RWMutex{},
+		archive:         make(map[NamespaceGroupVersionKind]cel.Program),
+		delete:          make(map[NamespaceGroupVersionKind]cel.Program),
+		archiveOnDelete: make(map[NamespaceGroupVersionKind]cel.Program),
+		path:            os.Getenv(mountPathEnvVar),
 	}
 }
 
 // NewFilters creates a Filters struct from the path to a directory where a ConfigMap was mounted. If path is empty
-// string or path does not exist, it returns a Filters struct with empty archive and delete slices. It will attempt to
-// create all the cel programs that it can from the ConfigMap. Any errors that are encountered are wrapped together and
-// returned. Even if this function returns an error, the Filters struct returned can still be used and will not be nil.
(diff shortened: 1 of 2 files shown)


cmd/sink/filters/filters.go
```diff
@@ -141,15 +150,20 @@ func (f *Filters) createNamespaceFilters(namespace, file string, globalArchive,
...
 func (f *Filters) createFilters(
-	namespace string,
-	file string,
-	globalArchive, globalDelete map[string]string,
-) (map[NamespaceGroupVersionKind]cel.Program, map[NamespaceGroupVersionKind]cel.Program, error) {
+	namespace, file string,
+	globalArchive, globalDelete, globalArchiveOnDelete map[string]string,
+) (
+	map[NamespaceGroupVersionKind]cel.Program,
+	map[NamespaceGroupVersionKind]cel.Program,
```
ggallen, line 157: Is there a blank line or something there?  No line number, but it looks strange.

cmd/sink/main.go
```diff
@@ -94,6 +98,46 @@ func (sink *Sink) Receive(ctx context.Context, event cloudevents.Event) {
...
+				)
+			}
+			defer cancel()
+			return
+		}
+		sink.Logger.Printf(
+			"Object %s from cloudevent %s does not need to be archived after deletion\n",
+			k8sObj.GetUID(),
+			event.ID(),
+		)
```
ggallen, line 139: It looks like delete events will still go through the code for updates and creates.

cmd/sink/filters/filters.go
skoved: I'm not seeing the blank line that you see...

cmd/sink/main.go
skoved: because i forgot the return at the end of the if. I didn't have time to test the changes yesterday which is why I marked this as a draft :smiley:

cmd/sink/filters/filters.go
ggallen: Ignore it.

cmd/sink/filters/filters.go
```diff
@@ -167,27 +181,39 @@ func (f *Filters) createFilters(
...
 
 // updateNamespaceFilters updates the cel filters for namespace, from file. It differs from createNamespaceFilters by
 // deleting all filters for namespace before inserting the new ones.
-func (f *Filters) updateNamespaceFilters(namespace, file string, globalArchive, globalDelete map[string]string) error {
-	namespaceArchive, namespaceDelete, err := f.createFilters(namespace, file, globalArchive, globalDelete)
+func (f *Filters) updateNamespaceFilters(
+	namespace, file string,
+	globalArchive, globalDelete, globalArchiveOnDelete map[string]string,
+) error {
+	namespaceArchive, namespaceDelete, namespaceArchiveOnDelete, err := f.createFilters(namespace, file, globalArchive, globalDelete, globalArchiveOnDelete)
```
beerparty, line 200: Shouldn't you be checking that err returned by `createFilters` is empty before updating the filters? The code below uses the logic of deleting and recreating the filters based on what is returned by  `createFilters`, if `createFilters` returns a non-empty err all the existing filters will be removed.

skoved: no, the goal of the filter update is to change the filter to match the new filters that the user provided, even if there were errors. the error from `createFilters` is returned to the caller of `updateNamespaceFilters` and will continue up the call stack until it gets printed to the container log.

PR #434, opened 2024-09-18
Fix POSTGRES_PORT in the database secret

### LINKED ISSUE #433: POSTGRES_PORT should be b64 encoded
Currently when installing using the DEVELOPMENT.md instructions the POSTGRES_PORT value on the `kubearchive-database-secret` is mangled, as is not encoded in base64 when creating the secret.

rh-hemartin:

Resolves #433 

## Release Note

```release-note
NONE
```

### DIFF
--- a/charts/kubearchive/templates/database/database_secret.yaml
+++ b/charts/kubearchive/templates/database/database_secret.yaml
@@ -5,7 +5,7 @@ type: Opaque
 metadata:
   name: {{ tpl .Values.database.secret . }}
 data:
-  POSTGRES_PORT: "5432"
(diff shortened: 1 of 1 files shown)


ggallen:
I also have this change as part of https://github.com/kubearchive/kubearchive/pull/417.  But we should take this and get it in.

rh-hemartin:
> I also have this change as part of #417. But we should take this and get it in.

Ah, I missed that. I found the problem while testing some changes related to telemetry. I have no problem in dropping this in favor of #417 . However this is smaller and may be merged it today. Let's see.

ggallen (approved):

PR #438, opened 2024-09-18
Design KubeArchive logging integration.

### LINKED ISSUE #437: Design KubeArchive logging integration.
Provide a design for KubeArchive to integrate with various loggers.

ggallen:

Resolves #437

## Release Note

```release-note
NONE
```

### DIFF
--- a/docs/modules/ROOT/pages/design/logging-integration.adoc
+++ b/docs/modules/ROOT/pages/design/logging-integration.adoc
@@ -1,0 +1,205 @@
+= Logging Integration
+
+== Overview
+
+KubeArchive needs to support logging, but it is not a logging system itself and
+does not implement logging. Instead, KubeArchive will integrate with logging systems
+and provide URLs for retrieving log files from the logging system for a specific
+Kubernetes resource.
+
+It is important to note that logs are tied to `Pods`. When a user requests the logs
+for a Tekton `PipelineRun`, what they expect to get back are the logs attached to the
+`Pods` that were part of the `PipelineRun`. Similar cases exist for requesting logs for
+`Jobs` and `CronJobs`. KubeArchive has to be able to handle this seamlessly for the user.
+
+== Retrieving Log Information
+
+In generic terms, this can be done using the owner reference field in a resource, a
+sort of backwards recursive search. When a `PipelineRun` is deleted, all the `TaskRuns`
+associated with the `PipelineRun`, and `Pods` associated with those `TaskRuns`, are
+deleted. This is done using the owner references.
+
+KubeArchive can do things similarly. When logs for a resource are requested, a query
+is made to find all the resources that have that initial resource as an owner. Then
+each resource returned is processed similarly, eventually building up a list of log
+file links. This generic approach should work for any resource.
+
+This also implies that KubeArchive is configured correctly to support this. It must
+be configured so that the initial resource and any dependent resources, all the way
+down to and including the `Pods`, are archived.
+
+Here's a sample `KubeArchiveConfig` as an example:
+[source,yaml]
+----
+---
+apiVersion: kubearchive.kubearchive.org/v1alpha1
+kind: KubeArchiveConfig
+metadata:
+  name: kubearchive
+  namespace: test
+spec:
+  resources:
+    - deleteWhen: has(status.completionTime)
+      selector:
+        apiVersion: ""
+        kind: CronJob
+    - archiveOnDelete: true
+      selector:
+        apiVersion: ""
+        kind: Pod
+----
+So in this case, the `CronJob` is configured to be archived and deleted when
+the status contains a "completionTime" key. When that deletion happens,
+kubernetes will turn around and delete the associated `Pod`. Since we have
+configured `archiveOnDelete` for `Pods` to be true, KubeArchive will archive
+the `Pod` itself and generate the URLs for all the associated logs. The
+configuration would be similar for `PipelineRuns`, with the addition of
+the archiving of the `TaskRuns`.
+
+[IMPORTANT]
+====
+
+- KubeArchive has no responsibility for sending the logs to the logging system.
+This is all configured elsewhere and outside of KubeArchive.
+- When the `Pod` is archived, the URL for accessing the log should be generated
+and stored with it. There is no attempt to query the logging system to verify
+the existence of the log.
+
+====
+
+Here's another sample `KubeArchiveConfig` for `PipelineRuns`:
+[source,yaml]
+----
+---
+apiVersion: kubearchive.kubearchive.org/v1alpha1
+kind: KubeArchiveConfig
+metadata:
+  name: kubearchive
+  namespace: test
+spec:
+  resources:
+    - selector:
+        apiVersion: tekton.dev/v1
+        kind: PipelineRun
+      deleteWhen: has(status.completionTime)
+    - selector:
+        apiVersion: tekton.dev/v1
+        kind: TaskRun
+      archiveOnDelete: true
+    - selector:
+        apiVersion: v1
+        kind: Pod
+      archiveOnDelete: has(body.metadata.labels["tekton.dev/pipeline"])
+----
+In this case the following happens:
+
+- `PipelineRuns` are archived when they complete.
+- `TaskRuns` are archived when they are deleted.
+- `Pods` are archived when they are deleted and are also part of a Tekton `Pipeline`.
+
+== Generating Log URLs
+
+The logging system is an integration in KubeArchive. All that is required for
+the integration is the successful generation of a URL to access the log for a
+specific `Pod` in the logging system. URLs will ONLY be generated for `Pods`.
+
+To support multiple logging systems, the URLs must be able to be parameterized
+based on the logging system. This is done via a `ConfigMap` named for the
+logging system being implemented. This `ConfigMap` requires a single entry,
+LOG_URL, whose string value will be interpolated using the other variables
+defined in the `ConfigMap`.  For example, take this `ConfigMap` for Splunk:
+[source,yaml]
+----
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: kubearchive-splunk
+  namespace: kubearchive
+data:
+  POD_ID: cel:metadata.uid
+  CONTAINER_NAME: cel:spec.containers.map(m, m.name)
+  POD: 'spath "kubernetes.pod_id" | search "kubernetes.pod_id"="{POD_ID}"'
+  CONTAINER: 'spath "kubernetes.container_name" | search "kubernetes.container_name"="{CONTAINER_NAME}"'
+  LOG_URL: http://127.0.0.1:8111/app/search/search?q=search * | {POD} | {CONTAINER}
+----
+The value of each variable is either a string or a CEL expression. A value that
+begins with the prefix "cel:" will be evaluated as a CEL expression against the
+body of the cloud event (i.e. the resource) to determine the real value of that
+variable used in the substitution. For example:
+
+So when generating the logging URL to be stored when a `Pod` is archived,
+the following steps are done:
+
+1. A map is created and populated all non-CEL expression key-value pairs from the `ConfigMap`.
+1. All variables containing CEL expression variables are added to the map, and the 
+   value for each of these variables is the value returned by evaluating the CEL expression.
+1. The value for LOG_URL is then interpolated recursively using this map until no more
+   substitutions are done, resulting in the final URL to the log in the logging system.
+
+== Implementation
+
+=== Database
+
+The KubeArchive database will have a table named `log_url` with three fields:
+
+1. A `uuid` field which is a foreign key to `resource.uuid`.  
+1. A `url` field which is the URL for one of the logs.
+
+The `uuid` field should point back to a `Pod` entry in the `resource` table.
+
+=== Sink
+
+When the sink archives a `Pod`, it must take the additional step go gather all
+the log information and generate the log URL for each. These are stored in the 
+`log_url` table.
+
+The sink should first delete any existing entries in the `log_url` table for the
+`Pod` being archived. Earlier archival requests may have already created records
+in the `log_url` table, and they should be removed to avoid duplicates.
+
+The sink will also need to mount and use the `ConfigMap` for logging as detailed 
+in the <<Configuration>> section.
+
+=== CLI
+
+The CLI will implement a `logs` command similar to `kubectl logs`. 
+[source,bash]
+----
+ka logs resource name
+----
+This command will return the URLs of the logs associated with the resource of type name. For
+example:
+[source,bash]
+----
+ka logs PipelineRun generate-logs-9fkp8 -n generate-logs-pipelines
+----
+This will return all the log URLs associated with the `PipelineRun` named "generate-logs-9fkp8".
+
+The CLI will have to traverse the owner references to gather all of the logs associated
+with the given resource. Note that logs could be queried for any resource. Most
+resource will not have any logs associated with them or their descendents, but queries on
+logs for `TaskRuns` and `Pods` are possible.
+
+=== API
+
+The API needs to providate a mechanism similar to the CLI for retrieving log URLs.
+
+=== Configuration
+
+KubeArchive will need to have a `ConfigMap` for configuring logging (and potentially 
+other things). This `ConfigMap` should be named `kubearchive-config` and reside in the
+installation namespace.
+[source,yaml]
+----
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: kubearchive-config
+  namespace: kubearchive
+data:
+  logging-configmap-name: kubearchive-splunk
+----
+The sink should check this `ConfigMap`, and if it exists and the `logging-configmap-name` key
+is set, it should mount that `ConfigMap` and use it for generating log URLs.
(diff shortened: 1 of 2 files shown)


docs/design/logging-integration.adoc
rh-hemartin, line 1: I would add it to be part of the docs:

* Move this to `docs/modules/ROOT/pages/design/ilogging-integration.adoc`
* Add a new navigation section:

```
* Design
** xref: design/logging-integration.adoc[]
```

```

File: docs/design/logging-integration.adoc [hunk 1]
```diff
@@ -0,0 +1,196 @@
...
+  resources:
+    - deleteWhen: has(status.completionTime)
+      selector:
+        apiVersion: ""
+        kind: CronJob
+    - archiveOnDelete: true
+      selector:
+        apiVersion: ""
+        kind: Pod
+```
```
maruiz93, line 49: In Asciidoc [code blocks](https://docs.asciidoctor.org/asciidoc/latest/verbatim/source-blocks/) should be written like this:
```suggestion
[source,yaml]
----
---
apiVersion: kubearchive.kubearchive.org/v1alpha1
kind: KubeArchiveConfig
metadata:
  name: kubearchive
  namespace: test
spec:
  resources:
    - deleteWhen: has(status.completionTime)
      selector:
        apiVersion: ""
        kind: CronJob
    - archiveOnDelete: true
      selector:
        apiVersion: ""
        kind: Pod
----
```

docs/design/logging-integration.adoc
```diff
@@ -0,0 +1,196 @@
...
+
+In generic terms, this can be done using the owner reference field in a resource, a
+sort of backwards recursive search. When a `PipelineRun` is deleted, all the `TaskRuns`
+associated with the `PipelineRun`, and `Pods` associated with those `TaskRuns`, are
+deleted. This is done using the owner references.
+
+KubeArchive can do things similarly. When logs for a resource are requested, a query
+is made to find all the resources that have that initial resource as an owner.  Then
+each resource returned is processed similarly, eventually building up a list of log
+file links.  This generic approach should work for any resource.
```
maruiz93, line 25: Typo, double space
```suggestion
file links. This generic approach should work for any resource.
```

docs/design/logging-integration.adoc
```diff
@@ -0,0 +1,196 @@
...
+      deleteWhen: has(status.completionTime)
+    - selector:
+        apiVersion: tekton.dev/v1
+        kind: TaskRun
+      archiveOnDelete: true
+    - selector:
+        apiVersion: v1
+        kind: Pod
+      archiveOnDelete: has(body.metadata.labels["tekton.dev/pipeline"])
+```
```
maruiz93, line 91: See above comment:

```suggestion
[source, yaml]
----
---
apiVersion: kubearchive.kubearchive.org/v1alpha1
kind: KubeArchiveConfig
metadata:
  name: kubearchive
  namespace: test
spec:
  resources:
    - selector:
        apiVersion: tekton.dev/v1
        kind: PipelineRun
      deleteWhen: has(status.completionTime)
    - selector:
        apiVersion: tekton.dev/v1
        kind: TaskRun
      archiveOnDelete: true
    - selector:
        apiVersion: v1
        kind: Pod
      archiveOnDelete: has(body.metadata.labels["tekton.dev/pipeline"])
----
```

docs/design/logging-integration.adoc
```diff
@@ -0,0 +1,196 @@
...
+In this case the following happens:
+
+- `PipelineRuns` are archived when they complete.
+- `TaskRuns` are archived when they are deleted.
+- `Pods` are archived when they are deleted and are also part of a Tekton Pipeline
```
maruiz93, line 96: Missing dot at the end of the line:
```suggestion
- `Pods` are archived when they are deleted and are also part of a Tekton Pipeline.
```

docs/design/logging-integration.adoc
```diff
@@ -0,0 +1,196 @@
...
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: kubearchive-splunk
+  namespace: kubearchive
+data:
+  LOG_URL=https://splunk-host/app/search/search?q=search *| spath "kubernetes.pod_id" | search "kubernetes.pod_id"="${POD_ID}"
+  POD_ID=".body.pod_id"
+```
```
maruiz93, line 119: See above comment:

```suggestion
[source, yaml]
----
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: kubearchive-splunk
  namespace: kubearchive
data:
  LOG_URL=https://splunk-host/app/search/search?q=search *| spath "kubernetes.pod_id" | search "kubernetes.pod_id"="${POD_ID}"
  POD_ID=".body.pod_id"
----
```

docs/design/logging-integration.adoc
```diff
@@ -0,0 +1,196 @@
...
+
+== Implementation
+
+=== Database
+
+The KubeArchive database will have a table named `log_url` with three fields:
+
+1. An `id` field as a primary key.
+1. A `resource_id` field which is a foreign key to `resource.id`.  
+1. A `url` field which is the URL for one of of logs.
```
maruiz93, line 141: There is a typo here:
```suggestion
1. A `url` field which is the URL for one of the logs.
```

BTW, I have a question here, let's say we have a user that uses splunk for example, so we have un our database the path to the specific URL to find the logs of the pod, something like this: `http://<the-splunk-host>:<the-splunk-port>/<path>?<queryfilter>`

Now, let's say that the user need to update something about the URL, can be the host, or maybe the port, or maybe even the protocol, then, all the stored URLs will be outdated and stop working.

I think, that it may be more useful to build the URL dynamically based on the content of the `ConfigMap` that stores the URL.

docs/design/logging-integration.adoc
```diff
@@ -0,0 +1,196 @@
...
+
+The `resource_id` field should point back to a `Pod` entry in the `resource` table.
```
maruiz93, line 143: Shouldn't we also include a way to provide the relationship of "owner-owned" among the different resources and their pods? For example, between a `Deployment` and their `Pods`, between a `CronJob` and their `Pods` between a `PipelineRun` and their `Pods`.

Consider that currently with `kubectl` we can run `kubectl logs deploy/<my-deploy>` and retrieve the logs of all the pods that are currently running (or that ran in the past) this deployment. I guess that KubeArchive should provide a pointer to the same logs based on the same parameters.

The deployment and their pods are related based on a labelselector. We currently have all the information in the db, we have the labelSelector of a deployment and the labels of the pods but that query might be tricky, so I suggest adding an additional table to relate these objects more easily.

docs/design/logging-integration.adoc
```diff
@@ -0,0 +1,196 @@
...
+
+The sink will also need to mount and use the `ConfigMap` for logging as detailed 
+in the <<Configuration>> section.
+
+=== CLI
+
+The CLI will implement a `log` command similar to `kubectl log`. 
+```
+ka log resource name
+```
```
maruiz93, line 163:
```suggestion
[source, bash]
----
ka log resource name
----
```

docs/design/logging-integration.adoc
```diff
@@ -0,0 +1,196 @@
...
+This command will return the URLs of the logs associated with the resource of type name. For
+example:
+```
+ka log PipelineRun generate-logs-9fkp8 -n generate-logs-pipelines
+```
```
maruiz93, line 168:
```suggestion
[source, bash]
----
ka log PipelineRun generate-logs-9fkp8 -n generate-logs-pipelines
----
```

docs/design/logging-integration.adoc
```diff
@@ -0,0 +1,196 @@
...
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: kubearchive-config
+  namespace: kubearchive
+data:
+  logging-configmap-name: kubearchive-splunk
+```
```
maruiz93, line 194:
```suggestion
[source, yaml]
----
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: kubearchive-config
  namespace: kubearchive
data:
  logging-configmap-name: kubearchive-splunk
----
```

maruiz93 (changes requested):

docs/design/logging-integration.adoc
```diff
@@ -0,0 +1,196 @@
...
+The sink should first delete any existing entries in the `log_url` table for the
+`Pod` being archived. Earlier archival requests may have already created records
+in the `log_url` table, and they should be removed to avoid duplicates.
+
+The sink will also need to mount and use the `ConfigMap` for logging as detailed 
+in the <<Configuration>> section.
+
+=== CLI
+
+The CLI will implement a `log` command similar to `kubectl log`.
```
maruiz93, line 160: Typo here, the command is `kubectl logs` not `kubectl log`
```suggestion
The CLI will implement a `logs` command similar to `kubectl logs`. 
```

docs/design/logging-integration.adoc
```diff
@@ -0,0 +1,196 @@
...
+=== CLI
+
+The CLI will implement a `log` command similar to `kubectl log`. 
+```
+ka log resource name
+```
+This command will return the URLs of the logs associated with the resource of type name. For
+example:
+```
+ka log PipelineRun generate-logs-9fkp8 -n generate-logs-pipelines
```
maruiz93, line 167:
```suggestion
ka logs PipelineRun generate-logs-9fkp8 -n generate-logs-pipelines
```

maruiz93 (changes requested):

docs/design/logging-integration.adoc
ggallen: Fixed.

docs/design/logging-integration.adoc
ggallen: Fixed.

docs/design/logging-integration.adoc
ggallen: Fixed.

docs/design/logging-integration.adoc
ggallen: Fixed.

docs/design/logging-integration.adoc
ggallen: Typo is fixed.

docs/design/logging-integration.adoc
ggallen: Fixed.

docs/design/logging-integration.adoc
ggallen: Fixed.

docs/design/logging-integration.adoc
ggallen: Fixed.

docs/design/logging-integration.adoc
ggallen: Fixed.

docs/design/logging-integration.adoc
ggallen: Fixed.

docs/design/logging-integration.adoc
ggallen: The more I think about this the more I am think that generating the log URL at archive time is the way to do it.  We've come up with a number of ways these URLs can be broken based on settings.

- The hostname changes
- The configuration of the logging system changes.
- The setting for LOG_URL or other variables in the `ConfigMap` changes.

These can all break it.  I conclude that there's no perfect solution here.  Even splitting the "prefix" as you suggest isn't perfect either. It's broken just as easily as anything else., and here's one scenario:

What if the there is new logging host, but the old logging host is still around, available, and contains the old logs?

In this scenario all the old logging links would be broken if we allowed the "prefix" to change.

I think the safest thing, and the best we can do, is to generate the logs at archive time and store them in the `log_url` table. It also keeps things simple as the log URL generation is all done in the sink, instead of being split between the sink and the API.

If there are changes that make the log URLs incorrect, admins have other ways to fix them.  They can write SQL queries to update the URLs with the appropriate changes. Or, in the case of a hostname change as you mentioned, then can either do that or provide a DNS entry to point to the new server.  There are lots of different solutions, but this problem is outside of KubeArchive.

docs/design/logging-integration.adoc
ggallen: Fixed.

docs/design/logging-integration.adoc
ggallen: Fixed.

docs/modules/ROOT/pages/design/logging-integration.adoc
```diff
@@ -0,0 +1,202 @@
...
+
+=== Sink
+
+When the sink archives a `Pod`, it must take the additional step go gather all
+the log information and generate the log URL for each. These are stored in the 
+`log_url` table.
+
+The sink should first delete any existing entries in the `log_url` table for the
+`Pod` being archived. Earlier archival requests may have already created records
+in the `log_url` table, and they should be removed to avoid duplicates.
```
rh-hemartin, line 156: If there is a single entry per `Pod`, we could use the `uid` of the `Pod` as `id` on the `log_url` table, so our deletes/updates are specific. This would remove the need for deletion, we just update the record.

ggallen: There is not necessarily a single entry per `Pod`.  A `Pod` can have multiple containers, each with it's own log.

ggallen: But as I think about it we may not be able to tell the difference.  We may only be able to have a single log URL for all the `Pod` logs.  In this case the update would work, and we could use an upsert.

ggallen: And if there is only a single record, do we even want another table?  We could just add it as a field in the resource table, and it would only be filled in for `Pods`.

docs/design/logging-integration.adoc
maruiz93: I agree, you have made a fair point and I am conviced of storing it the way you suggest, on "archiving" time. BTW, I think that this is a concern that might be brought up in the future so we should write this decission somewhere in the docs.

docs/design/logging-integration.adoc
maruiz93: > I'm not sure how your comment relates to the log_url table. That table is simply a list of log URLs that tie back to pods in the resource table.

Yes, sorry for introducing confusion, my concern here is about the need of an extra table, it's not related to the `log_url` table :)

> This may be true, but they are also related by "owner-owned". But there is a ReplicaSet in between, so that will have to be archived as well. 

Ok so we have the `ownerReference` in the `metadata` of the different resources, and based on that we can go from a pod to an owner (like a `Deployment`) directly or indirectly (in the case of the `Deployment` we need to go through the `ReplicaSet` as you mentioned). 
But my point here is that the user may want to provide the name of a `Deployment` to get the logs of its "owned" pods.
The problem I see is going from the `Deployment` to the `Pods` and not the other way around. We currently have some information in the database that allow us to do that, the `labelSelector` of the `Deployment` and the `labels` in the `metadata` of the `Pods` but I believe that will be somewhat unefficient.

So my suggestion was to think of a new relational database to store the `owner` -> `owned` relationship instead to be able to extract the log references based on the name of an "owner"

maruiz93 (commented):
Added a concern about the owner/owned resources

docs/design/logging-integration.adoc
ggallen: @maruiz93, the JSON indexes should allow for easy retrieval of the owner relationship data without the need for an extra table.  These queries are very fast.

Yes, you may have to make multiple queries to get the entire list of logs.  But it is doable. But one thing to consider is possible writing a stored procedure to do this work.

docs/modules/ROOT/pages/design/logging-integration.adoc
ggallen: Each pod can have multiple logs, and they are differentiated by container name, which has to be unique within the pod.

The end result if that there can be multiple logs per pod, and hence multiple records in the `log_url` table.  And since there can be multiple records, a table is required. See https://github.com/kubearchive/kubearchive/pull/449.

As for deleting, there is a foreign key constraint on the `log_url` table with cascading delete, so if an entry is deleted from the `resource` table, the corresponding entries in the `log_url` table are deleted as well.

docs/design/logging-integration.adoc
ggallen: @maruiz93, I just realized this is already mentioned in this document and flagged as important:
```
[IMPORTANT]
====
- KubeArchive has no responsibility for sending the logs to the logging system.
This is all configured elsewhere and outside of KubeArchive.
- When the `Pod` is archived, the URL for accessing the log should be generated
and stored with it. There is no attempt to query the logging system to verify
the existence of the log.
====
```
But it should probably also be included in any user documentation that is written for logging.

docs/design/logging-integration.adoc
maruiz93: For future reference, I also created an issue with a suggested stored procedure for doing the work of relate a resource with its owned `Pods` to test it in the future in case this could improve the performance: https://github.com/kubearchive/kubearchive/issues/492

maruiz93 (approved):

docs/design/logging-integration.adoc
ggallen: This is an implementation decision that should be made when the implementation is done.  Both should be tried.  I think the stored procedure might be simpler than doing it in code.

PR #445, opened 2024-09-19
Update DEVELOPMENT.md to reflect the latest state of installing.

### LINKED ISSUE #442: Update DEVELOPMENT.md to reflect the latest state of installing.
After https://github.com/kubearchive/kubearchive/pull/407 the documentation for installing is now out of date.

This issue is to update it.

ggallen:

Resolves #442

## Release Note

```release-note
NONE
```


DEVELOPMENT.md
```diff
@@ -42,53 +42,9 @@ Install these tools:
...
-   helm install kubearchive charts/kubearchive --create-namespace -n kubearchive \
-       --set-string apiServer.image=$(ko build github.com/kubearchive/kubearchive/cmd/api) \
-       --set-string sink.image=$(ko build github.com/kubearchive/kubearchive/cmd/sink) \
-       --set-string operator.image=$(ko build github.com/kubearchive/kubearchive/cmd/operator)
-   ```
-1. Check that the KubeArchive deployments are Ready:
+## Install KubeArchive, its dependencies, and initialize the database
    ```bash
-   kubectl get -n kubearchive deployments
+   hack/quick-install.sh
```
maruiz93, line 47: 4 things related to this change:

1. Can you please add in line 14 that the `kubectl` version needed is >1.31 ?
2. If we are not providing the `helm install` command to deploy kubearchive, `helm upgrade` makes no sense anymore as we don't have a helm chart deployed anymore (just a plain yaml). Should we than provide different instructions? Maybe a `kind delete cluster` and then reapply the quick-install?
3. Same applies for the section "Uninstall KubeArchive"
4. Now the instructions for testing the API Server with curl and the instructions to use the CLI don't work anymore (because we aren't creating the test namespace and resources, i.e, we don't have a service account with permissions to query a k8s resource). To solve this I can create a separate test script.

maruiz93 (changes requested):

DEVELOPMENT.md
ggallen: @maruiz93 , I have addressed 1, 2, and 3.

How do you want to handle 4?

hack/kubearchive-install.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
+# Copyright KubeArchive Authors
+# SPDX-License-Identifier: Apache-2.0
+
+# Install KubeArchive by itself.  Use by quick install and also to update existing
```
maruiz93, line 5: Typo
```suggestion
# Install KubeArchive by itself. Used by quick install and also to update existing
```

hack/kubearchive-delete.sh
```diff
@@ -0,0 +1,15 @@
...
+set -o xtrace
+
+YAML=$(mktemp --suffix=.yaml -t kubearchive-XXX)
+
+helm template kubearchive charts/kubearchive -n kubearchive \
+    --include-crds \
+    --set "global.production=true" > ${YAML}
+kubectl delete -n kubearchive -f ${YAML}
+
+rm -f ${YAML}
```
maruiz93, line 15: What about removing the whole namespace for deleting kubearchive?

```suggestion
kubectl delete ns kubearchive
```

maruiz93 (changes requested):

hack/kubearchive-delete.sh
ggallen: Deleting the namespace won't delete any cluster-level resources that got installed.

And the namespace is getting deleted anyway as `global.production` is set to `true`.

hack/kubearchive-install.sh
ggallen: Fixed.

DEVELOPMENT.md
maruiz93: I've created https://github.com/kubearchive/kubearchive/pull/456

As it's shorter we can merge it before this one. Other possibility is that you include those changes in this PR.

hack/kubearchive-delete.sh
maruiz93: Sorry for reopening this thread but I have another question related to this script:

We are removing the resources that are created from the helm chart, but in this `DEVELOPMENT` instructions we apply other test resources like the `log-generators` or the `test` users I'm introducing in https://github.com/kubearchive/kubearchive/pull/456. When "deleting" kubeArchive, should we also consider removing these stuff? Or maybe not because they aren't kubeArchive components?

It's probably ok this way, but I just wanted to raise my concern so you could also think about it.

hack/quick-install.sh
```diff
@@ -15,28 +18,4 @@ kubectl apply -f https://github.com/knative/eventing/releases/download/knative-$
...
-    echo -n .
-    sleep 0.5
-done
-echo .
-
-kubectl get -n kubearchive deployments
-
-# Kill all background jobs, including the port-forward started earlier.
-trap 'kill $(jobs -p)' EXIT
+hack/kubearchive-install.sh
```
beerparty, line 21: The other call to a script in this code uses the syntax `bash <dir>/install.sh` (line 11), can you do the same here to make them consistent?

```suggestion
bash hack/kubearchive-install.sh
```

hack/kubearchive-install.sh
```diff
@@ -0,0 +1,50 @@
...
+done
+
+# Now make sure all the new pods are ready.
+kubectl -n kubearchive wait pod --all --for=condition=ready --timeout=90s
+
+# Make sure webhooks are up and running.
+LOCAL_PORT=8443
+kubectl -n kubearchive port-forward service/kubearchive-operator-webhooks ${LOCAL_PORT}:443 >& /dev/null &
+
+# Wait for $LOCAL_PORT to become available.
```
beerparty, line 40: Since you have echos in the loop to indicate progress, you should have an echo before the start so anyone watching knows what is happening.

```suggestion
echo "Waiting for port forwarding on $LOCAL_PORT to become available."
```

hack/kubearchive-install.sh
```diff
@@ -0,0 +1,50 @@
...
+    --include-crds \
+    --set "global.production=true" > /tmp/kubearchive-not-resolved.yaml
+ko resolve -f /tmp/kubearchive-not-resolved.yaml --base-import-paths > ${YAML}
+kubectl apply -n kubearchive -f ${YAML}
+
+rm -f ${YAML}
+
+kubectl -n kubearchive rollout status deployment --timeout=90s
+
+# Wait for all the existing pods to terminate.
```
beerparty, line 28: Since the script will be waiting here, you should have an echo so anyone watching knows what is happening.

```suggestion
echo "Waiting for all the existing pods to terminate."
```

hack/kubearchive-install.sh
```diff
@@ -0,0 +1,50 @@
...
+for pod in ${PODS}; do
+    kubectl -n kubearchive wait pod $pod --for=delete --timeout=90s
+done
+
+# Now make sure all the new pods are ready.
```
beerparty, line 33: Since the script will be waiting here, you should have an echo so anyone watching knows what is happening.

```suggestion
echo "Waiting for all the new pods to start up."
```

hack/kubearchive-install.sh
```diff
@@ -0,0 +1,50 @@
...
+bash cmd/operator/generate.sh
+helm template kubearchive charts/kubearchive -n kubearchive \
+    --include-crds \
+    --set "global.production=true" > /tmp/kubearchive-not-resolved.yaml
+ko resolve -f /tmp/kubearchive-not-resolved.yaml --base-import-paths > ${YAML}
+kubectl apply -n kubearchive -f ${YAML}
+
+rm -f ${YAML}
+
+kubectl -n kubearchive rollout status deployment --timeout=90s
```
beerparty, line 26: I think this would be a good place to echo what is happening in the script, before the waits start.

```suggestion
echo "Start a rollout of the KubeArchive deployment."
kubectl -n kubearchive rollout status deployment --timeout=90s
```

beerparty (changes requested):

rh-hemartin:
These changes do not address the debugging with IDE. What will be the process for those?

hack/kubearchive-delete.sh
ggallen: IMO those things are completely separate from KubeArchive.

maruiz93:
> These changes do not address the debugging with IDE. What will be the process for those?
I thought that we would stick with the current process with `helm install`

hack/kubearchive-install.sh
ggallen: This information is in the output, first because the command is echoed and second because `kubectl` prints a status.

```
+ kubectl -n kubearchive rollout status deployment --timeout=90s
Waiting for deployment "kubearchive-api-server" rollout to finish: 0 of 1 updated replicas are available...
deployment "kubearchive-api-server" successfully rolled out
Waiting for deployment "kubearchive-operator" rollout to finish: 0 of 1 updated replicas are available...
deployment "kubearchive-operator" successfully rolled out
deployment "kubearchive-sink" successfully rolled out 
```

hack/kubearchive-install.sh
ggallen: The command is echoed here so the user will know what is going on, and `kubectl` prints the status.

```
+ kubectl -n kubearchive wait pod --all --for=condition=ready --timeout=90s
pod/kubearchive-api-server-759b549d69-wtxds condition met
pod/kubearchive-operator-6f7b5c8cd9-ghg2h condition met
pod/kubearchive-sink-66dc8fdb68-zvdlb condition met
```

hack/kubearchive-install.sh
ggallen: The command is echoed here so the user will know what is going on, and kubectl prints the status.

I don't have an example output here as their were no terminating pods from my run.

hack/quick-install.sh
ggallen: Fixed.

hack/kubearchive-install.sh
ggallen: Fixed.

hack/kubearchive-install.sh
beerparty: OKAY

hack/kubearchive-install.sh
beerparty: OKAY

DEVELOPMENT.md
ggallen: @maruiz93 , I merged https://github.com/kubearchive/kubearchive/pull/456.  Can this conversation be resolved?

hack/kubearchive-install.sh
beerparty: OKAY

beerparty (approved):

ggallen:
> > These changes do not address the debugging with IDE. What will be the process for those?
> 
> I thought that we would stick with the current process with `helm install`
> 
> EDIT: We need to include the instruction of doing an uninstall of kubearchive first.
> 
> @ggallen can you add that to the `DEVELOPMENT.md`?

If the current instructions still work, then I'm fine with that.  I'm not sure what changes would be needed otherwise as I've not used the debugger at all. I'd need someone else to put those together, maybe in a separate PR.

@maruiz93 , I'm not sure what you mean about adding "the instruction of doing an uninstall of kubearchive first".  I've added and uninstall section, so that is already there. If you're talking about adding it as part of debugging, then there not one currently there, so why would we need it now?

maruiz93 (approved):

PR #450, opened 2024-09-20
Use slog on KubeArchive

### LINKED ISSUE #423: Implement logging using slog.
KubeArchive is currently using the standard log package for logging.  This does not provide log levels, and errors can be hard to find.  Here's an example:
```
2024/09/16 11:18:00 http: superfluous response.WriteHeader call from go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp
/internal/request.(*RespWriterWrapper).writeHeader (resp_writer_wrapper.go:78)
2024/09/16 11:18:00.135948 Received cloudevent:  07d1a64b-a2b8-4391-a959-f55ef60bdf4d
2024/09/16 11:18:00.136290 Cloudevent 07d1a64b-a2b8-4391-a959-f55ef60bdf4d contains all required fields. Checking if its object ff
f80249-8dae-494d-83ed-8aa186bcc3df needs to be archived
2024/09/16 11:18:00.136350 Writing object fff80249-8dae-494d-83ed-8aa186bcc3df from cloudevent 07d1a64b-a2b8-4391-a959-f55ef60bdf4
d into the database
2024/09/16 11:18:00.137579 Failed to write object fff80249-8dae-494d-83ed-8aa186bcc3df from cloudevent 07d1a64b-a2b8-4391-a959-f55ef60bdf4d to the database: write to database failed: pq: relation "kubearchive.resource" does not exist
2024/09/16 11:18:00 http: superfluous response.WriteHeader call from go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp/internal/request.(*RespWriterWrapper).writeHeader (resp_writer_wrapper.go:78)
2024/09/16 11:18:00.140040 Received cloudevent:  010a72fd-8604-410b-a69c-de3c8d324976
2024/09/16 11:18:00.140293 Cloudevent 010a72fd-8604-410b-a69c-de3c8d324976 contains all required fields. Checking if its object e700ef49-146a-4dcc-8849-290c476a94df needs to be archived
2024/09/16 11:18:00.140322 Object e700ef49-146a-4dcc-8849-290c476a94df from cloudevent 010a72fd-8604-410b-a69c-de3c8d324976 does not need to be archived

```
There is an error embedded there, and it's hard to see.

We should be using slog instead, and all of the logging in KubeArchive should have a level.

rh-hemartin:

Resolves #423 

## Release Note

```release-note
NONE
```

## Notes for Reviewers

* I removed the logger parameter forwarding in the sink and went with using the default logger. Its simpler. If we need to pass a logger for whatever reason, then we can add it again.

### DIFF
--- a/cmd/api/main.go
+++ b/cmd/api/main.go
@@ -5,7 +5,7 @@ package main
 
 import (
 	"fmt"
-	"log"
+	"log/slog"
 	"net/http"
 	"os"
 	"time"
@@ -16,6 +16,7 @@ import (
 	"github.com/kubearchive/kubearchive/cmd/api/routers"
 	"github.com/kubearchive/kubearchive/pkg/cache"
 	"github.com/kubearchive/kubearchive/pkg/database"
+	"github.com/kubearchive/kubearchive/pkg/middleware"
 	"github.com/kubearchive/kubearchive/pkg/observability"
 	"go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin"
 	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
@@ -55,7 +56,9 @@ func getKubernetesClient() *kubernetes.Clientset {
 }
 
 func NewServer(k8sClient kubernetes.Interface, controller routers.Controller, cache *cache.Cache, cacheExpirations *routers.CacheExpirations) *Server {
-	router := gin.Default()
+	router := gin.New()
+	router.Use(gin.Recovery())
+	router.Use(middleware.Logger)
 	router.Use(otelgin.Middleware("")) // Empty string so the library sets the proper server
 
 	apiGroup := router.Group("/api")
@@ -88,28 +91,30 @@ func NewServer(k8sClient kubernetes.Interface, controller routers.Controller, ca
 }
 
 func main() {
-	log.Printf("Starting KubeArchive API with version '%s', commit SHA '%s', built '%s'", version, commit, date)
+	slog.Info("Starting KubeArchive API", "version", version, "commit", commit, "built", date)
 	err := observability.Start(otelServiceName)
 	if err != nil {
-		log.Printf("Could not start opentelemetry: %s", err)
+		slog.Info("Could not start opentelemetry", "error", err.Error())
 	}
 
 	cacheExpirations, err := getCacheExpirations()
 	if err != nil {
-		log.Fatal(err)
+		slog.Error(err.Error())
 	}
 	memCache := cache.New()
 
 	db, err := database.NewDatabase()
 	if err != nil {
-		log.Fatalf("Could not connect to database: %s", err)
+		slog.Error("Could not connect to database", "error", err.Error())
+		os.Exit(1)
 	}
 	controller := routers.Controller{Database: db, CacheConfiguration: *cacheExpirations}
 
 	server := NewServer(getKubernetesClient(), controller, memCache, cacheExpirations)
 	err = server.router.RunTLS("0.0.0.0:8081", "/etc/kubearchive/ssl/tls.crt", "/etc/kubearchive/ssl/tls.key")
 	if err != nil {
-		log.Printf("Could not run server on localhost: %s", err)
+		slog.Error("Could not run server on localhost", "error", err.Error())
+		os.Exit(1)
 	}
 }
 
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -7,11 +7,13 @@ import (
 	"crypto/tls"
 	"flag"
 	"log"
+	"log/slog"
 	"os"
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
 	// to ensure that exec-entrypoint and run can make use of them.
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
+	"k8s.io/klog/v2"
 
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
@@ -22,6 +24,7 @@ import (
 	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 
+	"github.com/go-logr/logr"
 	kubearchiveapi "github.com/kubearchive/kubearchive/cmd/operator/api/v1alpha1"
 	"github.com/kubearchive/kubearchive/cmd/operator/internal/controller"
 	"github.com/kubearchive/kubearchive/pkg/observability"
@@ -33,11 +36,10 @@ import (
 const otelServiceName = "kubearchive.operator"
 
 var (
-	version  = "main"
-	commit   = ""
-	date     = ""
-	scheme   = runtime.NewScheme()
-	setupLog = ctrl.Log.WithName("setup")
+	version = "main"
+	commit  = ""
+	date    = ""
+	scheme  = runtime.NewScheme()
 )
 
 func init() {
@@ -49,6 +51,7 @@ func init() {
 }
 
 func main() {
+	klog.SetSlogLogger(slog.Default()) // klog is used by the election leader process
 	err := observability.Start(otelServiceName)
 	if err != nil {
 		log.Printf("Could not start OpenTelemetry: %s\n", err)
@@ -74,7 +77,7 @@ func main() {
 	opts.BindFlags(flag.CommandLine)
 	flag.Parse()
 
-	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
+	ctrl.SetLogger(logr.FromSlogHandler(slog.Default().Handler()))
 
 	// if the enable-http2 flag is false (the default), http/2 should be disabled
 	// due to its vulnerabilities. More specifically, disabling http/2 will
@@ -83,7 +86,7 @@ func main() {
 	// - https://github.com/advisories/GHSA-qppj-fm5r-hxr3
 	// - https://github.com/advisories/GHSA-4374-p667-p6c8
 	disableHTTP2 := func(c *tls.Config) {
-		setupLog.Info("disabling http/2")
+		slog.Info("disabling http/2")
 		c.NextProtos = []string{"http/1.1"}
 	}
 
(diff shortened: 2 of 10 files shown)


cmd/api/main.go
```diff
@@ -86,28 +89,30 @@ func NewServer(k8sClient kubernetes.Interface, controller routers.Controller, ca
 }
 
 func main() {
-	log.Printf("Starting KubeArchive API with version '%s', commit SHA '%s', built '%s'", version, commit, date)
+	slog.Info("Starting KubeArchive API", "version", version, "commit", commit, "built", date)
 	err := observability.Start(otelServiceName)
 	if err != nil {
-		log.Printf("Could not start opentelemetry: %s", err)
+		slog.Info("Could not start opentelemetry:", "error", err.Error())
```
ggallen, line 95: Why does this message have a colon?  None of the others seem to.

pkg/middleware/logging.go
```diff
@@ -0,0 +1,37 @@
+// Copyright KubeArchive Authors
```
ggallen, line 1: What is this package for?

cmd/operator/main.go
```diff
@@ -7,11 +7,13 @@ import (
 	"crypto/tls"
 	"flag"
 	"log"
+	"log/slog"
 	"os"
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
 	// to ensure that exec-entrypoint and run can make use of them.
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
+	"k8s.io/klog/v2"
```
ggallen, line 16: Why is this package added?

cmd/operator/main.go
```diff
@@ -22,6 +24,7 @@ import (
 	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 
+	"github.com/go-logr/logr"
```
ggallen, line 27: Why is this package added?

ggallen (commented):
What does the format of the log messages look like?

pkg/middleware/logging.go
rh-hemartin: `gin` does not support `slog`, so this replaces their `Logger` middleware by our own, which uses `slog`.

cmd/operator/main.go
rh-hemartin: If we want to use `slog` on the operator we need to adapt some stuff, this package is needed because we need to configure `klog` (used by the election process) to use `slog`.

cmd/operator/main.go
rh-hemartin: Same as `klog`, we need it so we can use `slog` on internal parts of the operator.

cmd/api/main.go
rh-hemartin: No reason, I just forgot to remove this one.

pkg/middleware/logging.go
```diff
@@ -0,0 +1,37 @@
...
+	}
+
+	slog.Info(
+		"Served request",
+		"method", c.Request.Method,
+		"status", c.Writer.Status(),
+		"latency", latency,
+		"client", c.ClientIP(),
+		"path", path,
+		"errors", c.Errors.ByType(1).String(),
```
maruiz93, line 35: Should we use `Info` method if the served request response is >= 500 with errors?

Maybe we can do something like:

```suggestion
        var logLevel int
        if c.Writer.Status() >= 500 {
		logLevel = slog.LevelError
	} else {
		logLevel = slog.LevelInfo
	}
       
	slog.Log(
	        c, logLevel,
		"Served request",
		"method", c.Request.Method,
		"status", c.Writer.Status(),
		"latency", latency,
		"client", c.ClientIP(),
		"path", path,
		"errors", c.Errors.ByType(1).String(),
```

maruiz93 (changes requested):
Maybe we can address this later, but thinking of proper logs in an API Server, it's useful to have an ID for correlating the logs related to the same request.

Considering that we are making KubeArchive OpenTelemetry compatible, I guess that we could take advantage of the trace and the span when logging.

Maybe we could take advantage of the [trace.SpanFromContext function](https://pkg.go.dev/go.opentelemetry.io/otel/trace#SpanFromContext) and add it to the logged attributes.

We could do that in a slog custom Handler for the default logger that can check based on the the otel configuration (if we have it enabled or not)

pkg/middleware/logging.go
rh-hemartin: I think it should be the function that returns the 500 that logs at the error level. This middleware is just "neutral" information.

maruiz93 (approved):

rh-hemartin:
> What does the format of the log messages look like?

```
2024/09/23 13:20:43 INFO Starting KubeArchive Sink version=v0.1.0 commit=023005ba461e47e5374e92da2e598b8dc1f4b4b3 built=2024-09-23T15:20:31+02:00
2024/09/23 13:20:43 INFO Successfully connected to the database
```

pkg/middleware/logging.go
```diff
@@ -0,0 +1,37 @@
...
+
+func Logger(c *gin.Context) {
+	start := time.Now()
+	c.Next()
+	end := time.Now()
+
+	latency := end.Sub(start)
+	if latency > time.Minute { // Avoid very large latency number string
+		latency = latency.Truncate(time.Second)
+	}
```
beerparty, line 20: You should probably be using `Round` instead of `Truncate` here.

Why are you only truncating the latency if it is more than a minute? Comparing the results is going to look weird if some of the values have nano-second accuracy and some do not. As I understand it, the default display of duration (with a "latency" label) in Go looks like this:

`latency: 1m26.550805438s`

With your conditional Truncation (or Round), you will end up comparing strings like this:

```
latency: 1m10s
latency: 26.550805438s
```

If you are worried about the final string being too long, why not shorten everything to the same precision? If you did a `latency = latency.Round(time.Millisecond)` on all values, you will truncate everything to seconds + 3 decimal points and end up with this:

```
latency: 1m10.123s
latency: 26.550s
```

beerparty (changes requested):

skoved (approved):

pkg/middleware/logging.go
rh-hemartin: If we round to milisecond we loose all the microsecond precision. I prefer to remove the `Truncate` and leave it as is. Let's see how it results.

cmd/sink/main.go
```diff
@@ -61,89 +57,107 @@ func NewSink(db database.DBInterface, logger *log.Logger, filters *filters.Filte
...
 }
 
 // Processes incoming cloudevents and writes them to the database
 func (sink *Sink) Receive(ctx context.Context, event cloudevents.Event) {
-	sink.Logger.Println("Received cloudevent: ", event.ID())
+	slog.Info("Received cloudevent", "id", event.ID())
 	k8sObj, err := models.UnstructuredFromByteSlice(event.Data())
 	if err != nil {
-		sink.Logger.Printf("Cloudevent %s is malformed and will not be processed: %s\n", event.ID(), err)
+		slog.Error("Cloudevent is malformed and will not be processed", "id", event.ID(), "err", err)
```
beerparty, line 88: Errors indicate a failure - since you are not failing at this point, shouldn't this be a warning (`slog.Warn`) instead of an error?

cmd/sink/filters/filters.go
```diff
@@ -248,93 +248,92 @@ func (f *Filters) deleteFiltersWithMatcher(matcher NamespaceMatcher) {
...
-		logger.Println("Creating filters for namespace:", fileName)
+		slog.Info("Creating filters for namespace", "namespace", fileName)
 		if event.Has(fsnotify.Write) {
 			err = f.updateNamespaceFilters(fileName, event.Name, globalArchive, globalDelete, globalArchiveOnDelete)
 		} else {
 			err = f.createNamespaceFilters(fileName, event.Name, globalArchive, globalDelete, globalArchiveOnDelete)
 		}
 		if err != nil {
-			logger.Printf("Problem creating filters for namespace %s: %s\n", fileName, err)
+			slog.Error("Problem creating filters for namespace", "namespace", fileName, "err", err)
```
beerparty, line 302: Shouldn't you have an exit here? If you don't, you'll have an error message "Problem creating filters for namespace" immediately followed by an info message "Created filters for namespace". If you are going to keep processing even with this failure, I suggest you (1) make this message a warning (`slog.Warn`) and (2) put the info message in an else.

cmd/sink/filters/filters.go
```diff
@@ -248,93 +248,92 @@ func (f *Filters) deleteFiltersWithMatcher(matcher NamespaceMatcher) {
...
+			slog.Info("Could not read global filters", "err", err)
 			break
 		}
 		if fileName == globalKey {
-			logger.Println("Creating global filters")
+			slog.Info("Creating global filters")
 			err = f.changeGlobalFilters(filePaths, globalArchive, globalDelete, globalArchiveOnDelete)
 			if err != nil {
-				logger.Println("Problem creating global filters:", err)
+				slog.Error("Problem creating global filters", "err", err)
```
beerparty, line 290: Shouldn't you have an exit here? If you don't, you'll have an error message "Problem creating global filters" immediately followed by an info message "Successfully created global filters". If you are going to keep processing even with this failure, I suggest you (1) make this message a warning (`slog.Warn`) and (2) put the info message in an else.

beerparty (changes requested):

cmd/sink/filters/filters.go
rh-hemartin: A warning will slip easier as something we don't need to pay attention to, so I think its better to have it as an error. About the logging statements themselves... @skoved

cmd/sink/filters/filters.go
rh-hemartin: A warning will slip easier as something we don't need to pay attention to, so I think its better to have it as an error. About the logging statements themselves... @skoved

cmd/sink/main.go
rh-hemartin: A warning could be understood as something we don't need to address, I don't know if here that would be true. If running we get this problem a lot and its just life, we can use Warn instead.

cmd/sink/filters/filters.go
```diff
@@ -248,93 +248,92 @@ func (f *Filters) deleteFiltersWithMatcher(matcher NamespaceMatcher) {
...
 			err = f.updateNamespaceFilters(fileName, event.Name, globalArchive, globalDelete, globalArchiveOnDelete)
 		} else {
 			err = f.createNamespaceFilters(fileName, event.Name, globalArchive, globalDelete, globalArchiveOnDelete)
 		}
 		if err != nil {
-			logger.Printf("Problem creating filters for namespace %s: %s\n", fileName, err)
+			slog.Error("Problem creating filters for namespace", "namespace", fileName, "err", err)
 		}
-		logger.Println("Created filters for namespace:", fileName)
+		slog.Info("Created filters for namespace", "namespace", fileName)
```
beerparty, line 304: This log message needs to go into an `else`, otherwise you will have a failure message immediately followed by a success message.

```suggestion
        else {
		    slog.Info("Created filters for namespace", "namespace", fileName)
		}
```

rh-hemartin: Added

beerparty (approved):

PR #454, opened 2024-09-22
Add MySQL support.

ggallen:
## Which Issue(s) this Pull Request Resolves

Resolves #453

## Release Note

```release-note
NONE
```

Issue #453, opened 2024-09-22
Add MySQL support.


ggallen:
Add MySQL support to KubeArchive.

rh-hemartin:
Why do we need MySQL support?

ggallen:
> Why do we need MySQL support?

Because it increases our potential user base?  And because it's not hard?

```
As Postgres gains more momentum, it becomes harder to pick between Postgres and MySQL. MySQL is probably still the world's most popular open source database by install base. While Postgres positions itself as the world's most advanced open source relational database.```
From https://www.bytebase.com/blog/postgres-vs-mysql/.

PR #461, opened 2024-09-24
Refactor database logic to be compatible with new drivers

maruiz93:

Related to #453

## Release Note

```release-note
NONE
```

## Notes for Reviewers

This is supposed to be a quick way to show the things I propose to change in https://github.com/kubearchive/kubearchive/pull/454.

That's the reason I create this PR as  a Draft


pkg/database/database.go
```diff
@@ -34,16 +34,18 @@ type DBInterface interface {
 }
 
 type Database struct {
-	db                *sql.DB
-	resourceTableName string
+	db   *sql.DB
+	info DatabaseInfo
```
beerparty, line 38: This change will require changes to the test code (pkg/database/database_test.go), which I do not see in this MR.

beerparty (changes requested):

pkg/database/database.go
ggallen: This is a draft, and essentially @maruiz93 created it for me to look at.

rh-hemartin:
Take a look at #464  as well

maruiz93:
> Take a look at #464 .

I have done some more changes inspired by your draft PR. I think it's a good idea to have each driver database self contained, meaning, when new driver is added it's nice to only create a new file with the needed implementations and a new line inside the `NewDatabase()` function.

> I think the databases need to know themselves, no need to have a companion "Info" struct as well.
I kept the info for now, I think it's a nice way to say "all the database implementations need to have this information" but while we can keep the "abstract class" that allow to share some database method implementations among the different drivers, I am not against flattening that information directly into the DBInterface

pkg/database/database_env.go
```diff
@@ -0,0 +1,55 @@
...
+
+func newDatabaseEnvironment() (*databaseEnvironment, error) {
+	var err error
+
+	kind := getEnv(DbKindEnvVar, err)
+	name := getEnv(DbNameEnvVar, err)
+	user := getEnv(DbUserEnvVar, err)
+	password := getEnv(DbPasswordEnvVar, err)
+	host := getEnv(DbHostEnvVar, err)
+	port := getEnv(DbPortEnvVar, err)
```
rh-hemartin, line 40: Reeturn the error instead:

```suggestion
	kind, err := getEnv(DbKindEnvVar, err)
	name, err := getEnv(DbNameEnvVar, err)
	user, err := getEnv(DbUserEnvVar, err)
	password, err := getEnv(DbPasswordEnvVar, err)
	host, err := getEnv(DbHostEnvVar, err)
	port, err := getEnv(DbPortEnvVar, err)
```

And inside the `getEnv` join if there is an error and return.

Because currently this does not work, and we shouldn't pass a pointer to an error. At least its now working for me on this example:

```
package main

import (
	"errors"
	"fmt"
)

func a(err error) {
	err = errors.Join(err, fmt.Errorf("Something bad happened"))
}

func main() {
	var err error
	a(err)

	if err != nil {
		fmt.Println("There was an error")
	}

}
```

maruiz93:
Added new changes for:
* Exposing only what is needed outside of the database package
* Extract connection data out of the Database object as it is only used when instantiating the database

pkg/database/database.go
```diff
@@ -7,22 +7,18 @@ import (
...
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
 
-const (
-	resourceTableName        = "resource"
-	resourcesQuery           = "SELECT data FROM %s WHERE kind=$1 AND api_version=$2"
-	namespacedResourcesQuery = "SELECT data FROM %s WHERE kind=$1 AND api_version=$2 AND namespace=$3"
-	writeResource            = `INSERT INTO %s (uuid, api_version, kind, name, namespace, resource_version, cluster_deleted_ts, data) Values ($1, $2, $3, $4, $5, $6, $7, $8) ON CONFLICT(uuid) DO UPDATE SET name=$4, namespace=$5, resource_version=$6, cluster_deleted_ts=$7, data=$8`
-)
+type queryData struct {
```
rh-hemartin, line 16: What about?

```suggestion
type DatabaseStatements struct {
```

pkg/database/database_env.go
maruiz93: ops, you are right, I missed this

pkg/database/database.go
maruiz93: I like the name better but why having it exposed?

rh-hemartin: I wrote it with capital letter out of habit. It does not need to be exposed.

On a second thought this is consumed outside this file, so it may well be "exposed" if the mysql and postgresql implementations were modules other than `package database`.

pkg/database/database_env.go
maruiz93: I fixed it by passing the pointer (my intention at the beginning but I missed it)

pkg/database/connection.go
```diff
@@ -1,58 +1,40 @@
...
-// Returns the value of the environment variable dbEnvVar. If dbEnvVar is not set, readDbEnvVar returns an error
-func readDbEnvVar(dbEnvVar EnvVar) (string, error) {
-	envVarVal, exists := os.LookupEnv(string(dbEnvVar))
-	if !exists {
-		return "", fmt.Errorf(dbConnectionErrStr, string(dbEnvVar))
-	}
-	return envVarVal, nil
+type connectionData struct {
+	driver           string
+	connectionString string
```
rh-hemartin, line 14: This struct feels a little bit useless. When we use it we are initializing it, we are passing these two parameters, just to call `establishConnection()` on it. What about having a  `establishConnection(driverName, connectionString) (*sql.DB, error)` function and call that instead?

pkg/database/database_env.go
maruiz93: it's less explicit, in general I wouldn't do it that way but it's a helper private method and it's point is to save some lines of code, and for this situation I think could be ok

pkg/database/database.go
maruiz93: what about `SQLStatements`? I updated to that, I think it's even more explicit.

As for the exposing the variable outside of the package, you are right, if we want to move the different driver db files outside we will need to expose them.

pkg/database/connection.go
maruiz93: yes that makes sense

pkg/database/database.go
rh-hemartin: Read nice to me. At the end of the day, they are called statements too.

pkg/database/connection.go
maruiz93: Done

pkg/database/database_env.go
rh-hemartin: Nevermind, I had something wrong.

pkg/database/database.go
beerparty: Sorry - I missed the Draft label.

ggallen:
Superseded by https://github.com/kubearchive/kubearchive/pull/469.

PR #464, opened 2024-09-25
Pr 454

rh-hemartin:

Resolves #

## Release Note

```release-note
NONE
```


maruiz93 (changes requested):
Is there any reason of explicitly reimplementing all the methods of the database in each "class" (i.e. MySQLDatabase and PostgreSQLDatabase structs)

In my approach https://github.com/kubearchive/kubearchive/pull/461 I tried to do the following:

image

While in this PR I see this:

image

Meaning, that the code in this PR it is not "encapsulating and sharing" the common methods across the different database implementation like the method "Ping", or the other querymethods that are likely to remain the same across different drivers.

NOTE that I used common Object Oriented Progamming language in my drawings although the concepts are slightly different in go (the inheritance of methods is done through composition, the abstract class CAN be instantiated, but it can't be considered a database implementation as it doesn't contain all the methods, classes are structs...) but the goal that I want to share keeps applying.

rh-hemartin:
> Is there any reason of explicitly reimplementing all the methods of the database in each "class" (i.e. MySQLDatabase and PostgreSQLDatabase structs)

I didn't want to nest structs, but nesting the structs is something I can live with.

rh-hemartin:
I'm closing this in favor of #461

PR #468, opened 2024-09-25
make sure existing filters update if CEL expressions change

### LINKED ISSUE #467: Sink doesn't react to filters updating
If you update the filters in a KubeArchiveConfig that already exists, the sink will not react to the update and will continue to use the old filters. The sink should be able to update the filters when the ConfigMap is updated to include the updated filters.

The work around for updating the KubeArchiveConfig is to delete the KubeArchiveConfig and create a new one with the desired filters

skoved:

Resolves #467 

## Release Note

```release-note
NONE
```

## Notes for Reviewers

Initially the sink was mounting the ConfigMap and using fsnotify to figure out when values in the ConfigMap change. However, the way that kubernetes creates files for the mounted ConfigMap in a container makes it so that fsnotify will not send events when data for a key that already exists in the ConfigMap is updated. Since the sink needs to be able to update existing filters, this PR has the sink perform a watch on the sink-filters ConfigMap. One side effect is the sink becomes aware of changes to filters with in a couple of seconds.


cmd/sink/filters/filters.go
```diff
@@ -332,8 +332,26 @@ func (f *Filters) handleFsEvent(logger *log.Logger, event fsnotify.Event) {
...
+		globalArchive, globalDelete, globalArchiveOnDelete, err := getGlobalCelExprs(filePaths)
+		if err != nil && !errors.Is(err, ErrNoGlobal) {
+			logger.Println("Could not read global filters:", err)
+			break
+		}
+		err = f.changeGlobalFilters(filePaths, globalArchive, globalDelete, globalArchiveOnDelete)
+		if err != nil {
+			logger.Println("Problem updating filters:", err)
+		}
+		logger.Println("Successfully updated filters")
```
ggallen, line 353: I haven't been able to figure it out exactly, but it seems like there's a lot of duplicated code in the `switch` statement.  Is there a way to simplify this?  It seems like `chmod` is similar to `write`.

This section from https://github.com/fsnotify/fsnotify concerns me:
```
Why do I get many Chmod events?

Some programs may generate a lot of attribute changes; for example
Spotlight on macOS, anti-virus programs, backup applications, and
some others are known to do this. As a rule, it's typically best to
ignore Chmod events. They're often not useful, and tend to cause problems.
```
Why do we have to use `chmod` here? Is it because these are all symlinks?

skoved: I'm no longer using `fsnotify`, because it does not work with the two layers of symlinks that kubernetes creates when mounting a `ConfigMap`. I think the new logic is much simpler

rh-hemartin (approved):

cmd/sink/filters/filters.go
```diff
@@ -9,92 +9,62 @@ import (
 	"fmt"
 	"log/slog"
 	"maps"
-	"os"
-	"strings"
 	"sync"
 
-	"github.com/fsnotify/fsnotify"
 	"github.com/google/cel-go/cel"
 	kubearchiveapi "github.com/kubearchive/kubearchive/cmd/operator/api/v1alpha1"
```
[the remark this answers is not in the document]
rh-hemartin: While your suggestion makes sense to me, if we want to do it I would do it another PR. Making it here will make this more difficult, as the changes to move that to another package are not simple.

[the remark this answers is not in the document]
maruiz93: That's ok by me. If @skoved agrees we can create a separate issue to address it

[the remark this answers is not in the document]
skoved: I agree with @rh-hemartin. I'd prefer to create a separate issue for this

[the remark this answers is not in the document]
skoved: created #532

maruiz93 (approved):

rh-hemartin (changes requested):
Found a bug

charts/kubearchive/templates/sink/role.yaml
```diff
@@ -0,0 +1,14 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: {{ tpl .Values.sink.name . }}
```
rh-hemartin, line 5: The bug found can be solved by suffixing this with `-configmaps` so the operator does not modify it.

charts/kubearchive/templates/sink/role_binding.yaml
```diff
@@ -0,0 +1,13 @@
...
+metadata:
+  name: {{ tpl .Values.sink.name . }}
+subjects:
+  - kind: ServiceAccount
+    name: {{ tpl .Values.sink.name . }}
+    namespace: {{ .Release.Namespace }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: {{ tpl .Values.sink.name . }}
```
rh-hemartin, line 13: This will need to change as well.

skoved:
> I got an error related to the interaction between the operator and the kubearchive sink role. The operator modifies the sink role and removes the permissions to read the configmap, so after applying a KubeArchiveConfig and doing a rollout of the sink, the sink loses the ability to read the configmap. It does not happen until you rollout, or until the watcher expires and needs to restart watching.
> 
> To reproduce, run:
> 
> ```
> # Fine
> bash hack/quick-install.sh
> kubectl logs -n kubearchive deployment/kubearchive-sink
> kubectl get -n kubearchive role kubearchive-sink
> 
> # Fine
> bash hack/kubearchive-install.sh
> kubectl logs -n kubearchive deployment/kubearchive-sink
> kubectl get -n kubearchive role kubearchive-sink
> 
> # Fine
> # Create a KubeArchiveConfig on the "kubearchive" namespace
> 
> # Goes wrong
> kubectl rollout -n kubearchive restart deployment/kubearchive-sink
> kubectl logs -n kubearchive deployment/kubearchive-sink | grep "Watch failed"
> 
> # The operator modifies the role and removes the configmap permissions
> kubectl get -n kubearchive role kubearchive-sink
> ```

I just tried that out and you're right! I'm making the changes that you suggested now, but I'm confused why the operator does that

charts/kubearchive/templates/sink/role.yaml
skoved: i went with `-watch` instead because I think it better describes what role is for

rh-hemartin (approved):

PR #469, opened 2024-09-25
Add support for database integrations.

### LINKED ISSUE #453: Add MySQL support.

ggallen:

Resolves #453

## Release Note

```release-note
NONE
```

## Notes for Reviewers

This supersedes the following PRs:

- https://github.com/kubearchive/kubearchive/pull/454
- https://github.com/kubearchive/kubearchive/pull/461
- https://github.com/kubearchive/kubearchive/pull/464


rh-hemartin (approved):

pkg/database/database.go
```diff
@@ -31,19 +44,37 @@ type DBInterface interface {
...
+
+	err = database.TestConnection(env)
 	if err != nil {
 		return nil, err
 	}
-	var db *sql.DB
+	return database, nil
+}
+
+func (db *Database) TestConnection(env map[string]string) error {
```
maruiz93, line 75: I think that this method name is missleading, it states `TestConnection` but if I understand correctly this is doing more than that (the method that just test the connection is the `Ping()`), this is establishing the database connection inside the `Database` object. So here are several things to consider:
* This method shouldn't be called from the outside (api/sink), it's just meant to be used from the database constructor so I believe it shouldn't be part of the `DBInterface`. We are falling into the [interface pollution](https://100go.co/5-interface-pollution/) practice.
* When the `NewDatabase` actually creates the database object by running the proper function of the `RegisteredDatabases` map it returns and non-functional database because the db connection hasn't been established. I would suggest to establish the connection in the `New<db>Database` instead of here.
* As  setting the connection is common code to be used, I think it makes sense to extract it to another file and provide it as a "tool" for the different database implementations to use. it would just have to receive the driver and the connection string and return the connection object `sql.DB`. Please take a look at the `connection.go` file in https://github.com/kubearchive/kubearchive/pull/461 where I do that

pkg/database/postgresql_database.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+type PostgreSQLDatabase struct {
+	*Database
+}
+
+func NewPostgreSQLDatabase() DBInterface {
+	var db *sql.DB
+	return PostgreSQLDatabase{&Database{db, PostgreSQLDatabaseInfo{&DatabaseInfo{}}}}
+}
```
maruiz93, line 51: Following my previous comment, this function is returning a nonfunctional object, and I think we shouldn't have a constructor that returns something that doesn't work.
I would rather put here the actual initialization of the database connection and the database info:
```suggestion
func NewPostgreSQLDatabase() DBInterface {
	dbInfo := DatabaseInfo{} 
	dbInfo.SetEnvironment() 
	db := establishConnection(dbInfo.GetDriver(), dbInfo.GetConnectionString()) // you don't have this function yet, but I'm assuming we can have one if we extract this part to a function in another place
	return PostgreSQLDatabase{&Database{db, PostgreSQLDatabaseInfo{&dbInfo}}}
}
```

pkg/database/database_env.go
```diff
@@ -0,0 +1,42 @@
...
+	DbUserEnvVar     string = "DATABASE_USER"
+	DbPasswordEnvVar string = "DATABASE_PASSWORD" // nosec G101 not a password
+	DbHostEnvVar     string = "DATABASE_URL"
+	DbPortEnvVar     string = "DATABASE_PORT"
+)
+
+var DbEnvVars = [...]string{DbKindEnvVar, DbNameEnvVar, DbUserEnvVar, DbPasswordEnvVar, DbHostEnvVar, DbPortEnvVar}
+
+// Reads database connection info from the environment variables and returns a map of variable name to value.
+func NewDatabaseEnvironment() (map[string]string, error) {
```
maruiz93, line 26: Do we need this method to be exposed? Is this meant to be used outside of the `database` package?

maruiz93 (changes requested):
I wasn't aware of the posibilities of the [go init function]() and I think it's good idea to use it :)

I've added a couple of suggestions regarding where is the database initialization taking place.

skoved (changes requested):

pkg/database/postgresql_database.go
```diff
@@ -0,0 +1,51 @@
...
+
+	_ "github.com/lib/pq"
+)
+
+func init() {
+	RegisteredDatabases["postgresql"] = NewPostgreSQLDatabase
+}
+
+type PostgreSQLDatabaseInfo struct {
+	*DatabaseInfo
```
maruiz93, line 17: Now that you are passing the environment to the constructor, I don't think that there is a need for having `DatabaseInfo` inside `PostgreSQLDatabaseInfo` I think that we can just have:
```suggestion
	env map[string]string
```

pkg/database/database.go
```diff
@@ -7,22 +7,26 @@ import (
...
+	GetDriverName() string
+	GetConnectionString() string
+	GetResourcesSQL() string
+	GetNamespacedResourcesSQL() string
+	GetWriteResourceSQL() string
+}
+
+type DatabaseInfo struct {
+	env map[string]string
+}
```
maruiz93, line 29: I believe there is no need for this struct anymore
```suggestion
```

maruiz93 (changes requested):

pkg/database/database.go
ggallen: Fixed.

pkg/database/database_env.go
```diff
@@ -0,0 +1,42 @@
...
+			env[name] = value
+		} else {
+			err = errors.Join(err, fmt.Errorf(dbConnectionErrStr, name))
+		}
+	}
+	if err == nil {
+		return env, nil
+	} else {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
ggallen: No, because `env` can be non-nill when `err` is non-nil.  When an error is returned the we should return nil for `err`.

pkg/database/postgresql_database.go
ggallen: But then new implementors will always have to add that `env`.  It seems cleaner to me to have the struct there and hide the implementation from the user.  No?

ggallen: And if `DatabaseInfo` were to change, it would require changes to all the implementers as well.

pkg/database/database_env.go
```diff
@@ -0,0 +1,42 @@
+// Copyright KubeArchive Authors
```
maruiz93, line 1: How about renaming this file to just environment? As we are already inside the `database` package I think there is no need to reiterate that

pkg/database/postgresql_database.go
```diff
@@ -0,0 +1,51 @@
+// Copyright KubeArchive Authors
```
maruiz93, line 1: How about renaming this file to postgresql? As we already are inside the `database` package I think there is no need to reiterate that.

pkg/database/postgresql_database.go
maruiz93: > But then new implementors will always have to add that env

And now they have to add `DatabaseInfo` how is one thing different to another?
Also, if at some point someone wouldn't need the environment variables there is no need to add it.

> And if DatabaseInfo were to change, it would require changes to all the implementers as well.

I am not sure I'm getting this, if I am suggesting removing this object, how is this something that should be maintained?

maruiz93 (changes requested):

pkg/database/postgresql_database.go
ggallen: Never mind. I've fixed this.

pkg/database/database_env.go
ggallen: Done.

pkg/database/postgresql_database.go
ggallen: Done.

pkg/database/database.go
ggallen: Fixed.

maruiz93 (approved):

skoved (approved):

PR #484, opened 2024-09-27
Add pagination support

### LINKED ISSUE #226: Add pagination to the KubeArchive API Server responses
It could be a large amount of resources stored in KubeArchive database.

We should include a default limit of resources to response by default and a way to paginate the full response.

K8s API uses the parameters `limit` and `continue` for that. The goal is to mimic that functionality.

More information in https://github.com/kubernetes/kubernetes/issues/49338

rh-hemartin:

Resolves #226 

## Release Note

```release-note
Endpoints that retrieve collections of resources support pagination similar to the Kubernetes API.
```

## Notes for Reviewers

* I decided to go with ORDER BY because I keep reading (with proof) that LIMIT + OFFSET is slow and its best to do this approach (known as keyset pagination).
* We talked about adding `remainingItems` in the List response. That will be handled in a separate issue (https://github.com/kubearchive/kubearchive/issues/531).

To test:

```
# Normal flow
kind create cluster
bash hack/quick-install.sh

# Load data
cat table.sql | kubectl exec -n postgresql -it pods/kubearchive-1 -- psql kubearchive

# Add GET jobs permissions to the default:default ServiceAccount
kubectl create role view-jobs --verb=get --resource=jobs,jobs/status --dry-run=client -o yaml | kubectl apply -f -
kubectl create rolebinding view-jobs --role view-jobs --user system:serviceaccount:default:default --dry-run=client -o yaml | kubectl apply -f -

# Normal flow
kubectl get -n kubearchive secrets kubearchive-api-server-tls -o jsonpath='{.data.ca\.crt}' | base64 -d > ca.crt
kubectl port-forward -n kubearchive svc/kubearchive-api-server 8081:8081 &

```
```
echo "TEST CASE: Curl (should output 1000)"
curl -s --cacert ca.crt -H "Authorization: Bearer $(kubectl create token default)" "https://localhost:8081/apis/batch/v1/namespaces/default/jobs" | jq '.items | length'

echo "TEST CASE: Curl with limits (should output 10)"
curl -s --cacert ca.crt -H "Authorization: Bearer $(kubectl create token default)" "https://localhost:8081/apis/batch/v1/namespaces/default/jobs?limit=10" | jq '.items'
curl -s --cacert ca.crt -H "Authorization: Bearer $(kubectl create token default)" "https://localhost:8081/apis/batch/v1/namespaces/default/jobs?limit=10" | jq '.items[].metadata.name'

echo "TEST CASE: Curl with limits, should have a continue token"
export CONTINUE_TOKEN=$(curl -s --cacert ca.crt -H "Authorization: Bearer $(kubectl create token default)" "https://localhost:8081/apis/batch/v1/namespaces/default/jobs?limit=10" | jq -r '.metadata.continue')
echo $CONTINUE_TOKEN

echo "TEST CASE: Curl with continue"
curl -s --cacert ca.crt -H "Authorization: Bearer $(kubectl create token default)" "https://localhost:8081/apis/batch/v1/namespaces/default/jobs?limit=10&continue=$CONTINUE_TOKEN" | jq '.items | length'
curl -s --cacert ca.crt -H "Authorization: Bearer $(kubectl create token default)" "https://localhost:8081/apis/batch/v1/namespaces/default/jobs?limit=10&continue=$CONTINUE_TOKEN" | jq '.items[].metadata.name'

echo "TEST CASE: This should fail because the token does not follow a date format"
curl -s --cacert ca.crt -H "Authorization: Bearer $(kubectl create token default)" "https://localhost:8081/apis/batch/v1/namespaces/default/jobs?limit=10&continue=$(echo "abc"|base64)"

echo "TEST CASE: This should fail because the limit is not correct"
curl -s --cacert ca.crt -H "Authorization: Bearer $(kubectl create token default)" "https://localhost:8081/apis/batch/v1/namespaces/default/jobs?limit=ab&continue=$(echo "abc"|base64)"

echo "TEST CASE: This should return 1000 because `continue` on itself is ignored"
curl -s --cacert ca.crt -H "Authorization: Bearer $(kubectl create token default)" "https://localhost:8081/apis/batch/v1/namespaces/default/jobs?continue=$CONTINUE_TOKEN" | jq '.items | length'

echo "TEST CASE: The metadata shouldn't contain 'continue' if no limit was specified"
curl -s --cacert ca.crt -H "Authorization: Bearer $(kubectl create token default)" "https://localhost:8081/apis/batch/v1/namespaces/default/jobs" | jq '.metadata'
```

Download [table.tar.gz](https://github.com/user-attachments/files/17165553/table.tar.gz)  and extract with `tar -xzf table.tar.gz` to get the `table.sql` used in the previous instructions.

### DIFF
--- a/cmd/api/pagination/pagination.go
+++ b/cmd/api/pagination/pagination.go
@@ -1,0 +1,117 @@
+// Copyright KubeArchive Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package pagination
+
+import (
+	"encoding/base64"
+	"fmt"
+	"log"
+	"net/http"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/gin-gonic/gin"
+	"github.com/google/uuid"
+	"github.com/kubearchive/kubearchive/cmd/api/abort"
+)
+
+const (
+	limitKey        = "limit"
+	continueKey     = "continue"
+	continueDateKey = "continueDate"
+	continueUUIDKey = "continueUUID"
+	defaultLimit    = "100"
+	maxAllowedLimit = 1000
+)
+
+// GetValuesFromContext is a helper function for routes to retrieve the
+// information needed. This is kept here so its close to the function
+// that sets these values in the context (Middleware)
+func GetValuesFromContext(context *gin.Context) (string, string, string) {
+	return context.GetString(limitKey), context.GetString(continueUUIDKey), context.GetString(continueDateKey)
+}
+
+func CreateToken(uuid, date string) string {
+	// The date is returned as a quoted string, so remove the quotes
+	date = strings.TrimPrefix(date, "\"")
+	date = strings.TrimSuffix(date, "\"")
+	tokenString := fmt.Sprintf("%s %s", uuid, date)
+	return base64.StdEncoding.EncodeToString([]byte(tokenString))
+}
+
+// Middleware validates the `limit` and `continue` query parameters
+// and populates `limit` and `continueValue` in the context with their
+// respective values so they are retrieved by the endpoints that need it
+func Middleware() gin.HandlerFunc {
+	return func(context *gin.Context) {
+		limitString := context.Query(limitKey)
+		continueToken := context.Query(continueKey)
+
+		var limit string
+		if limitString == "" {
+			// If not specified, we set a limit so users don't retrieve
+			// large collections by mistake. Those collections could
+			// make us crash if they are too large
+			limit = defaultLimit
+		} else {
+			var err error
+			limitInteger, err := strconv.Atoi(limitString)
+			if err != nil {
+				abort.Abort(context, fmt.Sprintf("limit '%s' could not be converted to integer", limitString), http.StatusBadRequest)
+				return
+			}
+
+			if limitInteger > maxAllowedLimit {
+				abort.Abort(context, fmt.Sprintf("limit '%s' exceeds the maximum allowed '%d'", limitString, maxAllowedLimit), http.StatusBadRequest)
+				return
+			}
+
+			// We reserialize to avoid SQL injection. There is the possibility the
+			// value is a valid integer, but in SQL does something else.
+			limit = strconv.Itoa(limitInteger)
+		}
+
+		var continueDate string
+		var continueUUID string
+		if continueToken != "" {
+			continueBytes, err := base64.StdEncoding.DecodeString(continueToken)
+			if err != nil {
+				abort.Abort(context, "could not decode the continuation token", http.StatusBadRequest)
+				return
+			}
+
+			continueParts := strings.Split(string(continueBytes), " ")
+			if len(continueParts) != 2 {
+				abort.Abort(context, "expected two elements on the continue token, received a different amount", http.StatusBadRequest)
+				return
+			}
+
+			continueUUID = continueParts[0]
+			err = uuid.Validate(continueUUID)
+			if err != nil {
+				abort.Abort(context, "first element of the continue token is not a valid UUID", http.StatusBadRequest)
+				return
+			}
+
+			continueDate = continueParts[1]
+			continueTimestamp, err := time.Parse(time.RFC3339, continueDate)
+			if err != nil {
+				log.Printf("Error: %s", err)
+				abort.Abort(context, fmt.Sprintf("continuation token decoded value '%s' is not valid. It should match '%s'", continueDate, time.RFC3339), http.StatusBadRequest)
+				return
+			}
+
+			// We reserialize to avoid SQL injection. There is the possibility the
+			// value is a valid date, but in SQL does something else.
+			continueDate = continueTimestamp.Format(time.RFC3339)
+		}
+
+		// Pass the values using the context, these should be retrieved
+		// using `GetValuesFromContext`
+		context.Set(limitKey, limit)
+		context.Set(continueDateKey, continueDate)
+		context.Set(continueUUIDKey, continueUUID)
+	}
+}
--- a/pkg/database/database.go
+++ b/pkg/database/database.go
@@ -19,19 +19,29 @@ var RegisteredDatabases = make(map[string]newDatabaseFunc)
 type DBInfoInterface interface {
 	GetDriverName() string
 	GetConnectionString() string
+
 	GetResourcesSQL() string
+	GetResourcesLimitedSQL() string
+	GetResourcesLimitedContinueSQL() string
+
 	GetNamespacedResourcesSQL() string
+	GetNamespacedResourcesLimitedSQL() string
+	GetNamespacedResourcesLimitedContinueSQL() string
+
 	GetNamespacedResourceByNameSQL() string
+
 	GetWriteResourceSQL() string
 }
 
 type DBInterface interface {
-	QueryResources(ctx context.Context, kind, group, version string) ([]*unstructured.Unstructured, error)
-	QueryCoreResources(ctx context.Context, kind, version string) ([]*unstructured.Unstructured, error)
-	QueryNamespacedResources(ctx context.Context, kind, group, version, namespace string) ([]*unstructured.Unstructured, error)
-	QueryNamespacedResourceByName(ctx context.Context, kind, group, version, namespace, name string) (*unstructured.Unstructured, error)
-	QueryNamespacedCoreResources(ctx context.Context, kind, version, namespace string) ([]*unstructured.Unstructured, error)
+	QueryCoreResources(ctx context.Context, kind, version, limit, continueUUID, continueDate string) ([]*unstructured.Unstructured, string, string, error)
+	QueryNamespacedCoreResources(ctx context.Context, kind, version, namespace, limit, continueUUID, continueDate string) ([]*unstructured.Unstructured, string, string, error)
 	QueryNamespacedCoreResourceByName(ctx context.Context, kind, version, namespace, name string) (*unstructured.Unstructured, error)
+
+	QueryResources(ctx context.Context, kind, group, version, limit, continueUUID, continueDate string) ([]*unstructured.Unstructured, string, string, error)
+	QueryNamespacedResources(ctx context.Context, kind, group, version, namespace, limit, continueUUID, continueDate string) ([]*unstructured.Unstructured, string, string, error)
+	QueryNamespacedResourceByName(ctx context.Context, kind, group, version, namespace, name string) (*unstructured.Unstructured, error)
+
 	WriteResource(ctx context.Context, k8sObj *unstructured.Unstructured, data []byte) error
 	Ping(ctx context.Context) error
 	CloseDB() error
@@ -61,23 +71,67 @@ func (db *Database) Ping(ctx context.Context) error {
 	return db.db.PingContext(ctx)
 }
 
-func (db *Database) QueryResources(ctx context.Context, kind, group, version string) ([]*unstructured.Unstructured, error) {
+func (db *Database) QueryResources(ctx context.Context, kind, group, version, limit, continueUUID, continueDate string) ([]*unstructured.Unstructured, string, string, error) {
 	apiVersion := fmt.Sprintf("%s/%s", group, version)
-	return db.performResourceQuery(ctx, db.info.GetResourcesSQL(), kind, apiVersion)
+
+	args := []string{kind, apiVersion}
+	query := db.info.GetResourcesSQL()
+	if limit != "" {
+		if continueUUID == "" || continueDate == "" {
+			query = db.info.GetResourcesLimitedSQL()
+			args = append(args, limit)
+		} else {
+			query = db.info.GetResourcesLimitedContinueSQL()
+			args = append(args, continueDate, continueUUID, limit)
+		}
+	}
+
+	return db.performResourceQuery(ctx, query, args...)
 }
 
-func (db *Database) QueryCoreResources(ctx context.Context, kind, version string) ([]*unstructured.Unstructured, error) {
-	return db.performResourceQuery(ctx, db.info.GetResourcesSQL(), kind, version)
+func (db *Database) QueryCoreResources(ctx context.Context, kind, version, limit, continueUUID, continueDate string) ([]*unstructured.Unstructured, string, string, error) {
+	args := []string{kind, version}
+	query := db.info.GetResourcesSQL()
+
+	if limit != "" {
+		if continueUUID == "" || continueDate == "" {
+			query = db.info.GetResourcesLimitedSQL()
+			args = append(args, limit)
+		} else {
+			query = db.info.GetResourcesLimitedContinueSQL()
+			args = append(args, continueDate, continueUUID, limit)
+		}
+	}
+
+	return db.performResourceQuery(ctx, query, args...)
 }
 
-func (db *Database) QueryNamespacedResources(ctx context.Context, kind, group, version, namespace string) ([]*unstructured.Unstructured, error) {
+func (db *Database) QueryNamespacedResources(ctx context.Context, kind, group, version, namespace, limit, continueUUID, continueDate string) ([]*unstructured.Unstructured, string, string, error) {
 	apiVersion := fmt.Sprintf("%s/%s", group, version)
-	return db.performResourceQuery(ctx, db.info.GetNamespacedResourcesSQL(), kind, apiVersion, namespace)
+
+	// Build query and args depending on `limit` and `continueValue`
+	args := []string{kind, apiVersion, namespace}
+	query := db.info.GetNamespacedResourcesSQL()
+	if limit != "" {
+		if continueUUID == "" || continueDate == "" {
+			query = db.info.GetNamespacedResourcesLimitedSQL()
+			args = append(args, limit)
+		} else {
+			query = db.info.GetNamespacedResourcesLimitedContinueSQL()
+			args = append(args, continueDate, continueUUID, limit)
+		}
+	}
+
+	return db.performResourceQuery(ctx, query, args...)
 }
 
 func (db *Database) QueryNamespacedResourceByName(ctx context.Context, kind, group, version, namespace, name string) (*unstructured.Unstructured, error) {
 	apiVersion := fmt.Sprintf("%s/%s", group, version)
-	resources, err := db.performResourceQuery(ctx, db.info.GetNamespacedResourceByNameSQL(), kind, apiVersion, namespace, name)
+	resources, _, _, err := db.performResourceQuery(ctx, db.info.GetNamespacedResourceByNameSQL(), kind, apiVersion, namespace, name)
+	if err != nil {
+		return nil, err
+	}
+
 	if len(resources) == 0 {
 		return nil, err
 	} else if len(resources) == 1 {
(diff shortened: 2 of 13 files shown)


pkg/database/postgresql.go
```diff
@@ -40,6 +40,14 @@ func (info PostgreSQLDatabaseInfo) GetWriteResourceSQL() string {
 		"ON CONFLICT(uuid) DO UPDATE SET name=$4, namespace=$5, resource_version=$6, cluster_deleted_ts=$7, data=$8"
 }
 
+func (info PostgreSQLDatabaseInfo) GetNamespacedResourcesLimitedSQL() string {
+	return "SELECT created_at, data FROM resource WHERE kind=$1 AND api_version=$2 AND namespace=$3 ORDER BY created_at LIMIT $4"
+}
+
+func (info PostgreSQLDatabaseInfo) GetNamespacedResourcesLimitedContinueSQL() string {
+	return "SELECT created_at, data FROM resource WHERE kind=$1 AND api_version=$2 AND namespace=$3 AND created_at > $4 ORDER BY created_at LIMIT $5"
+}
```
maruiz93, line 49: I don't know how likely it is that two resources share the order key (`created_at` in this case) but in case this could happen, couldn't we miss some resources in the pagination?

Let's say we have resources 1, 2, 3 and 4 created at ts A, B, B and C respectively and we are paginating the content in pieces of size 2.
The first page will have elements: 1 and 2, but the second page could have just the element 4 because we are querying that `created_at`>B
  
If we want to go with this route my proposal would be to do the `ORDER BY` by combining `created_at` with another field that makes the resource unique (e.g., the `uuid`)

```suggestion
func (info PostgreSQLDatabaseInfo) GetNamespacedResourcesLimitedContinueSQL() string {
	return "SELECT created_at, data FROM resource WHERE kind=$1 AND api_version=$2 AND namespace=$3 AND created_at > $4 AND uuid > $5 ORDER BY created_at, uuid LIMIT $6"
}
```

rh-hemartin: Yes, that could be possible if both have the same nanosecond of insertion. I will prepare 4 rows with the scenario you exposed here and check.

rh-hemartin:
> The total number of elements
    The "range" or "page" that is being provided

I will provide the `remainingItems` as the Kubernetes API implementation does, it will introduce another query but I do think it makes sense.
    
> It also implies being able to go backwards and forward:
    I want to go to the previous page
    I want to go to the last page

I think it should be enough if you as a client store the tokens we are sending. I will create a case and add it to the PR bash script.

> Do we want to use that instead of the actual timestamp of creation of the resource in the cluster (the one that Kubernetes API uses?

As we discussed in the meeting I will use the timestamp of creation of the resource in the cluster.

> This kind of things are also important if we want the KubeArchive CLI (or other UI) to be able to "page" merged resources that come from kubernetes API and resources that come from kubeArchive API.

I don't think the CLI will need to page merged resources. Paging implies state, and CLI tools are normally stateless. In case you want to paginate you would pipe into `less` and scroll, or pipe into a file and open it.

pkg/database/postgresql.go
rh-hemartin: I fixed the scenario and created a little experiment:

```
## Example where two share `created_date` to ensure its working properly
kubectl exec -n postgresql -it pods/kubearchive-1 -- psql kubearchive -c "DELETE FROM resource;"
cat experiment.txt | kubectl exec -n postgresql -i pods/kubearchive-1 -- psql kubearchive
curl -s --cacert ca.crt -H "Authorization: Bearer $(kubectl create token default)" "https://localhost:8081/apis/batch/v1/namespaces/default/jobs" | jq '.items[].metadata.name'
curl -s --cacert ca.crt -H "Authorization: Bearer $(kubectl create token default)" "https://localhost:8081/apis/batch/v1/namespaces/default/jobs?limit=2" | jq '.items[].metadata.name'
export CONTINUE_TOKEN=$(curl -s --cacert ca.crt -H "Authorization: Bearer $(kubectl create token default)" "https://localhost:8081/apis/batch/v1/namespaces/default/jobs?limit=2" | jq -r '.metadata.continue')
curl -s --cacert ca.crt -H "Authorization: Bearer $(kubectl create token default)" "https://localhost:8081/apis/batch/v1/namespaces/default/jobs?limit=2&continue=$CONTINUE_TOKEN" | jq '.items[].metadata.name'
```

[experiment.txt](https://github.com/user-attachments/files/17258455/experiment.txt)

rh-hemartin:
I started doing tests. I'm having some problems with mocking the database. The error I get with the latest pushed code is:

```
=== RUN   TestQueryResources/mariadb/Results_query
    database_test.go:91: 
        	Error Trace:	/home/hemartin/git/kubearchive/pkg/database/database_test.go:91
        	Error:      	Received unexpected error:
        	            	invalid character 'c' after top-level value
        	Test:       	TestQueryResources/mariadb/Results_query
    database_test.go:95: 
        	Error Trace:	/home/hemartin/git/kubearchive/pkg/database/database_test.go:95
        	Error:      	Expected value not to be nil.
        	Test:       	TestQueryResources/mariadb/Results_query
    database_test.go:97: 
        	Error Trace:	/home/hemartin/git/kubearchive/pkg/database/database_test.go:97
        	Error:      	Not equal: 
        	            	expected: 1
        	            	actual  : 0
        	Test:       	TestQueryResources/mariadb/Results_query
```

I think the problem is related with the creation of the row for mocking, but I don't know how to do it so it gets the proper values from the JSON. I tried to fake the first element which should be a timestamp, and the last that should be a UUID.

Any ideas?

pkg/database/postgresql.go
maruiz93: Nice! this works as expected :)

rh-hemartin:
I will start working on adding the `remainingItems` count to the answer.

rh-hemartin:
> I will start working on adding the `remainingItems` count to the answer.

I have some progress but it will take a while, I will create a issue so we can implement it later. I will leave details there.

pkg/database/database.go
```diff
@@ -87,12 +141,29 @@ func (db *Database) QueryNamespacedResourceByName(ctx context.Context, kind, gro
...
+	if limit != "" {
+		if continueUUID == "" || continueDate == "" {
+			query = db.info.GetNamespacedResourcesLimitedSQL()
+			args = append(args, limit)
+		} else {
+			query = db.info.GetNamespacedResourcesLimitedContinueSQL()
+			args = append(args, continueDate, continueUUID, limit)
+		}
+	}
+
```
maruiz93, line 157: this block is repeated 4 times, could you consider extracting it to a private function?

maruiz93 (changes requested):
I left a couple of comments mostly regarding style and readability

cmd/api/pagination/pagination.go
```diff
@@ -0,0 +1,104 @@
...
+	limitKey        = "limit"
+	continueKey     = "continue"
+	continueDateKey = "continueDate"
+	continueUUIDKey = "continueUUID"
+)
+
+// Helper function for routes to retrieve the information needed
+// This is kept here so its close to the function that sets these
+// values in the context (Middleware)
+func GetValuesFromContext(context *gin.Context) (string, string, string) {
```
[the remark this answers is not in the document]
rh-hemartin: Changed

cmd/api/pagination/pagination.go
```diff
@@ -0,0 +1,104 @@
...
+	return context.GetString(limitKey), context.GetString(continueUUIDKey), context.GetString(continueDateKey)
+}
+
+func CreateToken(uuid, date string) string {
+	// The date is returned as a quoted string, so remove them quotes
```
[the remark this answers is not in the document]
rh-hemartin: Chaned

cmd/api/pagination/pagination.go
```diff
@@ -0,0 +1,104 @@
...
+	date = strings.TrimPrefix(date, "\"")
+	date = strings.TrimSuffix(date, "\"")
+	tokenString := fmt.Sprintf("%s %s", uuid, date)
+	return base64.StdEncoding.EncodeToString([]byte(tokenString))
+}
+
+// This middleware validates the `limit` and `continue` query parameters
```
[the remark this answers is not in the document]
rh-hemartin: Changed

cmd/api/pagination/pagination.go
```diff
@@ -0,0 +1,104 @@
...
+	return func(context *gin.Context) {
+		limitString := context.Query(limitKey)
+		continueToken := context.Query(continueKey)
+
+		var limit string
+		if limitString != "" {
+			var err error
+			limitInteger, err := strconv.Atoi(limitString)
+			if err != nil {
+				abort.Abort(context, fmt.Sprintf("Limit '%s' could not be converted to integer.", limitString), http.StatusBadRequest)
```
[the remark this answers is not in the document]
rh-hemartin: As far as I know this is not an error this string gets returned in the API. However it does represent an error.  I changed it, but I'm not sure what are you referring to.

cmd/api/pagination/pagination.go
```diff
@@ -0,0 +1,104 @@
...
+				abort.Abort(context, fmt.Sprintf("Continuation token decoded value '%s' is not valid. It should match '%s'.", continueDate, time.RFC3339), http.StatusBadRequest)
+				return
+			}
+
+			// We reserialize to avoid SQL injection. There is the possibility the
+			// value is a valid date, but in SQL does something else.
+			continueDate = continueTimestamp.Format(time.RFC3339)
+		}
+
+		// Pass the values using the context, even if not recommended
```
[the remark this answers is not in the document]
rh-hemartin: I also have a blog post that says that is a bad idea: https://www.calhoun.io/pitfalls-of-context-values-and-how-to-avoid-or-mitigate-them/

My feeling is that this values are magic. You set them from a place and them retrieve them in another place. Its kind of a global value, even if scoped to the request itself. This is why I added the function that retrieves them at the top of this file: it can avoid drifting, both use the same set of constants so for a change the chances of mismatching context value names is lower, etc.

I will remove the "not recommended" as there are multiple points of view.

integrations/database/mariadb/install.sh
```diff
@@ -13,7 +13,7 @@ NAMESPACE="mariadb"
 kubectl create ns mariadb-operator --dry-run=client -o yaml | kubectl apply -f -
 helm repo add mariadb-operator https://helm.mariadb.com/mariadb-operator
 helm repo update mariadb-operator
-helm upgrade --install --namespace mariadb-operator mariadb-operator-crds mariadb-operator/mariadb-operator-crds
```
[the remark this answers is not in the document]
rh-hemartin: I had to patch it for testing, its out of sync with main. I'll fix it.

pkg/database/database.go
rh-hemartin: The block is repeated twice: one for the pair `QueryResources` and `QueryCoreResources`, and a second time for the pair `QueryNamespacedResources` and `QueryNamespacedCoreResources`.

So I think the repetition is on the functions instead. Instead of extracting this block out, I prefer  we address the duplication of the whole function (`QueryResources` and `QueryCoreResources` are the same except for `apiVersion := fmt.Sprintf("%s/%s", group, version)`, and the other pair its the same) in another issue.

rh-hemartin: In other words, `QueryResources` is just `QueryCoreResources` with `version = apiVersion`. The same for `QueryNamespacedResources` and `QueryNamespacedCoreResources`, etc.

cmd/api/pagination/pagination.go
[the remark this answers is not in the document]
maruiz93: I believe both blog posts say the same (I wasn't against the idea of "using context isn't recommended", just wanted to make sure that our reasons to backup that were the same)

My post:
> With context.Context, function signature becomes simpler and consistent. Not only is it easier to read, we can also write middlewares to add more functionalities
> However, such convenience doesn't come with zero downside. Firstly, the dependency and input of a function become hidden
> Another compromise when using Go context is the missing of type information.

Your post:
> Downsides to using context values
> Required data for functions is obscured
> We lose type safety at compile time

So we were exactly on the same page.

However, I believe it's good to remove the comment because we are using gin as our API framework and with gin middleware the only way to share information with the routes is to do it through the context.
Let me share here [another blog post](https://app.studyraid.com/en/read/5926/130187/gin-context#best-practices-for-using-gin-context) that gathers some best practices of using gin context, among them:

> Use context values judiciously: While it's convenient to store data in the context, avoid overusing this feature. Reserve it for request-scoped data that needs to be passed between middleware and handlers.

And this is our case, AFAICS, all the variables that we are storing in the context are request specific, in this, and in the other middleware functions that we have, so I believe we are good to go with it.

pkg/database/database.go
maruiz93: Good point. I agree that the change we should do (at least first) is to refactor the methods so we "merge" `QueryResources` with `QueryCoreResources` and `QueryNamespacedResources` with `QueryNamespacedCoreResources`.

We can create a separate issue that

cmd/api/pagination/pagination.go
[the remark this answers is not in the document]
maruiz93: You are right, this is currently being returned as is, and we are not sending it as an error, but I think that we should.

I opened a new issue to address this topic properly: https://github.com/kubearchive/kubearchive/issues/536

Thank you for considering my suggestion by the way :)

skoved (approved):
I don't know how I feel about having to check if `limit`, `continueUUID`, and `continueDate` are empty string, but I don't have a better alternative to recommend right now

ggallen:
> I don't know how I feel about having to check if `limit`, `continueUUID`, and `continueDate` are empty string, but I don't have a better alternative to recommend right now

I'm coming in late, but how does `continueUUID` actually work?  Are we assuming that the records will always be coming in the same order?  Are we using an `ORDER BY uuid` clause?  Because otherwise I don't think there's any guarantee on the ordering (which follows the UUID PK discussion we've been having elsewhere).

rh-hemartin:
> I'm coming in late, but how does `continueUUID` actually work? Are we assuming that the records will always be coming in the same order? Are we using an `ORDER BY uuid` clause? Because otherwise I don't think there's any guarantee on the ordering (which follows the UUID PK discussion we've been having elsewhere).

Check the queries, there are a number of changes that include ordering by creationTimestamp and uuid.

rh-hemartin:
> I don't know how I feel about having to check if `limit`, `continueUUID`, and `continueDate` are empty string, but I don't have a better alternative to recommend right now

At some point we need to check. I played around a little with the level in which we check for that and it made sense to me that it was where it is.

ggallen (commented):
I know there's been a lot of discussion around this, and it's almost approved, but I'd like to understand what is going on here.

Is there a design document for pagination that explains what is being implemented and how? I'd like to see that. I'm trying to read the code and figure it out, but reading the code isn't the most efficient way to understand what is going on.

pkg/database/postgresql.go
```diff
@@ -27,11 +27,31 @@ func (info PostgreSQLDatabaseInfo) GetConnectionString() string {
 }
 
 func (info PostgreSQLDatabaseInfo) GetResourcesSQL() string {
-	return "SELECT data FROM resource WHERE kind=$1 AND api_version=$2"
+	return "SELECT data->'metadata'->'creationTimestamp', uuid, data FROM resource WHERE kind=$1 AND api_version=$2 ORDER BY ((data->'metadata'->'creationTimestamp')::text::timestamptz, uuid)"
```
ggallen, line 30: Are we going to need an index on `creationTimestamp`?  Have you checked to see if this query is efficient?

ggallen: I wonder if it makes more sense to create a `BIGSERIAL` PK and use that instead of `creationTimestamp`?

ggallen: What good is it to have `uuid` in the `ORDER BY` clause?  I guess you have to in the case the `creationTimestamp` is equal?  If we use a `BIGSERIAL` for this, I think it would be simpler.

maruiz93: > I wonder if it makes more sense to create a BIGSERIAL PK and use that instead of creationTimestamp?

I don't think it makes more sense. If we want to go with the same approach that Kubernetes API has, the order of the items is based on the time they were created on the cluster.
We can't ensure that the order of creation in KubeArchive database is the same as the order of creation of the resource in the cluster... so for me it makes more sense doing it this way.

> What good is it to have uuid in the ORDER BY clause? I guess you have to in the case the creationTimestamp is equal? If we use a BIGSERIAL for this, I think it would be simpler.

Exactly, it is for managing the case that creationTimestamp will be the same. We already had a discussion around this. Check https://github.com/kubearchive/kubearchive/pull/484#discussion_r1785888691

A BIGSERIAL will solve this, but still, I think we should stick with the order of the elements based on creation timestamp to follow coherence with what Kubernetes does

pkg/database/postgresql.go
```diff
@@ -27,11 +27,31 @@ func (info PostgreSQLDatabaseInfo) GetConnectionString() string {
...
+}
+
+func (info PostgreSQLDatabaseInfo) GetResourcesLimitedSQL() string {
```
ggallen, line 33: The number of routines here is exploding and I fear will be hard to maintain. There's also the duplication of lots of SQL in each of these queries.

It might make sense to go back to a single routine and have all the SQL generated behind the scenes.  Instead, the user would have to pass a map or struct containing parameter values, which they can change or leave empty to be defaulted.  The routine can then generate the SQL to be used.

I think this would 1) cut down the method bloat we see here, 2) reduce duplication of SQL,  and 3) simplify things at the call site. Generating the SQL itself might be a little tricky, but it should be able to be done as the SQL statement itself has to be well-structured.

This might be even easier if go allowed for default values. But it doesn't, so something else needs to be done here. Here's [one blog](https://joneisen.me/programming/2013/06/23/golang-and-default-values.html) with some suggestions.

maruiz93: This was already discussed in https://github.com/kubearchive/kubearchive/pull/484#discussion_r1795734437 and I opened an issue to start addressing refactorings here: https://github.com/kubearchive/kubearchive/issues/537

ggallen: Why are we planning on refactoring code that hasn't even been committed yet? That just seems wrong.

I don't think this feature is a rush, so I'd rather get it right the first time rather than commit it knowing we are already going to change it.

pkg/database/postgresql.go
ggallen: Understood.  I didn't think about the implications of ordering based on the cluster creation timestamp.

And I will admit I didn't read all the comments.  There are too many, which tells me there probably should have been a design document for this.

pkg/database/postgresql.go
maruiz93: ok, now that I am rereading your first comment in this thread, I see that I wasn't talking about the same.

I complained about duplicated code in the `database.go` file, but you are talking explicitly about the SQL sentences (my bad).

That means that my previous comment doesn't apply, and BTW, we created a separate issue because what is proposed is to change code that was already there, not new code.

So, going back to your first suggestion and now that I am focused on the problem that you mentioned: having too many routines with different SQL sentences

> It might make sense to go back to a single routine and have all the SQL generated behind the scenes. Instead, the user would have to pass a map or struct containing parameter values, which they can change or leave empty to be defaulted. The routine can then generate the SQL to be used.

I am not sure of how this will look like.
Now the approach is more explicit, and I usually like that. I also like how with different database integrations just need to worry about implementing these routines.

I can't say anything "bad" about your proposal because I don't envision it right now, and that makes me wonder if it should be in this PR or not.

This PR just keeps going with the last design that we did, refactoring that wouldn't be part of the scope of the pagination issue and could get this PR harder to review.

> There's also the duplication of lots of SQL in each of these queries.

How is this? every SQL statement in this file is different from each other, of course, there are duplication at some parts of the query (mostly, this part: `SELECT data->'metadata'->'creationTimestamp', uuid FROM resource WHERE kind=$1 AND api_version=$2`) but getting this part separated from the rest of the query I think that just get things less readable and maintainable

ggallen: > This PR just keeps going with the last design that we did,

Where is this design?

rh-hemartin: Implementing the pagination I'm just following the same patterns we had, so in a way the code will need to be refactored after merging the pagination feature. I can do the refactor here, but the PR will just get bigger and it will more difficult to review.

pkg/database/postgresql.go
rh-hemartin: I created a design document this morning, it adds a little bit of context, not sure if it helps. https://docs.google.com/document/d/1F21wq8JdPGfgQfWH7YrCOZwF9G4u6OgnRs2XndBCero/edit

pkg/database/postgresql.go
maruiz93: > Where is this design?

I was referring to using the same patterns that we had, the ones that were defined in https://github.com/kubearchive/kubearchive/pull/469

PR #486, opened 2024-09-27
fix bug in integrations/database/postgresql/install.sh

### LINKED ISSUE #485: `hack/quick-install.sh` fails when trying to create postgres database
when I run `hack/quick-install.sh` it fails with the following error

```
error: unrecognized condition: "create"
```

it comes from [this line in `integrations/database/postgresql/install.sh`. ](https://github.com/kubearchive/kubearchive/blob/main/integrations/database/postgresql/install.sh#L19).

skoved:
Get rid of `error: unrecognized condition: "create"` when the script is run as part of hack/quick-install.sh while still waiting for kubearchive-1 pod to start.

Resolves #485 

## Release Note

```release-note
NONE
```

## Notes for Reviewers

`--for=create` is the incorrect syntax, it should be `--for=condition=create` however, this command fails if the pod was not started. For me it can take a couple of seconds so instead I used a while loop to wait until the pod is created.


skoved:
I missed kubectl >= 1.31 in development.md

PR #489, opened 2024-09-29
Represent owners in the database.

### LINKED ISSUE #488: Represent owners in the database.
Add a new table to the database and use triggers to add owners references to it.

ggallen:

Resolves #488

## Release Note

```release-note
NONE
```

### DIFF
--- a/integrations/database/mariadb/kubearchive.sql
+++ b/integrations/database/mariadb/kubearchive.sql
@@ -1,7 +1,7 @@
 /*!999999\- enable the sandbox mode */ 
 -- MariaDB dump 10.19  Distrib 10.11.8-MariaDB, for Linux (x86_64)
 --
--- Host: 127.0.0.1    Database: kubearchive
+-- Host: localhost    Database: kubearchive
 -- ------------------------------------------------------
 -- Server version	11.4.3-MariaDB-ubu2404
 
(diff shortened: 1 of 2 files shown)


integrations/database/mariadb/kubearchive.sql
```diff
@@ -55,11 +70,44 @@ CREATE TABLE `resource` (
...
+      SET len = JSON_LENGTH(refs);
+      WHILE i < len DO
+        SET ref = JSON_EXTRACT(refs, CONCAT('$[', i, ']'));
+        SET owner = JSON_VALUE(ref, '$.uid');
+        IF owner IS NOT NULL THEN
+          INSERT INTO owner (uuid, owner_uuid) VALUES (NEW.uuid, owner) ON DUPLICATE KEY UPDATE uuid=uuid;
+        END IF;
+        SET i = i + 1;
+      END WHILE;
+    END */;;
```
maruiz93, line 92: I am confused about this, it is commented and it doesn't seem to match the same triggers defined for postgresql

maruiz93 (changes requested):

integrations/database/mariadb/kubearchive.sql
ggallen: It's different because the syntax is different. Are you saying the functionality is different?  I believe both are iterating through the `ownerReferences` of the resource and adding those records to the `owner` table.

ggallen: This is also going to change based on what we've learned with indexing.

maruiz93: > It's different because the syntax is different.
I was worried about what seemed to be comments (after searching it, things like `/*!50003 CREATE*/` are conditional execution tokens that mean a minimum version of mariadb 5.00.03 is expected to run the command. 

The other thing I saw different was the name of the trigger, for mariadb is `set_owners` while in postgresql we are usign `insert_owners`. I believe we should be consistent with the names even though they would work regardless as they are internal for the database.

> This is also going to change based on what we've learned with indexing.
Right, but wanted to mention this difference as it caught my attention.

rh-hemartin:
@ggallen we had some discussions about owners in the past weeks, does this PR change because of those discussions? Just so I can start reviewing this.

ggallen:
> @ggallen we had some discussions about owners in the past weeks, does this PR change because of those discussions? Just so I can start reviewing this.

@rh-hemartin , this is superseded by https://github.com/kubearchive/kubearchive/pull/518.

PR #493, opened 2024-10-01
Add a script for deploy in debug mode

### LINKED ISSUE #478: Create a script for deploying kubearchive in debug mode
With the changes of https://github.com/kubearchive/kubearchive/issues/442 we are just using `helm install` command for deploying KubeArchive in debug mode.

This is not consistent with the rest of the deployment instructions as we are moving out from the `helm install` command so we are not dealing with the helm charts in the cluster anymore and with the `helm uninstall` or `helm upgrade` command.

We should provide new helper scripts to deploy KubeArchive in debug mode that address the following:

* Receive a parameter `operator` `sink` or `api` to debug
* Build that image in debug mode with `ko`
* (Re)Apply the proper kubernetes resources to successfully use that image and expose the debugging ports

maruiz93:

Resolves #478

## Release Note

```release-note
NONE
```

## Notes for Reviewers

I end up using the `jq` + `kubectl` apply after trying several approaches with `kubectl patch` but none seemed to fit very good.
Note that we want to remove the probe keys and that is difficult to do with `kubectl patch`.
I also replaced `apply` for `replace` in the `hack/kubearchive-install.sh` because if not, the services keep the `debug` port.

### DIFF
--- a/test/debug/patch-security-context.yaml
+++ b/test/debug/patch-security-context.yaml
@@ -1,0 +1,6 @@
+---
+spec:
+  template:
+    spec:
+      securityContext:
+        runAsNonRoot: false
--- a/test/debug/patch-service.yaml
+++ b/test/debug/patch-service.yaml
@@ -1,0 +1,7 @@
+---
+spec:
+  ports:
+    - name: debug
+      protocol: TCP
+      port: 40000
+      targetPort: 40000
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -150,31 +150,19 @@ export KO_DEFAULTBASEIMAGE=registry.redhat.io/rhel9/support-tools
 **[NOTE]**: Using this images usually needs the pod property `runAsNonRoot` set to `false`.
 
 ## Remote IDE debugging
-   **[NOTE]**: Before running these steps make sure that your cluster has the kubearchive
-   dependencies from running `quick-install.sh` but without kubearchive itself (run `kubearchive-delete.sh`)
 
 Use [delve](https://golangforall.com/en/post/go-docker-delve-remote-debug.html)
 to start a debugger to which attach from your IDE.
 
-1. Deploy the chart with `ko` and `helm` in debug mode using an image with `delve`:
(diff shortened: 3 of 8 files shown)


hack/kubearchive-install.sh
```diff
@@ -19,7 +19,7 @@ helm template kubearchive charts/kubearchive -n kubearchive \
     --include-crds \
     --set "global.production=true" > /tmp/kubearchive-not-resolved.yaml
 ko resolve -f /tmp/kubearchive-not-resolved.yaml --base-import-paths > ${YAML}
-kubectl apply -n kubearchive -f ${YAML}
+kubectl replace -n kubearchive -f ${YAML}
```
rh-hemartin, line 22: I'm testing the PR locally and this does not work if I run the normal process, I think you changed this after installing for the first time. So this command can't `replace` if there are no resources. The process to debug another component needs to change to `hack/kubearchive-delete.sh` and then `hack/kubearchive-install.sh` again.

rh-hemartin: You can see the error same error I got in the integration tests.

rh-hemartin:
This PR needs to remove the corresponding stuff from the Helm chart.

hack/kubearchive-install.sh
maruiz93: Right, I tested just with some resources and see that even when they were not created with replace they were, but it seems that this doesn't apply for every resource in our script.

I've rollbacked the change in this file

maruiz93:
> This PR needs to remove the corresponding pieces from the Helm chart.

Included

test/debug/patch-security-context.yaml
skoved, line 1: I think you should add `---` to the top of the file to get rid of the linter warning

test/debug/patch-service.yaml
skoved, line 1: I think you should add --- to the top of the file to get rid of the linter warning

rh-hemartin (approved):

test/debug/patch-security-context.yaml
maruiz93: Done!

PR #516, opened 2024-10-08
Add documentation for global filters

### LINKED ISSUE #509: Document how to configure global filters
Hey! There are mention of global filters at #223 , but the linked PR contains no documentation that explain how to configure that. I'm trying to come up with some process for it, but its not being successful. The script I'm using (after installing as per the DEVELOPMENT.md) is:

<details>
<summary>Code to reproduce</summary>

```
set -o errexit
set -o xtrace

cat <<EOF | kubectl apply -f -
apiVersion: v1
kind: ConfigMap
metadata:
  name: sink-filters
  namespace: kubearchive
data:
  _global: |
    - selector:
        apiVersion: batch/v1
        kind: CronJob
      archiveWhen: "true"
    - selector:
        apiVersion: batch/v1
        kind: Job
      archiveWhen: has(status.startTime)
    - selector:
        apiVersion: v1
        kind: Pod
      archiveOnDelete: "true"
EOF


cat <<EOF | kubectl apply -f -
kind: KubeArchiveConfig
metadata:
  name: kubearchive
  namespace: default
spec:
  resources: []
EOF

cat <<EOF | kubectl apply -f -
---
apiVersion: batch/v1
kind: CronJob
metadata:
  name: cronjob
spec:
  schedule: "* * * * *"
  successfulJobsHistoryLimit: 10
  failedJobsHistoryLimit: 2
  jobTemplate:
    spec:
      template:
        spec:
          containers:
            - name: generate
              image: busybox:latest
              imagePullPolicy: IfNotPresent
              args: [date]
          restartPolicy: OnFailure
EOF

```
</details>

This modifies the `_global_` property on the `sink-filters` ConfigMap and then creates an empty KubeArchiveConfig on my target namespace. Is this the right approach?

PD: I got an error while doing this, I'm opening another issue with the error.

Edit: the code in this description is no longer the last version, I added the new version in a comment below.

rh-hemartin:

Resolves #509 

## Release Note

```release-note
NONE
```

### DIFF
--- a/docs/modules/ROOT/pages/configuration/global-filters.adoc
+++ b/docs/modules/ROOT/pages/configuration/global-filters.adoc
@@ -1,0 +1,68 @@
+= Configuring Global Filters on KubeArchive
+
+KubeArchive allows cluster owners to configure global filters. These are applied
+to any KubeArchive namespace containing a `KubeArchiveConfig` custom resource.
+
+To configure the global filters create a `KubeArchiveConfig` in the
+"kubearchive" namespace. When it is created in the "kubearchive" namespace
+these filters are considered global:
+
+[source,yaml]
+----
+---
+apiVersion: kubearchive.kubearchive.org/v1alpha1
+kind: KubeArchiveConfig
+metadata:
+  name: kubearchive
+  namespace: kubearchive
+spec:
+  resources:
+    - selector:
+        apiVersion: v1
+        kind: Pod
+      archiveOnDelete: "true"
+----
+
+== Interaction With Local Filters
+
+Global filters only work in namespaces containing a `KubeArchiveConfig` custom
+resource. These can be understood as "local" filters in opposition with global
+filters. Both global and local filters are combined with an "OR" type of logic
+and used for that specific namespace.
+
+For example the following `KubeArchiveConfig` in the namespace "my-team" KubeArchive
+archives jobs as long as they have `status.startTime` and deletes them when they
+have `status.completionTime`. However KubeArchive also archives pods when they are
+about to be deleted, as configured in the global filters in the previous section:
+
+[source,yaml]
+----
+---
+apiVersion: kubearchive.kubearchive.org/v1alpha1
+kind: KubeArchiveConfig
+metadata:
+  name: kubearchive
+  namespace: my-team
+spec:
+  resources:
+    - selector:
+        apiVersion: batch/v1
+        kind: Job
+      archiveWhen: has(status.startTime)
+      deleteWhen: has(status.completionTime)
+----
+
+For example the following `KubeArchiveConfig` in the namespace "production" KubeArchive
+only archives pods when they are about to be deleted, as configured in the global filters
+in the previous section:
+
+[source,yaml]
+----
+---
+apiVersion: kubearchive.kubearchive.org/v1alpha1
+kind: KubeArchiveConfig
+metadata:
+  name: kubearchive
+  namespace: production
+spec: {}
+----
--- a/docs/modules/ROOT/nav.adoc
+++ b/docs/modules/ROOT/nav.adoc
@@ -3,6 +3,7 @@
 
 * Configuration
 ** xref:configuration/kubearchiveconfig.adoc[KubeArchiveConfig CRD]
+** xref:configuration/global-filters.adoc[Global Filters]
 ** xref:configuration/cache-expiration-time.adoc[Cache Expiration Time]
 ** xref:configuration/enable-observability.adoc[Enable Observability]
 


docs/modules/ROOT/pages/configuration/global-filters.adoc
```diff
@@ -0,0 +1,48 @@
+= Configuring Global Filters on KubeArchive
+
+KubeArchive allows cluster owners to configure a set of filters that are applied
+to any KubeArchive-enabled namespace. To configure them follow the next steps:
```
maruiz93, line 4: What does "KubeArchive-enabled namespace" mean? Should the user set up specific permissions to the namespace they want to be managed by a global workspace?

docs/modules/ROOT/pages/configuration/global-filters.adoc
```diff
@@ -0,0 +1,48 @@
...
+    - selector:
+        apiVersion: batch/v1
+        kind: Job
+      archiveWhen: has(status.startTime)
+      deleteWhen: has(status.completionTime)
+----
++
+In the namespace `my-team` KubeArchive archives jobs as long as they have `status.startTime`
+and deletes them when they have `status.completionTime`. KubeArchive also archives pods when
+they are about to be deleted, as configured in the global filters.
```
maruiz93, line 48: As a user I have some questions after reading this example:

* With this example I get that KubeArchiveConfig is "merged" for any namespace with the global one in the `kubearchive` namespace, but does the global KubeArchiveConfig apply on `my-team` namespace if I don't have any `KubeArchiveConfig` configured?

* Also, what if I want `my-team` namespace not to have their pods archived? How should I configure the KubeArchiveConfig in `kubearchive` namespace for that?

I believe that the second question could be answered with a `KubeArchiveConfig` like the following:

```
---
kind: KubeArchiveConfig
metadata:
  name: kubearchive
  namespace: kubearchive
spec:
  resources:
    - selector:
        apiVersion: v1
        kind: Pod
        selector:
          matchExpressions:
            - key: prevent-archive
              operator: DoesNotExist
      archiveOnDelete: "true"
```

And make sure that the pods of the `my-team` namespace have that label applied

maruiz93 (changes requested):

docs/modules/ROOT/pages/configuration/global-filters.adoc
rh-hemartin: Hmm.. I tried to come up with a fancy word for "namespace where you already have a KubeArchiveConfig", but since its not mentioned anywhere else I can use the full phrase. Any other ideas?

docs/modules/ROOT/pages/configuration/global-filters.adoc
rh-hemartin: > With this example I get that KubeArchiveConfig is "merged" for any namespace with the global one in the kubearchive namespace, but does the global KubeArchiveConfig apply on my-team namespace if I don't have any KubeArchiveConfig configured?

The docs say:

>  Create a `KubeArchiveConfig` on the namespace you want the global filters to apply. This
configuration could be empty or add its own filters as well. Both global and per-namespace
filters are combined. For example:

I think that solves the question.

> Also, what if I want my-team namespace not to have their pods archived? How should I configure the KubeArchiveConfig in kubearchive namespace for that?

With this I can be more clear. There is nothing you can do to avoid Pods being archived in this scenario other than removing the filter from the global filters.

maruiz93: > I think that solves the question.

Yes, my bad, I overlooked that paragraph.

> With this I can be more clear. There is nothing you can do to avoid Pods being archived in this scenario other than removing the filter from the global filters.

I think it's good to be explicit about that but why not showing also the workaround of the labels?
I have also checked our current [KubeArchiveConfig CRD section](https://kubearchive.github.io/kubearchive/main/configuration/kubearchiveconfig.html) and there is nothing related to the label filtering, any reason not to show an example of the usage?

docs/modules/ROOT/pages/configuration/global-filters.adoc
maruiz93: Maybe we could go with the same approach than with the acronyms: stating at the begining what "KubeArchive-enabled namespace" means so we can refer to it in the rest of the page.

Something like:

```To get a namespaced managed by a global filter the namespace must contain a KubeArchiveConfig. This CR could have the following placeholder values if no aditional archiving logic need to be provided:

<insert a KAC CR with placeholder values>

From now on we will refer to these namespaces as "KubeArchive-enabled namespaces".
```

rh-hemartin: We don't use it here that much, I opted for the phrase.

maruiz93: I am ok with that. But what about showing an example of a dummy KubeArchiveConfig resource so that the namespace take the configuration of the global filter without adding new filters?
For that case could this work?

```yaml
---
kind: KubeArchiveConfig
metadata:
  name: kubearchive
  namespace: ns
spec:
  resources: []
```

maruiz93: I see - https://github.com/kubearchive/kubearchive/issues/526

ggallen: I think this works:
```---
kind: KubeArchiveConfig
metadata:
  name: kubearchive
  namespace: ns
spec:
  ```
  The `resources` are optional.

Comment (ggallen) [reply docs/modules/ROOT/pages/configuration/global-filters.adoc]: Ah, I see the bug in #526. We still need to create the global filters in this case.

Comment (ggallen) [reply docs/modules/ROOT/pages/configuration/global-filters.adoc]: > That does not work currently, we have a story to solve it.

From the users perspective this may appear to be working!

Comment (ggallen) [reply docs/modules/ROOT/pages/configuration/global-filters.adoc]: > "any namespace where you already have a KubeArchiveConfig"

I would suggest "any namespace containing a `KubeArchiveConfig` custom resource".

File: docs/modules/ROOT/pages/configuration/global-filters.adoc (hunk 2 above)
Comment (ggallen) [reply docs/modules/ROOT/pages/configuration/global-filters.adoc]: > if the cluster owners are willing to let cluster users avoid their filters.

I don't understand this statement. This is *not* possible.

Global filters are always or'd with the users filters, so they are always in effect. The user cannot control this at all. Essentially, all they can do is add additional filtering.

**Comment by rh-hemartin** [comment]
Gave it another rewrite

File: docs/modules/ROOT/pages/configuration/global-filters.adoc (hunk 2 above)
Comment (maruiz93) [reply docs/modules/ROOT/pages/configuration/global-filters.adoc]: Imagine that the owner of the cluster has configured that a resource of a specific kind, say, `Pods` need to be always deleted and archived when finished.

Let's say me and my team are running some tests on our namespace of a controller that manages the pods. 
I want the pods to be removed by the controller and I don't need the data of those pods because that's not the point of the test.

I may want to talk to the cluster admin and say "hey, please give me a way to label or categorize these pods I am creating so that they aren't managed by kubearchive when they finish". 
If that's the case the admin may want to modify the global filter to provide a label that means "do not archive/delete" this `Pod`.

My point here was to provide an example of this type of global configuration filter for a case like this.

Of course, global filters are added to the users filters, but if the cluster admin *is willing* to provide a label for preventing the archive/deletion of the resource, that could be done.

**Review by maruiz93 — APPROVED** [review]

File: docs/modules/ROOT/pages/configuration/global-filters.adoc [hunk 3]
```diff
@@ -0,0 +1,95 @@
...
+  namespace: production
+spec: {}
+----
+
+== Skipping Global Filters
+
+Global filters are enforced but cluster owners can configure them so users
+can choose to avoid them. The following `KubeArchiveConfig` allows
+users to specify the label "prevent-archive" in their pods so they are not
+archived:
```
skoved, line 75: maybe I missed discussion about this, but this is also functionality that will need to be added to the sink

skoved (approved):
There's functionality missing in the sink for some of what is described here but otherwise LGTM :+1:

docs/modules/ROOT/pages/configuration/global-filters.adoc
rh-hemartin: I'm pretty sure this functionality is provided by the CEL expressions. If the global filters says "archive everything but things starting with "abc", you can prefix your thing with "abc" and it won't be archived. Does this make sense to you?

docs/modules/ROOT/pages/configuration/global-filters.adoc
ggallen: I don't see how this works. Are you saying this is some functionality that `ApiServerSource` provides that would allow exclusion of pods with a particular label?  We have changed to a global (single) `ApiServerSource` instance, and I don't think there's a way there to provide this functionality per-namespace.

docs/modules/ROOT/pages/configuration/global-filters.adoc
ggallen: > Global filters are enforced but cluster owners can configure them so users
can choose to avoid them. The following `KubeArchiveConfig` allows
users to specify the label "prevent-archive" in their pods so they are not
archived:

No, no, no. This is not permitted.

rh-hemartin: > No, no, no. This is not permitted.

I'm not sure if you are joking :rofl:

We can't enforce what the cluster owner writes in their KubeArchiveConfig custom resources, so they can have any exceptions they want.

ggallen: >We can't enforce what the cluster owner writes in their KubeArchiveConfig custom resources, so they can have any exceptions they want.

Actually, we can. They can write it, but we don't have to use it.

See [this code](https://github.com/kubearchive/kubearchive/blob/main/cmd/operator/internal/controller/kubearchiveconfig_controller.go#L435-L441).

Even if we allowed the global configs to use it, there's no way for a user to specify `ApiServerSource` configuration specific to their namespace.

rh-hemartin: I see. So this means that the specific example is wrong, as `kar.Selector.Selector` is not passed. However this is achievable with the CEL expressions. Is this right?

ggallen: > However this is achievable with the CEL expressions. Is this right?

What is achievable? The user overriding what is specified in the global filters?  No, I don't think it is.

rh-hemartin: > What is achievable? The user overriding what is specified in the global filters? No, I don't think it is.

That is clear, users can't override global filters.

In a scenario where the global `archiveWhen` is:

```
!(has(metadata.labels) && 'prevent-archive' in metadata.labels)
```

If a user has a resource that contains the `prevent-archive` label, that resource does not get archived. This is not the user avoiding the global filters, but the cluster owners providing a hole in their filter so users can choose to skip it.

ggallen: Sure, that will work. The cluster admin is allowing it.

But the user isn't doing anything with their filters to allow it to happen.

I don't know why we have to document this specific use case.

docs/modules/ROOT/pages/configuration/global-filters.adoc
```diff
@@ -0,0 +1,95 @@
+= Configuring Global Filters on KubeArchive
+
+KubeArchive allows cluster owners to configure global filters. Those are applied
```
skoved, line 3:
```suggestion
KubeArchive allows cluster owners to configure global filters. These filters are applied
```

docs/modules/ROOT/pages/configuration/global-filters.adoc
```diff
@@ -0,0 +1,95 @@
...
+to any KubeArchive namespace containing a `KubeArchiveConfig` custom resource.
+
+To configure the global filters create a `KubeArchiveConfig` in the
+"kubearchive" namespace. Because it is created on the "kubearchive" namespace
```
skoved, line 7:
```suggestion
"kubearchive" namespace. When it is created in the "kubearchive" namespace
```

docs/modules/ROOT/pages/configuration/global-filters.adoc
```diff
@@ -0,0 +1,95 @@
...
+----
+
+== Interaction With Local Filters
+
+Global filters only work in namespaces containing a `KubeArchiveConfig` custom
+resource. These can be understood as "local" filters in opposition with global
+filters. Both global and local filters are combined and used for that specific
+namespace.
+
+For example the following `KubeArchiveConfig` on the namespace "my-team" KubeArchive
```
skoved, line 33:
```suggestion
For example the following `KubeArchiveConfig` in the namespace "my-team" KubeArchive
```

docs/modules/ROOT/pages/configuration/global-filters.adoc
```diff
@@ -0,0 +1,95 @@
...
+spec:
+  resources:
+    - selector:
+        apiVersion: batch/v1
+        kind: Job
+      archiveWhen: has(status.startTime)
+      deleteWhen: has(status.completionTime)
+----
+
+For example the following `KubeArchiveConfig` on the namespace "production" KubeArchive
```
skoved, line 55:
```suggestion
For example the following `KubeArchiveConfig` in the namespace "production" KubeArchive
```

docs/modules/ROOT/pages/configuration/global-filters.adoc
maruiz93: I believe that this information is useful for the cluster admins, in case they want to configure kubeArchive in a flexible way for the users and also for the users in case they want to start a conversation with the cluster admin and see if an exception could be provided.

docs/modules/ROOT/pages/configuration/global-filters.adoc
```diff
@@ -0,0 +1,95 @@
...
+        kind: Pod
+      archiveOnDelete: "true"
+----
+
+== Interaction With Local Filters
+
+Global filters only work in namespaces containing a `KubeArchiveConfig` custom
+resource. These can be understood as "local" filters in opposition with global
+filters. Both global and local filters are combined and used for that specific
+namespace.
```
skoved, line 31: you mention that both the local and global filters are combined, but I think we should explicitly mention that we combine the results using an or, not an and

docs/modules/ROOT/pages/configuration/global-filters.adoc
skoved: I see this as more of tip rather than a feature, but think pointing out stuff like this can be useful

ggallen: We can document this if you like. But we can't give tips on everything. We have to assume some knowledge of CEL expressions and how they can be used.  There's logic involved here, and there's no way around it.

You can also have things archived after a certain time interval has passed.  Or if the name begins with a certain prefix. Or if they have an annotation.  Or a particular status. There's a million different options.

 And while this is possible and your use case is plausible, it's not what we were asked for.  The request was to be able to set global filtering to *force* archiving regardless of what the user specifies.  What's implemented allows for both.

maruiz93: > The request was to be able to set global filtering to force archiving regardless of what the user specifies

The thing is that when I read that global filtering "forces" archiving, I can't help to ask: but is there any way to avoid that or create exceptions?

My intention wasn't to document exactly one way of not applying the global filters for all resources of a kind but just letting know the reader that CEL expressions are flexible enough to generate exceptions for archiving a resource so the different namespaces can generate "non-archivable" resources if the global filter allows that.

If you think that giving an example is too much, I am ok with using just a sentence in a "TIP" section.

rh-hemartin: After let it simmer for a bit, I will remove the block and merge without it. I there is a lot of questions around this we can add the block again to the documentation.

rh-hemartin:
I removed the block about skipping global filters. If we get the question a lot from users we can add a section about it.

mafh314 (approved):

docs/modules/ROOT/pages/configuration/global-filters.adoc
skoved: closing this thread since the lines for this comment were removed

PR #534, opened 2024-10-10
use environment variable to get the key for global filters

### LINKED ISSUE #524: sink uses the wrong key to find the global filters in the ConfigMap
the sink assumes that the `sink-filters` ConfigMap in the KubeArchive namespace will have the global filters under the key `_global` but they will actually be in under the key `kubearchive`.

This const in [`cmd/sink/filters/filters.go`](https://github.com/kubearchive/kubearchive/blob/main/cmd/sink/filters/filters.go#L26) needs to be changed to `kubearchive`

skoved:

Resolves #524 

## Release Note

```release-note
NONE
```

## Notes for Reviewers

Instead of using a hardcoded string for the global filters key, it can be obtained from an environment variable

### DIFF
--- a/cmd/sink/filters/filters.go
+++ b/cmd/sink/filters/filters.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"log/slog"
 	"maps"
+	"os"
 	"sync"
 
 	"github.com/google/cel-go/cel"
@@ -25,12 +26,18 @@ import (
 )
 
 const (
-	globalKey     = "kubearchive"
-	k9eNamespace  = "kubearchive"
+	globalKeyEnvVar = "KUBEARCHIVE_NAMESPACE"
 	filtersCmName = "sink-filters"
 )
 
-var ErrNoGlobal = errors.New("no global expressions exist")
+var (
+	ErrNoGlobal = errors.New("no global expressions exist")
+	globalKey   string // gets set in init() and should be treated as const
+)
+
+func init() {
+	globalKey = os.Getenv(globalKeyEnvVar)
+}
 
 type Filters struct {
 	*sync.RWMutex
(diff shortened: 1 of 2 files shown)


cmd/sink/filters/filters.go
```diff
@@ -23,12 +23,18 @@ import (
...
-	globalKey       = "kubearchive"
+	globalKeyEnvVar = "KUBEARCHIVE_NAMESPACE"
 	mountPathEnvVar = "MOUNT_PATH"
 )
 
 var ErrNoGlobal = errors.New("no global expressions exist")
 
+// getGlobalKey returns the key where global filters can be found in the sink-filters ConfigMap. Will return empty
+// string if the environment variable with the name specified by globalKeyEnvVar is not set.
+func getGlobalKey() string {
```
ggallen, line 34: What is the advantage of having a function in this case?

cmd/sink/filters/filters.go
```diff
@@ -283,7 +289,7 @@ func (f *Filters) handleFsEvent(event fsnotify.Event) {
 			slog.Info("Could not read global filters", "err", err)
 			break
 		}
-		if fileName == globalKey {
+		if fileName == getGlobalKey() {
```
ggallen, line 292: Why not call this once, instead of here and 30 lines down?

cmd/sink/filters/filters.go
maruiz93: I agree that calling `os.Getenv(globalKeyEnvVar)` twice could be avoided.

Let me suggest the usage of an IIF ([Immediately Invoked Function](https://golangbyexample.com/function-golang-complete-guide/#IIF_or_Immediately_Invoked_Function))
```go
var globalKey = func() string {
    return os.Getenv(globalKeyEnvVar)
}()
```

ggallen: > Let me suggest the usage of an IIF ([Immediately Invoked Function](https://golangbyexample.com/function-golang-complete-guide/#IIF_or_Immediately_Invoked_Function))

But I still don't see then point of a one line IIF.  Why not just call `os.Getenv` directly?

skoved: maybe in the future we decide that `globalKey` won't get passed as an environment variable. we could just change this function instead of having to change it in 4 different places

cmd/sink/filters/filters.go
skoved: done

cmd/sink/filters/filters.go
maruiz93: also, I know that `os.Getenv` isn't a instruction that has a high cost, but I feel that it's ok to load this parameter just once. In general terms I like to prevent doing reads, even from the environment variables, if it's not needed.

skoved: turns out that with Marta's suggestion as is, it the IIF can be called more than once. I moved the call to `os.Getenv` to the `init` function which Go guarantees will only be run once even if the module is imported multiple times

maruiz93: interesting, I thought that it would run only once

maruiz93 (approved):

rh-hemartin (approved):

mafh314 (approved):

PR #559, opened 2024-10-16
Merged QueryCoreResources and QueryResources

### LINKED ISSUE #537: Refactor the methods of the Database interface
The [DBInterface](https://github.com/kubearchive/kubearchive/blob/main/pkg/database/database.go#L28) currently has methods whose implementations are really similar:

* `QueryResources` & `QueryCoreResources`
* `QueryNamespacedResources` & `QueryNamespacedCoreResources`
* `QueryNamespacedResourceByName` & `QueryNamespacedCoreResourceByName`

A core resource and a non-core resource only differentiate on the apigroup, core resources don't have group, just version (e.g. `v1`) while non-core resources have both (e.g. `apps/v1`)

We are starting to have duplicated code and we need to figure out if we want to keep having separated methods on the interface and use private methods to handle the duplicated logic or merge those methods and just receive an "apiGroup", which makes sense considering that in kubernetes you don't explicitly differentiate core and non-core resources.

EDIT based on the decided path:

**Acceptance Criteria**:
* Remove `QueryCoreResources` in favor of `QueryResources`
* Remove `QueryNamespacedCoreResources` in favor of `QueryNamespacedResources`
* Remove `QueryNamespacedCoreResourceByName` in favor of `QueryNamespacedResourceByName`
* Merge the parameters `group` and `version` of these methods to `apiVersion`
* Move the logic of building the `apiversion` from `group` and `version` to the callers

rh-hemartin:

Resolves #537

## Release Note

```release-note
NONE
```

## Notes for Reviewers

Extra things:
* Fixed a block in one method regarding limit, continueToken and pagination
* Added propert `t.Run` to some subtests.
* Fixed some leftovers from pagination: removed limit == "" logic, as its never empty since we use a default.
* Default to some limit using `DefaultQuery`.
* Removed queries without limit, as there is always a limit.

### DIFF
--- a/cmd/api/routers/routers.go
+++ b/cmd/api/routers/routers.go
@@ -4,6 +4,7 @@
 package routers
 
 import (
+	"fmt"
 	"net/http"
 	"os"
 	"time"
@@ -43,6 +44,7 @@ type List struct {
 func (c *Controller) GetAllResources(context *gin.Context) {
 	group := context.Param("group")
 	version := context.Param("version")
+	apiVersion := fmt.Sprintf("%s/%s", group, version)
 	limit, uuid, date := pagination.GetValuesFromContext(context)
 
 	kind, err := discovery.GetAPIResourceKind(context)
@@ -51,25 +53,20 @@ func (c *Controller) GetAllResources(context *gin.Context) {
 		return
 	}
 
-	resources, lastUUID, lastDate, err := c.Database.QueryResources(context.Request.Context(), kind, group, version, limit, uuid, date)
+	resources, lastUUID, lastDate, err := c.Database.QueryResources(context.Request.Context(), kind, apiVersion, limit, uuid, date)
 	if err != nil {
 		abort.Abort(context, err.Error(), http.StatusInternalServerError)
 		return
 	}
 
-	var continueToken string
-	if limit != "" {
-		// If the limit was specified, we create the token otherwise
-		// send an empty string
-		continueToken = pagination.CreateToken(lastUUID, lastDate)
-	}
-
+	continueToken := pagination.CreateToken(lastUUID, lastDate)
 	context.JSON(http.StatusOK, NewList(resources, continueToken))
 }
 
 func (c *Controller) GetNamespacedResources(context *gin.Context) {
 	group := context.Param("group")
 	version := context.Param("version")
+	apiVersion := fmt.Sprintf("%s/%s", group, version)
 	namespace := context.Param("namespace")
 	limit, uuid, date := pagination.GetValuesFromContext(context)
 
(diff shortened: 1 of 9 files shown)


pkg/database/database.go
```diff
@@ -34,13 +34,9 @@ type DBInfoInterface interface {
...
-	QueryNamespacedCoreResourceByName(ctx context.Context, kind, version, namespace, name string) (*unstructured.Unstructured, error)
-
-	QueryResources(ctx context.Context, kind, group, version, limit, continueUUID, continueDate string) ([]*unstructured.Unstructured, string, string, error)
-	QueryNamespacedResources(ctx context.Context, kind, group, version, namespace, limit, continueUUID, continueDate string) ([]*unstructured.Unstructured, string, string, error)
-	QueryNamespacedResourceByName(ctx context.Context, kind, group, version, namespace, name string) (*unstructured.Unstructured, error)
+	QueryResources(ctx context.Context, kind, version, limit, continueUUID, continueDate string) ([]*unstructured.Unstructured, string, string, error)
+	QueryNamespacedResources(ctx context.Context, kind, version, namespace, limit, continueUUID, continueDate string) ([]*unstructured.Unstructured, string, string, error)
+	QueryNamespacedResourceByName(ctx context.Context, kind, version, namespace, name string) (*unstructured.Unstructured, error)
```
maruiz93, line 39: I would suggest renaming the `version` parameter to `apiVersion` (the same name used in the k8s resources field to wrap either the `version` or the `group/version`)
```suggestion
	QueryResources(ctx context.Context, kind, apiVersion, limit, continueUUID, continueDate string) ([]*unstructured.Unstructured, string, string, error)
	QueryNamespacedResources(ctx context.Context, kind, apiVersion, namespace, limit, continueUUID, continueDate string) ([]*unstructured.Unstructured, string, string, error)
	QueryNamespacedResourceByName(ctx context.Context, kind, apiVersion, namespace, name string) (*unstructured.Unstructured, error)
```

cmd/api/routers/routers.go
```diff
@@ -51,17 +53,17 @@ func (c *Controller) GetAllResources(context *gin.Context) {
...
-	var continueToken string
-	if limit != "" {
-		// If the limit was specified, we create the token otherwise
-		// send an empty string
-		continueToken = pagination.CreateToken(lastUUID, lastDate)
+	continueToken := pagination.CreateToken(lastUUID, lastDate)
+	if limit == "" {
+		// When the limit is empty, we return an empty continueToken
+		// so there is no `continue` in the metadata answer
+		continueToken = ""
```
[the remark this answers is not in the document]
rh-hemartin: This was the only different block, so I opted for changing it to the common one. However you have a point in calling `CreateToken` when its not needed.

Also you made me remember that now `limit` cant' be an empty string, since we are setting a default in `pagination.Middleware`. So I'm removing these conditionals and doing more changes related to `limit` not being an empty string.

pkg/database/database.go
rh-hemartin: It makes sense, apiVersion is the correct term here. Thanks.

maruiz93 (approved):

PR #560, opened 2024-10-16
Sink now applies global filters as well

### LINKED ISSUE #526: sink should create filters for a kind in all namespaces even if there is no namespaced filter provided for that kind
When the sink is creating filters for each namespace with a KubeArchiveConfig, filters should be generated for global filters when there is no associated namespaced filter for the same kind.

For example if I have a global archiveOnDelete filter for Pods and I have an empty KubeArchiveConfig in the namespace foo, a cel expression should be generated for the global archiveOnDelete filter that will be executed when Pods are deleted in foo

rh-hemartin:

Resolves #526 

## Release Note

```release-note
NONE
```

## Notes for Reviewers

* My original intention was to refactor the code so global and local are treated in the same way and then do "namespace filters" + "global filters". However I found this approach in which we keep treating global filters in the same way. I will probably open an issue to refactor it. I think making this work has more priority than refactoring.

### DIFF
--- a/cmd/sink/filters/filters.go
+++ b/cmd/sink/filters/filters.go
@@ -27,7 +27,7 @@ import (
 
 const (
 	globalKeyEnvVar = "KUBEARCHIVE_NAMESPACE"
-	filtersCmName = "sink-filters"
+	filtersCmName   = "sink-filters"
 )
 
 var (
@@ -56,17 +56,16 @@ func NewFilters(clientset kubernetes.Interface) *Filters {
 		delete:          make(map[NamespaceGroupVersionKind]cel.Program),
 		archiveOnDelete: make(map[NamespaceGroupVersionKind]cel.Program),
 	}
-
 }
 
-// getGlobalCelExprs returns three maps that have GroupVersionKind strings as keys and cel expression strings as values.
+// getGlobalCelExprs returns three maps:
 // The first map is archive cel expressions, the second map is delete cel expressions, and the third map is archive on
 // delete cel expression. If no global key exists, it returns ErrNoGlobal. Otherwise it will return any error it
 // encounters to yaml.Unmarshal the []KubeArchiveConfigResources for the global key.
-func getGlobalCelExprs(stringResources map[string]string) (map[string]string, map[string]string, map[string]string, error) {
-	archiveExprs := make(map[string]string)
-	deleteExprs := make(map[string]string)
-	archiveOnDelete := make(map[string]string)
+func getGlobalCelExprs(stringResources map[string]string) (map[schema.GroupVersionKind]string, map[schema.GroupVersionKind]string, map[schema.GroupVersionKind]string, error) {
+	archiveExprs := make(map[schema.GroupVersionKind]string)
+	deleteExprs := make(map[schema.GroupVersionKind]string)
+	archiveOnDelete := make(map[schema.GroupVersionKind]string)
 	kacResources, exists := stringResources[globalKey]
 	if !exists {
 		return archiveExprs, deleteExprs, archiveOnDelete, ErrNoGlobal
@@ -77,9 +76,9 @@ func getGlobalCelExprs(stringResources map[string]string) (map[string]string, ma
 	}
 	for _, resource := range resources {
 		gvk := schema.FromAPIVersionAndKind(resource.Selector.APIVersion, resource.Selector.Kind)
-		archiveExprs[gvk.String()] = resource.ArchiveWhen
-		deleteExprs[gvk.String()] = resource.DeleteWhen
-		archiveOnDelete[gvk.String()] = resource.ArchiveOnDelete
+		archiveExprs[gvk] = resource.ArchiveWhen
+		deleteExprs[gvk] = resource.DeleteWhen
+		archiveOnDelete[gvk] = resource.ArchiveOnDelete
 	}
 	return archiveExprs, deleteExprs, archiveOnDelete, nil
 }
@@ -121,7 +120,7 @@ func (f *Filters) changeGlobalFilters(stringResources map[string]string) error {
 // not nil, the maps returned can still be used.
 func (f *Filters) createFilters(
 	namespace, kacResources string,
-	globalArchive, globalDelete, globalArchiveOnDelete map[string]string,
+	globalArchive, globalDelete, globalArchiveOnDelete map[schema.GroupVersionKind]string,
 ) (
 	map[NamespaceGroupVersionKind]cel.Program,
 	map[NamespaceGroupVersionKind]cel.Program,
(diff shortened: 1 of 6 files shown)


skoved (changes requested):
we need to make sure that errors that come from trying to compile the global CEL expression get captured.

Otherwise this looks good and thanks for adding unit test for the global filters!

cmd/sink/filters/filters.go
```diff
@@ -128,34 +127,83 @@ func (f *Filters) createFilters(
...
+		// compiling and adding any global filter that was not added already. Here the resources that
+		// are globally filtered but not locally filtered are added.
+		for globalGvk, expression := range globalArchive {
+			namespaceGvk := NamespaceGVKFromNamespaceAndGvk(namespace, globalGvk)
+			_, exists := archiveMap[namespaceGvk]
+			if exists {
+				continue
+			}
+
+			expressionCEL, err := ocel.CreateCelExprOr(expression)
```
ggallen, line 170: Can and should we rename `CreateCelExprOr`? 

What this method does is "or" the given CEL expressions together and return the compiled CEL program.  How about `CompileOrdCELExpression?  It really only does the "or" and then calls `CompileCELExpression`.

And can we standardize on "Cel" or CEL"?  From what I see in the documentation, they use  "CEL".

cmd/sink/filters/filters.go
```diff
@@ -128,34 +127,83 @@ func (f *Filters) createFilters(
...
-			archiveOnDeleteExpr, err := ocel.CreateCelExprOr(resource.ArchiveOnDelete, globalArchiveOnDelete[gvk.String()])
+			archiveOnDeleteExpr, err := ocel.CreateCelExprOr(resource.ArchiveOnDelete, globalArchiveOnDelete[gvk])
 			if err != nil {
 				errList = append(errList, err)
 			} else {
 				archiveOnDeleteMap[namespaceGvk] = *archiveOnDeleteExpr
 			}
 		}
+
+		// Now we cycle the global filters `globalArchive`, `globalDelete` and `globalArchiveOnDelete`
```
ggallen, line 160: The three loops below are identical except for the global and local maps.  Why not split them out into a private method that can be called for each of the global/local map pairs?

cmd/sink/filters/filters.go
```diff
@@ -128,34 +127,83 @@ func (f *Filters) createFilters(
 	if err != nil {
 		return archiveMap, deleteMap, archiveOnDeleteMap, err
 	}
+
 	var errList []error
 	for _, resource := range resources {
 		gvk := schema.FromAPIVersionAndKind(resource.Selector.APIVersion, resource.Selector.Kind)
 		namespaceGvk := NamespaceGVKFromNamespaceAndGvk(namespace, gvk)
 		if resource.ArchiveWhen != "" {
```
ggallen, line 135: The three ifs starting here are all identical code. Why not split out these into a local method that can be called for each?

cmd/sink/filters/filters.go
```diff
@@ -128,34 +127,83 @@ func (f *Filters) createFilters(
...
+			namespaceGvk := NamespaceGVKFromNamespaceAndGvk(namespace, globalGvk)
+			_, exists := archiveMap[namespaceGvk]
+			if exists {
+				continue
+			}
+
+			expressionCEL, err := ocel.CreateCelExprOr(expression)
+			if err != nil {
+				continue
+			}
```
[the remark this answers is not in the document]
rh-hemartin: Right.

cmd/sink/filters/filters.go
```diff
@@ -128,34 +127,83 @@ func (f *Filters) createFilters(
...
+			namespaceGvk := NamespaceGVKFromNamespaceAndGvk(namespace, globalGvk)
+			_, exists := archiveOnDeleteMap[namespaceGvk]
+			if exists {
+				continue
+			}
+
+			expressionCEL, err := ocel.CreateCelExprOr(expression)
+			if err != nil {
+				continue
+			}
```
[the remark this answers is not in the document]
rh-hemartin: Done.

cmd/sink/filters/filters.go
```diff
@@ -128,34 +127,83 @@ func (f *Filters) createFilters(
...
+			namespaceGvk := NamespaceGVKFromNamespaceAndGvk(namespace, globalGvk)
+			_, exists := deleteMap[namespaceGvk]
+			if exists {
+				continue
+			}
+
+			expressionCEL, err := ocel.CreateCelExprOr(expression)
+			if err != nil {
+				continue
+			}
```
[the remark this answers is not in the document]
rh-hemartin: Thanks for checking!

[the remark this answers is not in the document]
rh-hemartin: It helped surface a problem I was not seeing.

cmd/sink/filters/filters.go
rh-hemartin: Done.

cmd/sink/filters/filters.go
rh-hemartin: I was trying to not change too much, but done.

cmd/sink/filters/filters.go
rh-hemartin: I changed it.

cmd/sink/filters/filters.go
```diff
@@ -128,38 +127,87 @@ func (f *Filters) createFilters(
...
-		if resource.ArchiveWhen != "" {
-			archiveExpr, err := ocel.CreateCelExprOr(resource.ArchiveWhen, globalArchive[gvk.String()])
-			if err != nil {
-				errList = append(errList, err)
-			} else {
-				archiveMap[namespaceGvk] = *archiveExpr
-			}
+
+		errs := addLocalFilters(resource.ArchiveWhen, globalArchive[gvk], archiveMap, namespaceGvk)
+		if len(errs) > 0 {
```
ggallen, line 137: Do we even need this check?  Why not just always append?

rh-hemartin: Nice, I like it. Done.

cmd/sink/filters/filters.go
```diff
@@ -49,17 +49,16 @@ func NewFilters(clientset kubernetes.Interface) *Filters {
 		delete:          make(map[NamespaceGroupVersionKind]cel.Program),
 		archiveOnDelete: make(map[NamespaceGroupVersionKind]cel.Program),
 	}
-
 }
 
 // getGlobalCelExprs returns three maps that have GroupVersionKind strings as keys and cel expression strings as values.
```
skoved, line 54: need to update the doc for this function

```suggestion
// getGlobalCelExprs returns three maps that have GroupVersionKind as keys and cel expression strings as values.
```

skoved (changes requested):

cmd/sink/filters/filters.go
ggallen: Aha!

cmd/sink/filters/filters.go
```diff
@@ -128,36 +127,61 @@ func (f *Filters) createFilters(
...
-			archiveExpr, err := ocel.CreateCelExprOr(resource.ArchiveWhen, globalArchive[gvk.String()])
-			if err != nil {
-				errList = append(errList, err)
-			} else {
-				archiveMap[namespaceGvk] = *archiveExpr
-			}
+
+		errList = addLocalFilters(resource.ArchiveWhen, globalArchive[gvk], archiveMap, namespaceGvk, errList)
+		errList = addLocalFilters(resource.DeleteWhen, globalDelete[gvk], deleteMap, namespaceGvk, errList)
+		errList = addLocalFilters(resource.ArchiveOnDelete, globalArchiveOnDelete[gvk], archiveOnDeleteMap, namespaceGvk, errList)
```
skoved, line 138: I think errors from calls to `addLocalFilters` will be lost. I think there needs to be a call to append
```suggestion
		errList = append(errList, ...addLocalFilters(resource.ArchiveWhen, globalArchive[gvk], archiveMap, namespaceGvk, errList))
		errList = append(errlist, ...addLocalFilters(resource.DeleteWhen, globalDelete[gvk], deleteMap, namespaceGvk, errList))
		errList = append(errlist, ...addLocalFilters(resource.ArchiveOnDelete, globalArchiveOnDelete[gvk], archiveOnDeleteMap, namespaceGvk, errList))
```

cmd/sink/filters/filters.go
```diff
@@ -128,36 +127,61 @@ func (f *Filters) createFilters(
...
+	}
+
+	// Now we cycle the global filters `globalArchive`, `globalDelete` and `globalArchiveOnDelete`
+	// compiling and adding any global filter that was not added already. Here the resources that
+	// are globally filtered but not locally filtered are added.
+	errList = addGlobalFilters(globalArchive, archiveMap, namespace, errList)
+	errList = addGlobalFilters(globalDelete, deleteMap, namespace, errList)
+	errList = addGlobalFilters(globalArchiveOnDelete, archiveOnDeleteMap, namespace, errList)
```
skoved, line 146: same as above, but with `addGlobalFilters` instead of `addLocalFilters`

skoved (changes requested):
I think some errors still get lost

cmd/sink/filters/filters.go
rh-hemartin: The errors are now appended to `errList` within the function itself. I can change it back to append them in the caller function if that makes it easier.

ggallen: > to have the errors get logged by the sink, appending needs to happen in both places.

Why? The error list is passed into the method. New errors are appended to the list and returned, and that return value become the error list.  What is wrong?

skoved: > The error list is passed into the method

I missed that. This request is not relevant

cmd/sink/filters/filters.go
skoved: this can be ignored

skoved (approved):

skoved:
> I think some errors still get lost

I was wrong about this

rh-hemartin:
@ggallen could you approve this PR when you are OK with the changes?

ggallen (approved):

rh-hemartin:
@skoved @ggallen I had to do a major change, so I re-requested a review, to be able to test because of the implementation of #534, the `init` function won't let me specify an environment variable properly. Let me know your thoughts.

maruiz93 (approved):

ggallen (approved):
> because the init function provides a poor experience for testing.

In general I am going to say I hate this reasoning. The code got more complex  in order to support testing. That doesn't seem right.

skoved:
`globalKey` is not a constant, so it can be changed. My thought is that we can leave things as they were before but we can do some setup before running `filter_test.go` by creating a [TestMain function](https://pkg.go.dev/testing#hdr-Main) in it. This function runs before all the other Test* functions. ` It could look something like this

```
func TestMain(m *testing.M) {
    globalKey = "kubearchive" // or some other value
    m.Run()  // runs the Test* functions
    // clean up if necessary
}
```

skoved (changes requested):
I don't think that we need to make an changes to non test code to get the tests to work properly. `globalKey` is a `var`, not a `const` and that means that we can assign a value to it. We can combine this with a [TestMain function](https://pkg.go.dev/testing#hdr-Main). `TestMain` runs before any of the `Test* functions in the file and it could look something like this

```
func TestMain(m *testing.M) {
    globalKey = "kubearchive"
    // more setup if necessary
    m.Run() // runs all the Test* functions
    // cleanup if necessary
}
```

rh-hemartin:
I implemented what you suggested and it works properly, thanks for the suggestion.

ggallen:
> I don't think that we need to make an changes to non test code to get the tests to work properly.

Then there was/is no need for the last refactoring of this PR?

skoved (approved):

cmd/sink/filters/filters_test.go
```diff
@@ -15,13 +18,19 @@ import (
...
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/kubernetes/fake"
 	"k8s.io/client-go/tools/cache"
 	toolsWatch "k8s.io/client-go/tools/watch"
 	sourcev1 "knative.dev/eventing/pkg/apis/sources/v1"
 )
 
+func TestMain(m *testing.M) {
+	globalKey = "kubearchive"
+	os.Exit(m.Run())
```
skoved, line 31: will this prevent other test files that would run after this one will not run?

skoved: [From the docs](https://pkg.go.dev/testing#hdr-Main):

> If TestMain returns, the test wrapper will pass the result of m.Run to os.Exit itself

I think this line is unecessary

rh-hemartin: The linter told me it was needed, I disabled the linter for the function and removed the `os.Exit()` call.

PR #564, opened 2024-10-17
Use mktemp to generate tmp files in bash scripts

### LINKED ISSUE #410: Use mktemp rather than hard coding temp file names.
For temporary file safety in bash scripts, the code should use mktemp rather than hard coding temp file names.

Here are a few places I saw hard coded temp file names:

- cmd/operator/generate.sh line 19 (which uses $$ to create unique name based on PID, but mktemp is safer)
- hack/quick-install.sh line 36
- hack/release.sh line 64

The documentation has a few examples that should be updated as well:

- docs/modules/ROOT/pages/getting-started/installation.adoc line 46 and 53

mafh314:

Resolves #410 

## Release Note

```release-note
NONE
```

## Notes for Reviewers

Testing:
- Deploy from scratch (this validates the change in the operator script and kuberachive-install.sh).
- release.sh using a fork.

### DIFF
--- a/cmd/operator/generate.sh
+++ b/cmd/operator/generate.sh
@@ -16,8 +16,8 @@ cd ${SCRIPT_DIR}
 echo "Generating CRD."
 ${LOCALBIN}/controller-gen crd paths="./..." output:dir=../../charts/kubearchive/crds
 
-patch="/tmp/crd.$$"
-cat << EOF > $patch
+PATCH=$(mktemp -t crd.XXXXXXXX)
+cat << EOF > ${PATCH}
 strategy: Webhook
 webhook:
   clientConfig:
@@ -30,9 +30,9 @@ webhook:
 EOF
 CRD="../../charts/kubearchive/crds/kubearchive.kubearchive.org_kubearchiveconfigs.yaml"
 yq -i '.metadata.annotations."cert-manager.io/inject-ca-from"="kubearchive/kubearchive-operator-certificate"' ${CRD}
-yq -i ".spec.conversion = load(\"$patch\")" ${CRD}
+yq -i ".spec.conversion = load(\"${PATCH}\")" ${CRD}
 
-rm -f $patch
+rm -f ${PATCH}
 
 # Generate role.
 echo "Generating role."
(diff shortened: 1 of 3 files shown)


cmd/operator/generate.sh
```diff
@@ -16,8 +16,8 @@ cd ${SCRIPT_DIR}
 echo "Generating CRD."
 ${LOCALBIN}/controller-gen crd paths="./..." output:dir=../../charts/kubearchive/crds
 
-patch="/tmp/crd.$$"
-cat << EOF > $patch
+PATCH=$(mktemp -t crd.XXXXXXXX)
+cat << EOF > $PATCH
```
maruiz93, line 20: Several [styleguides](https://google.github.io/styleguide/shellguide.html#s5.6-variable-expansion) suggest the usage of the [shell expansion format](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html) in variables over the plain usage of variable (meaning ${PATCH} better than $PATCH.

Given that we are already using the expansion format for some variables (like `CRD`) I suggest we keep consistency with it and use it in the rest:

```suggestion
cat << EOF > ${PATCH}
```

cmd/operator/generate.sh
```diff
@@ -30,9 +30,9 @@ webhook:
 EOF
 CRD="../../charts/kubearchive/crds/kubearchive.kubearchive.org_kubearchiveconfigs.yaml"
 yq -i '.metadata.annotations."cert-manager.io/inject-ca-from"="kubearchive/kubearchive-operator-certificate"' ${CRD}
-yq -i ".spec.conversion = load(\"$patch\")" ${CRD}
+yq -i ".spec.conversion = load(\"$PATCH\")" ${CRD}
```
maruiz93, line 33: Following my previous suggestion:
```suggestion
yq -i ".spec.conversion = load(\"${PATCH}\")" ${CRD}
```

cmd/operator/generate.sh
```diff
@@ -30,9 +30,9 @@ webhook:
...
 
-rm -f $patch
+rm -f $PATCH
```
maruiz93, line 35: Following my previous suggestion:
```suggestion
rm -f ${PATCH}
```

maruiz93 (changes requested):

skoved (approved):

cmd/operator/generate.sh
mafh314: Done!

PR #573, opened 2024-10-18
Add host and runtime metrics

### LINKED ISSUE #490: Get system metrics
In order to get better at performance testing we need to gather system metrics, at least **memory** and **cpu** usage. I see two options, but there may be more:

* Use OpenTelemetry in some way so we send this information to the OTLP endpoint, just as we do with traces or other metrics.
* Use Kubernetes directly to get these system metrics directly from pods.

rh-hemartin:

Resolves #490 

## Release Note

```release-note
KubeArchive now sends Go host and runtime metrics using the OpenTelemetry instrumentation.
```

## Notes for Reviewers

* I set an arbitrary number as resources. 256Mi is reasonable for now.
* GOMEMLIMIT and GOMAXPROCS help the golang runtime figure out which resources has available. GOMAXPROCS gets set to 1 at minimum, even when cpu is requested as 500m (half a cpu).
* I added the port 9090 which goes to prometheus, as sometimes is faster than going into Grafana if I want a simpler view.


maruiz93 (approved):

skoved (approved):

charts/kubearchive/templates/api_server/api_server.yaml
```diff
@@ -36,6 +43,16 @@ spec:
               value: "disabled"
             - name: OTEL_EXPORTER_OTLP_ENDPOINT
               value: ""
+            - name: OTEL_GO_X_DEPRECATED_RUNTIME_METRICS
+              value: "false"
+            - name: GOMEMLIMIT
+              valueFrom:
+                resourceFieldRef:
+                  resource: limits.memory
```
skoved, line 51: [The Go Garbage Collector Documentation](https://tip.golang.org/doc/gc-guide#Memory_limit) has this to say about setting memory limits when deploying a go application in a container

```
a good rule of thumb is to leave an additional 5-10% of headroom to account for memory sources the Go runtime is unaware of
```

based this, we might wanna set the value as 90%-95% of `limits.memory` instead

charts/kubearchive/templates/operator/operator.yaml
```diff
@@ -63,6 +63,16 @@ spec:
               value: "disabled"
             - name: OTEL_EXPORTER_OTLP_ENDPOINT
               value: ""
+            - name: OTEL_GO_X_DEPRECATED_RUNTIME_METRICS
+              value: "false"
+            - name: GOMEMLIMIT
+              valueFrom:
+                resourceFieldRef:
+                  resource: limits.memory
```
skoved, line 71: same as above

charts/kubearchive/templates/sink/sink.yaml
```diff
@@ -26,6 +33,16 @@ spec:
...
+            - name: OTEL_GO_X_DEPRECATED_RUNTIME_METRICS
+              value: "false"
+            - name: GOMEMLIMIT
+              valueFrom:
+                resourceFieldRef:
+                  resource: limits.memory
+            - name: GOMAXPROCS
+              valueFrom:
+                resourceFieldRef:
+                  resource: limits.cpu
```
skoved, line 45: same as above

charts/kubearchive/templates/api_server/api_server.yaml
rh-hemartin: Specifying it this way is very convenient... What about setting  `requests.memory` to 90% and use it to set the GOMEMLIMIT? That would change the QoS class of the pod to Burstable, but I think we can manage with that.

skoved: I think that will work :+1:

rh-hemartin: Done

mafh314 (approved):

PR #576, opened 2024-10-22
Sink generates log urls for pods

### LINKED ISSUE #538: Sink should generate log URLs for Pods
Whenever a `Pod` is archived by KubeArchive, the sink should generate log URLs for the `Pod` if logging is configured.

### Configuration

The sink should check the KubeArchive configuration `ConfigMap` to see if logging is enabled. If it is, it should read the logging `ConfigMap` specified there and create a map for generating log URLs from the contents of that `ConfigMap`.  That map should contain key-value pairs for everything in the `ConfigMap`.  Values starting with the prefix "cel:" should be compiled as CEL expressions, and the value for those keys containing CEL expressions should be the CEL program that resulted from the compilation.  All other values are copied to the map as strings.

The logging `ConfigMap` needs to be watched for changes, and the map regenerated when changes to it are detected.

This map should be generated only once and reused for all log URL generation.

### Log URL Generation

If a `Pod` is archived and log URL generation is enabled as described above, the sink needs to generate the log URLs for that `Pod` and store them in the `log_url` table. These steps should be done:

1. All existing entries in the `log_url` table for the`Pod` should be deleted.
1. Log URLs are generated by calling `log_url.log_url.GenerateLogURLs, passing in the map of key-value pairs to be used for substitution.
1. For each log URL returned by the previous step, add a new record to the `log_url` table for this `Pod`.

### Acceptance Criteria

1. The sink detects and watches for logging configuration and handles it appropriately.
1. Log URLs are generated for `Pods` and stored in the table. 

### References

- [Logging design document](https://github.com/kubearchive/kubearchive/blob/main/docs/modules/ROOT/pages/design/logging-integration.adoc)

skoved:

Resolves #538 

## Release Note

```release-note
sink generates log urls for pods
```

## Notes for Reviewers

These are some of the changes made to enable logging in the sink

- add database interface method to return SQL query to delete rows in the `log_url` table based on `uuid`
- add database interface method to return SQL query to add add rows to the `log_url` table
- add database method that deletes all rows that have the same `uuid` as the kubernetes object passed to it and then write all urls passed to it to the database.
- update fake database test type to support the write and delete operations added
- refactored some code in `cmd/sink/main.go` make it easier to write log urls to the database in the right places


cmd/sink/logs/logs.go
```diff
@@ -0,0 +1,80 @@
...
+
+const (
+	configCmName     = "kubearchive-config"
+	loggingCmKeyName = "logging-configmap-name"
+	logUrlKey        = "LOG_URL"
+	celPrefix        = "cel:"
+)
+
+func getKubeArchiveLoggingCm(ctx context.Context, kubeClient kubernetes.Interface) (*corev1.ConfigMap, error) {
+	kaConfigCm, err := kubeClient.CoreV1().ConfigMaps("kubearchive").Get(
```
skoved, line 27: I plan to replace `"kubearchive"` with a constant

skoved: done

cmd/sink/logs/logs.go
```diff
@@ -0,0 +1,81 @@
...
+		ctx,
+		configCmName,
+		metav1.GetOptions{},
+	)
+	if err != nil {
+		return nil, fmt.Errorf("Could not get ConfigMap: %w", err)
+	}
+	loggingCmName, exists := kaConfigCm.Data[loggingCmKeyName]
+	if !exists {
+		return nil, fmt.Errorf("No logging ConfigMap was provided")
```
ggallen, line 38: Why is this an error?  Logging is not required.

cmd/sink/k8s/values.go
```diff
@@ -0,0 +1,18 @@
+// Copyright KubeArchive Authors
```
ggallen, line 1: Why does this merit its own package?

cmd/sink/main.go
```diff
@@ -72,13 +75,56 @@ func NewSink(db database.DBInterface, filters *filters.Filters) *Sink {
...
+			"err", err,
+		)
+	} else {
+		slog.Info(
+			"Successfully wrote object from cloudevent to the database",
+			"objectID", obj.GetUID(),
+			"id", event.ID(),
+		)
+	}
+	sink.writeLogs(ctx, obj)
```
ggallen, line 125: Why call this method if you don't have a `Pod`?

cmd/sink/k8s/values.go
skoved: This is not it's own package. This is one that already existed in the sink that provides the clients. Both  `cmd/sink/filters` and `cmd/sink/logs` need to determine what namespace this sink is in. I thought by putting it this package, which both can import, would reduces duplicated code.

cmd/sink/logs/logs.go
```diff
@@ -0,0 +1,81 @@
...
+
+const (
+	configCmName     = "kubearchive-config"
+	loggingCmKeyName = "logging-configmap-name"
+	logUrlKey        = "LOG_URL"
+	celPrefix        = "cel:"
+)
+
+func getKubeArchiveLoggingCm(ctx context.Context, kubeClient kubernetes.Interface) (*corev1.ConfigMap, error) {
+	kaConfigCm, err := kubeClient.CoreV1().ConfigMaps(k8s.KubeArchiveNamespace).Get(
```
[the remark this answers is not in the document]
skoved: see https://github.com/kubearchive/kubearchive/issues/538#issuecomment-2412173712

cmd/sink/logs/logs.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/kubearchive/kubearchive/pkg/log_urls"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	configCmName     = "kubearchive-config"
+	loggingCmKeyName = "logging-configmap-name"
```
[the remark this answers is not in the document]
skoved: We can do that and it will make it so that we don't have to change the code here if the the name of the `ConfigMap` or the key name for the logging `ConfigMap` change. However, there will still need to be a consts here with the names of the environment variables

cmd/sink/main.go
```diff
@@ -72,13 +75,56 @@ func NewSink(db database.DBInterface, filters *filters.Filters) *Sink {
...
+		Filters:       filters,
+		K8sClient:     k8sClient,
+		logUrlBuilder: builder,
 	}
 }
 
+func (sink *Sink) writeLogs(ctx context.Context, obj *unstructured.Unstructured) {
+	kind := strings.ToLower(obj.GetKind())
+	if kind != "pod" {
+		slog.Info("Object does not have logs associated with it", "objectID", obj.GetUID(), "kind", kind)
```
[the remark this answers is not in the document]
skoved: done

cmd/sink/main.go
skoved: writeLogs is only called when `kind == Pod` now

cmd/sink/logs/logs.go
skoved: `getKubeArchiveLoggingCm` no longer returns errors. If the an issue is encountered it returns a `ConfigMap` with an empty `Data` map

cmd/sink/logs/logs.go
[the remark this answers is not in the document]
skoved: and https://github.com/kubearchive/kubearchive/issues/538#issuecomment-2413753724

cmd/sink/logs/logs.go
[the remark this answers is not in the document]
skoved: unless you mean something different in which case, can you show me an example of what you're talking about?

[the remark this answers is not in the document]
ggallen: Something simple like this:
```spec:
  mariaDbRef:
    name: kubarchive
```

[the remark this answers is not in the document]
skoved: done!

ggallen (commented):
Please add the documentation that tells admins they have to manually rollout for the changes to take effect.

cmd/sink/logs/logs.go
```diff
@@ -0,0 +1,93 @@
...
+
+func getKubeArchiveLoggingCm(ctx context.Context, kubeClient kubernetes.Interface) *corev1.ConfigMap {
+	emptyCm := &corev1.ConfigMap{Data: map[string]string{}}
+	kaConfigCm, err := kubeClient.CoreV1().ConfigMaps(configCmRef.Namespace).Get(
+		ctx,
+		configCmRef.Name,
+		metav1.GetOptions{},
+	)
+	if err != nil {
+		return emptyCm
```
rh-hemartin, line 44: We need to log the error to show whats happening, maybe something like:
```suggestion
        slog.Warn("error while getting configmap", "configmap", fmt.Sprintf("%s/%s", configCmRef.Namespace, configCmRef.Name), "error", err.Error())
		return emptyCm
```

cmd/sink/logs/logs.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+	var exists bool
+	loggingCmRef.Name, exists = kaConfigCm.Data[loggingCmSelector.Key]
+	if !exists {
+		return emptyCm
```
rh-hemartin, line 49:
```suggestion
        slog.Warn("logging information does not exist", "key", loggingCmSelector.Key)
		return emptyCm
```

cmd/sink/logs/logs.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+	loggingCm, err := kubeClient.CoreV1().ConfigMaps(loggingCmRef.Namespace).Get(
+		ctx,
+		loggingCmRef.Name,
+		metav1.GetOptions{},
+	)
+	if err != nil {
+		return emptyCm
```
rh-hemartin, line 57:
```suggestion
        slog.Warn("error while getting configmap", "configmap", fmt.Sprintf("%s/%s", loggingCmRef.Namespace, loggingCmRef.Name), "error", err.Error())
		return emptyCm
```

rh-hemartin:
I am testing with:

```
bash test/log-generators/cronjobs/install.sh --num-jobs=10
cat <<EOF | kubectl apply -f -
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: kubearchive-splunk
  namespace: kubearchive
data:
  POD_ID: cel:metadata.uid
  CONTAINER_NAME: cel:spec.containers.map(m, m.name)
  POD: 'spath "kubernetes.pod_id" | search "kubernetes.pod_id"="{POD_ID}"'
  CONTAINER: 'spath "kubernetes.container_name" | search "kubernetes.container_name"="{CONTAINER_NAME}"'
  LOG_URL: http://127.0.0.1:8111/app/search/search?q=search * | {POD} | {CONTAINER}
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: kubearchive-config
  namespace: kubearchive
data:
  logging: |
    configMapRef: splunk-logging
EOF
kubectl exec -i -n postgresql pods/kubearchive-1 -- psql kubearchive -c "SELECT * FROM log_url;" | tail
```

The count increases but I'm not getting any URL. Following my own suggestions, I see this one the warns:

```
2024/10/29 10:14:13 WARN logging information does not exist key=logging-configmap-name keys=[logging]
```

So `logging-configmap-name` is what the logging configuration option should be called?

PD: the configMaps are from https://kubearchive.github.io/kubearchive/main/integrations/logging.html

skoved:
I had been following what was described here https://kubearchive.github.io/kubearchive/main/design/logging-integration.html#_configuration

I think we need @ggallen to help sort this out

cmd/sink/logs/logs.go
skoved: done

skoved:
> Please add the documentation that tells admins they have to manually rollout for the changes to take effect.

updated the documentation

ggallen:
> I think we need @ggallen to help sort this out

That's an illegal `ConfigMap` as @skoved pointed out.  The documentation incorrect. I'm not sure what I was thinking.

rh-hemartin (changes requested):
This needs to update the documentation at https://kubearchive.github.io/kubearchive/main/integrations/logging.html to fix the ConfigMap presented there.

rh-hemartin (approved):

pkg/database/mariadb.go
```diff
@@ -55,6 +55,14 @@ func (info MariaDBDatabaseInfo) GetWriteResourceSQL() string {
 		"ON DUPLICATE KEY UPDATE name=?, namespace=?, resource_version=?, cluster_deleted_ts=?, data=?"
 }
 
+func (info MariaDBDatabaseInfo) GetWriteUrlSQL() string {
+	return "INSERT INTO log_url (uuid, url) VALUES (?, ?)"
```
maruiz93, line 59: I am currently focused on addressing the retrieval of the log URLs without the container filter but based on a previous discussion I think that we agreed to add a `container` column on the `log_url` table and that the sink should introduce it so it's easier to make the filter.

This is a quite large PR so I'm ok with addressing it separately

ggallen (approved):

maruiz93 (approved):

skoved:
rebased the branch so I can merge this

Issue #527, opened 2024-10-10
Introduce Continuous Performance Testing


rh-hemartin:
We had a conversation with performance folks and they recommended us to start to test our performance regularly. The idea is that we run a GitHub job, as we do with integration tests, in which we generate traffic, measure the application and report metrics. This should help us identify if queries/inserts are getting slower, consuming more memory, etc.

The starting metrics we want starting are:

* Memory and CPU usage
* Event processing latency (POST to the sink and measure the latency)
* Query latency (GET to the API and measure the latency)

We don't have a number of queries to perform, but overall should take less than 10 minutes (integration tests are around 7 minutes when I'm writing this).

skoved:
> Event processing latency (POST to the sink and measure the latency)

the `testing` package allows you to write [`Benchmark` functions](https://pkg.go.dev/testing#hdr-Benchmarks), which I think would be useful.

By default `go test` will run tests first, then benchmarks, the fuzzing. Depending on how we have things set up, we might want to change the `go test` action to skip benchmarks and have a separate action that runs daily to run the benchmarks.

It might be possible to create container images for the database that already contain the data that we need for benchmarking, in which case it might be possible to run the benchmarks as part of CI and have the action comment the results on the PR using the github api.

rh-hemartin:
The last piece of this is #855  and I think we can call this done.

PR #578, opened 2024-10-23
Change API Gin metrics package

rh-hemartin:

Related to #527 527

## Release Note

```release-note
NONE
```

## Notes for Reviewers

* The current instrumentation package for gin does not provide metrics, I'm changing it to one that does.
* The current instrumentation used is deprecated and has no maintainer: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/6190

### DIFF
--- a/cmd/api/auth/authn_test.go
+++ b/cmd/api/auth/authn_test.go
@@ -67,7 +67,7 @@ func TestInvalidAuthHeader(t *testing.T) {
 			ftr := &fakeTokenReview{authenticated: false}
 			res := httptest.NewRecorder()
 			c, _ := gin.CreateTestContext(res)
-			c.Request, _ = http.NewRequest(http.MethodGet, "", nil)
+			c.Request = httptest.NewRequest(http.MethodGet, "/", nil)
 			c.Request.Header = tc.header
 
 			Authentication(ftr, cache, cacheExpirationDuration, cacheExpirationDuration)(c)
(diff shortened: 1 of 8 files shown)


maruiz93:
Can you please develop the need of why should this change done? Maybe by linking a ticket with the proposed change?

As far as I am aware the previous package `go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin` is the "official" one and the proposed one doesn't have details on what differentiates it from the official package.

maruiz93 (changes requested):
Can you please develop the need of why should this change done? Maybe by linking a ticket with the proposed change?

As far as I am aware the previous package `go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin` is the "official" one and the proposed one doesn't have details on what differentiates it from the official package.

rh-hemartin:
> Can you please develop the need of why should this change done? Maybe by linking a ticket with the proposed change?
> 
> As far as I am aware the previous package `go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin` is the "official" one and the proposed one doesn't have details on what differentiates it from the official package.

I added some reviewers notes to the PR.

maruiz93 (approved):

rh-hemartin:
Hey @maruiz93 and @skoved I changed something across all tests. We were using `http.NewRequest` with `nil` as body. This caused the Body of the request to be `nil`, which then made the new middleware we are introducing fail because of that. We changed all `http.NewRequest` to `httptest.NewRequest` which does not set `Body` as `nil` if `nil` is passed as body parameter.

I checked that in the real scenario, when you call the API with curl the body is not nil but an empty string. Instead of passing `strings.NewReader("")` to all `http.NewRequest` I think its better to go with the `httptest` package which already accounts for that.

Let me know what you think about this.

maruiz93 (approved):

PR #581, opened 2024-10-23
Fix operator liveness and readiness probes.

### LINKED ISSUE #439: Customize the readyz endpoint of the operator
We currently use the default `healthz.Ping` function to be run on the [readyz endpoint](https://github.com/maruiz93/kubearchive/blob/main/cmd/operator/main.go#L147) of the operator.

This is not enough, as we are currently manually checking if the `kubearchive-operator-webhooks` service is actually responding on port 443:

```bash
# Make sure webhooks are up and running.
LOCAL_PORT=8443
kubectl -n kubearchive port-forward service/kubearchive-operator-webhooks ${LOCAL_PORT}:443 >& /dev/null &

# Wait for $LOCAL_PORT to become available.
while ! nc -vz localhost ${LOCAL_PORT} > /dev/null 2>&1 ; do
    echo -n .
    sleep 0.5
done
echo .
```

If we don't wait for this check to be successful, and we try to apply a `KubeArchiveConfig` earlier, it won't work.

We should have a readiness probe more sophisticated to do this check and prevent us having to do it after the deployment is successfully rollout.

ggallen:

Resolves #439

## Release Note

```release-note
NONE
```

### DIFF
--- a/charts/kubearchive/templates/operator/operator.yaml
+++ b/charts/kubearchive/templates/operator/operator.yaml
@@ -1,5 +1,5 @@
 {{- $controlPlane := "controller-manager" -}}
-{{- $port := 8081 -}}
+{{- $port := 9443 -}}
 ---
 apiVersion: apps/v1
 kind: Deployment
@@ -76,10 +76,11 @@ spec:
           image: {{ .Values.operator.image }}
           livenessProbe:
             httpGet:
-              path: /healthz
+              path: /validate-kubearchive-kubearchive-org-v1alpha1-kubearchiveconfig
               port: {{ $port }}
+              scheme: HTTPS
             initialDelaySeconds: 15
-            periodSeconds: 20
+            periodSeconds: 60
           ports:
             - containerPort: 9443
               name: webhook-server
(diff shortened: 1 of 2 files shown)


rh-hemartin:
You need to remove the current health endpoint the operator runs, its on the cmd here: https://github.com/kubearchive/kubearchive/blob/main/charts/kubearchive/templates/operator/operator.yaml#L54, https://github.com/kubearchive/kubearchive/blob/main/cmd/operator/main.go#L116

maruiz93:
> You need to remove the current health endpoint the operator runs, its on the cmd here: https://github.com/kubearchive/kubearchive/blob/main/charts/kubearchive/templates/operator/operator.yaml#L54, https://github.com/kubearchive/kubearchive/blob/main/cmd/operator/main.go#L116

Also worth noticing that the endpoints were exposed on the port 8081 [here](https://github.com/ggallen/kubearchive/blob/issue-439/cmd/operator/main.go#L68-L68) so you'll need to remove that line too.

maruiz93:
I also have some doubts about using the validation endpoint for the liveness probe, is this a standard thing to do? I am having a hard time finding similar examples in other operators... so I don't have any specific suggestion but wanted to raise the concern in case you know better where to search

maruiz93 (changes requested):
Also, waiting for kubearchive operator webhooks is also used in https://github.com/kubearchive/kubearchive/blob/main/integrations/logging/elasticsearch/install.sh#L32-L32 (should be commented/removed)

ggallen:
> I also have some doubts about using the validation endpoint for the liveness probe, is this a standard thing to do?

Yeah, I'm doubting it too. I'm trying to find a better solution.

skoved:
@ggallen what's going on with this PR? If you're not working on it could you close it?

rh-hemartin:
Closing, the best approach will be to open a new PR, given that this one is very old.

PR #594, opened 2024-10-25
fix sink filter test setup

### LINKED ISSUE #593: `TestHandleUpdates` in `cmd/sink/filters/filters_test.go` is not setup correctly
`TestHandleUpdates` creates multiple test cases but does not run them as separate test cases. It should call `t.Run` for each one. Additionally, it calls `newMockWatcher` which should accept a `testing.TB` and call `Helper()` to mark itself as a helper function.

skoved:

Resolves #593 

## Release Note

```release-note
NONE
```

## Notes for Reviewers

- `newMockWatcher` registers as a helper function
- `TestHandleUpdates` registers each of its tests as a sub-test with `t.Run`
- change variable name `testCases` -> `tests` as well as loop variable `tc` -> `tt` to be more consistent with other tests

### DIFF
--- a/cmd/sink/filters/filters_test.go
+++ b/cmd/sink/filters/filters_test.go
@@ -4,7 +4,6 @@
 package filters
 
 import (
-	"bytes"
 	"context"
 	"fmt"
 	"io"
@@ -32,7 +31,8 @@ func TestMain(m *testing.M) {
 	m.Run()
 }
 
-func newMockWatcher(ch <-chan watch.Event) func(metav1.ListOptions) (watch.Interface, error) {
+func newMockWatcher(t testing.TB, ch <-chan watch.Event) func(metav1.ListOptions) (watch.Interface, error) {
+	t.Helper()
 	return func(options metav1.ListOptions) (watch.Interface, error) {
 		return watch.MockWatcher{
 			StopFunc:       func() {},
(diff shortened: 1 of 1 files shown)


cmd/sink/filters/filters_test.go
```diff
@@ -142,40 +143,43 @@ func TestHandleUpdates(t *testing.T) {
...
+					Namespace:       "kubearchive",
+					ResourceVersion: fmt.Sprintf("%d", i),
+				},
+				Data: tt.data,
+			}
+			event := watch.Event{Type: watch.Modified, Object: &cm}
+			ch <- event
+			wg.Add(1)
+			go func(rw *toolsWatch.RetryWatcher, filter *Filters) {
+				defer wg.Done()
```
rh-hemartin, line 171: In this case it does not matter, but for parallel tests, which we don't have yet, we should use `t.Cleanup` instead because defer does not guarantee order. See https://brandur.org/fragments/go-prefer-t-cleanup-with-parallel-subtests

skoved: I'll make the change anyway to get myself in the right habit

skoved: I _think_ this needs to remain a defer because `wg.Done` needs to be called when the `goroutine` that its in finishes not when the test it was called in completes. However, there might be something I'm missing

skoved: but I removed the `defer` and made `wg.Done()` a direct call at the end of the function instead

maruiz93 (approved):

rh-hemartin (approved):

skoved:
rebasing so that a merge can happen

PR #599, opened 2024-10-29
Fix result ordering

### LINKED ISSUE #586: Results should be from new to old
Curerntly the results are returned by the API from old entries to new entries, we want to return the new entries first, then the old ones. This is because the default sort for columns is ASC, and we want DESC. I didn't notice when I implemented the pagination (which introduced sorting for the first time).

I did play around with using DESC in the ORDER BY clause, but the uuid column is resisting to be tamed and no matter if the WHERE clause of `(timestamp, uuid) < (value, value)` is kept toghether or we split it `timestamp < value AND uuid > value` the results are not consistent. Yoou should be able to reproduce with +200 objects. Get the first page with 100, then the second page. Then get a page with 200 elements. page 1 + page 2 should equal the page with 200 objects. If you start doing comparison on uuids you will see there are overlappings or entries missed when using DESC.

We checked with a couple of database-knowledgeable people and they recommended us to not use UUIDv4 to sort because its fundamentally unsortable, as its random. They suggested to move to a autincremental ID to use in the pagination so we don't miss entries.

The point of this PR is that we change the return of the objects so the new are returned first. However that requires changing the UUIDv4 to something else. So:

**Acceptance Criteria**

* Return resources from new to old, so invert the current order with `DESC`.
* Introduce a BIGSERIAL as primary key for the `resource` table and use it for ordering alongside the creationTimestamp.

Additionaly:

* Introduce a BIGSERIAL as primary key for the table `log_url` as well.

Edit: changed SERIAL for BIGSERIAL

rh-hemartin:

Resolves #586 

## Release Note

```release-note
Now KubeArchive returns resources by creationTimestamp in descending order (new first)
```

## Notes for Reviewers

* The query gets a little slower in the big dataset because the DESC makes the filter remove more rows.
* Bumped CloudNative PG to 1.24.1 to PostgreSQL so it defaults to PostgreSQL v17.
* `BIGSERIAL` is not a true type, so its not reflected directly here. It is translated to a BIGINT and a series of sequence statements that make the thing autoincrement on insert.
* To test download and load 500.txt into the database, then run:

[500.txt](https://github.com/user-attachments/files/17554742/500.txt)

```
curl -s --cacert ca.crt -H "Authorization: Bearer $(kubectl create token default)" "https://localhost:8081/api/v1/pods?limit=100" | jq -r '.items[] | [.metadata.creationTimestamp, .metadata.uid] | @tsv' > 100-pods.txt
export CONTINUE_TOKEN=$(curl -s --cacert ca.crt -H "Authorization: Bearer $(kubectl create token default)" "https://localhost:8081/api/v1/pods?limit=100" | jq -r '.metadata.continue')
curl -s --cacert ca.crt -H "Authorization: Bearer $(kubectl create token default)" "https://localhost:8081/api/v1/pods?limit=100&continue=${CONTINUE_TOKEN}" | jq -r '.items[] | [.metadata.creationTimestamp, .metadata.uid] | @tsv' > 100-continue-pods.txt
curl -s --cacert ca.crt -H "Authorization: Bearer $(kubectl create token default)" "https://localhost:8081/api/v1/pods?limit=200" | jq -r '.items[] | [.metadata.creationTimestamp, .metadata.uid] | @tsv' > 200-pods.txt

# Should equal 202 (all entries are different + 2 lines introduced by diff)
diff 100-pods.txt 100-continue-pods.txt | wc -l

# Should equal 0 (no difference)
cat 100-pods.txt 100-continue-pods.txt | diff - 200-pods.txt | wc -l
```

### DIFF
--- a/cmd/api/pagination/pagination.go
+++ b/cmd/api/pagination/pagination.go
@@ -13,7 +13,6 @@ import (
 	"time"
 
 	"github.com/gin-gonic/gin"
-	"github.com/google/uuid"
 	"github.com/kubearchive/kubearchive/cmd/api/abort"
 )
 
@@ -21,7 +20,7 @@ const (
 	limitKey        = "limit"
 	continueKey     = "continue"
 	continueDateKey = "continueDate"
-	continueUUIDKey = "continueUUID"
+	continueIdKey   = "continueId"
 	defaultLimit    = "100"
 	maxAllowedLimit = 1000
 )
@@ -30,17 +29,17 @@ const (
 // information needed. This is kept here so its close to the function
 // that sets these values in the context (Middleware)
 func GetValuesFromContext(context *gin.Context) (string, string, string) {
-	return context.GetString(limitKey), context.GetString(continueUUIDKey), context.GetString(continueDateKey)
+	return context.GetString(limitKey), context.GetString(continueIdKey), context.GetString(continueDateKey)
 }
 
-func CreateToken(uuid, date string) string {
+func CreateToken(uuid int64, date string) string {
 	// The date is returned as a quoted string, so remove the quotes
 	date = strings.TrimPrefix(date, "\"")
 	date = strings.TrimSuffix(date, "\"")
-	if date == "" && uuid == "" {
+	if date == "" && uuid == 0 {
 		return ""
 	}
-	tokenString := fmt.Sprintf("%s %s", uuid, date)
+	tokenString := fmt.Sprintf("%d %s", uuid, date)
 	return base64.StdEncoding.EncodeToString([]byte(tokenString))
 }
 
@@ -69,7 +68,7 @@ func Middleware() gin.HandlerFunc {
 		limit = strconv.Itoa(limitInteger)
 
 		var continueDate string
-		var continueUUID string
+		var continueId string
 		if continueToken != "" {
 			continueBytes, err := base64.StdEncoding.DecodeString(continueToken)
 			if err != nil {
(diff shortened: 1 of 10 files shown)


maruiz93 (approved):

rh-hemartin:
> I thought we talked about adding a BIGSERIAL primary key and using that. Or is that a different issue?

`BIGSERIAL` is not a true type, its a BIGINT + some sequence stuff: https://www.postgresql.org/docs/current/datatype-numeric.html#DATATYPE-SERIAL

I created the table with `BIGSERIAL PRIMARY KEY`, but the dump does not return that but the true types and stuff underneath.

ggallen:
> > I thought we talked about adding a BIGSERIAL primary key and using that. Or is that a different issue?
> 
> `BIGSERIAL` is not a true type, its a BIGINT + some sequence stuff: https://www.postgresql.org/docs/current/datatype-numeric.html#DATATYPE-SERIAL
> 
> I created the table with `BIGSERIAL PRIMARY KEY`, but the dump does not return that but the true types and stuff underneath.

But that still doesn't answer my question: are you doing that in this issue or was it another issue?

rh-hemartin:
> But that still doesn't answer my question: are you doing that in this issue or was it another issue?

Its this issue, you can see how I'm adding a BIGINT and sequence stuff. I thought you asked because you saw a BIGINT instead of a BIGSERIAL

ggallen:
> > But that still doesn't answer my question: are you doing that in this issue or was it another issue?
> 
> Its this issue, you can see how I'm adding a BIGINT and sequence stuff. I thought you asked because you saw a BIGINT instead of a BIGSERIAL. I should stop answering non-asked questions, sorry.

No, I just missed that change completely. Thanks.

rh-hemartin:
@ggallen if you reviewed the changes and you agree with them, approve them.

ggallen (approved):

PR #603, opened 2024-10-30
Force KubeArchiveConfig resource to be named 'kubearchive'.

### LINKED ISSUE #585: Enforce a unique name for KubeArchiveConfig resources
Related to https://github.com/kubearchive/kubearchive/issues/352 we want CR `KubeArchiveConfig` resources to be a singleton, meaning, just one instance of each in a namespace.

We aren't currently providing any restirction for naming this resources but if we do, validating this singleton gets simpler:

* The admission webhook would just need to validate if the name is the enforced one (e.g. `kac`) which means that no queries to the Kubernetes API will be needed.
* Kubernetes will take care of denying the creation of a second `KubeArchiveConfig` resource in the same namespace with the same name because by design there can't be two resources of the same kind that share the same name in a namespace.

### Acceptance Criteria
* Decide a name for the `KubeArchiveConfig` resources (e.g. `kac`)
* Add this requirement to the user's documentation
* Update the admission webhook to check for this name

ggallen:

Resolves #585

## Release Note

```release-note
NONE
```

### DIFF
--- a/cmd/operator/api/v1alpha1/kubearchiveconfig_webhook.go
+++ b/cmd/operator/api/v1alpha1/kubearchiveconfig_webhook.go
@@ -4,19 +4,19 @@
 package v1alpha1
 
 import (
-	"context"
 	"errors"
+	"fmt"
 
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
-	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
 	cel "github.com/kubearchive/kubearchive/pkg/cel"
 )
+
+const kubearchiveResourceName = "kubearchive"
 
 // log is for logging in this package.
 var kubearchiveconfiglog = logf.Log.WithName("kubearchiveconfig-resource")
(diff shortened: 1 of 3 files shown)


cmd/operator/api/v1alpha1/kubearchiveconfig_webhook.go
```diff
@@ -4,20 +4,20 @@
...
-	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
 	cel "github.com/kubearchive/kubearchive/pkg/cel"
 )
 
+const kubearchiveResourceName = "kubearchive"
```
maruiz93, line 19: should we maybe have the name provided through an environment variable?

maruiz93 (approved):

cmd/operator/api/v1alpha1/kubearchiveconfig_webhook.go
ggallen: Why? What advantage does that provide?

skoved (approved):

rh-hemartin (approved):

cmd/operator/api/v1alpha1/kubearchiveconfig_webhook.go
maruiz93: just in case that the KubeArchive deployer wanted to override the resourcename with a different one, but it's not that important (I already approved the PR, I just thought that it will provide more versatility with little work), but I think that probably no one will care about it so feel free to ignore the suggestion

ggallen: > just in case that the KubeArchive deployer wanted to override the resourcename with a different one,

But doesn't that go completely against the whole point of this PR?  I guess you're changing or interpreting it to mean "force all `KubeArchiveConfig` resources to have the same name"? I was strictly adhering to "force all `KubeArchiveConfig` resources to be named 'kubearchive'".

If this becomes an problem we can open a new issue to support what I think you are looking for.

PR #608, opened 2024-10-31
add container name to log_url table

### LINKED ISSUE #605: Add container name to log_url table and populate it from the sink.
As noted in [this comment](https://github.com/kubearchive/kubearchive/pull/576#pullrequestreview-2407485221), KubeArchive needs to allow the retrieval of log URLs by container name.  This currently isn't possible as the container name is embedded in the JSON blob.

The `log_url` table should have a new column named `container_name`, which should be populated by the sink when the records are inserted.

### Acceptance Criteria:

- A column named `container_name` is added to the `log_url` table.
- The sink populates the `log_url.container_name` column when inserting log URL records.

skoved:

Resolves #605 

## Release Note

```release-note
adds `container_name` field to the log_url table
sink writes container names for logs to the database
```

## Notes for Reviewers

Adds `container_name` field to the `log_url` table in the database and sink writes the container name along with the log url and uuid

Additionally:
- renamed the `log_urls` package to `logurls` to follow [Go package name recommendations](https://go.dev/blog/package-names)
- renamed constants in `pkg/logurls/logurls.go` to follow https://google.github.io/styleguide/go/decisions.html#constant-names


integrations/database/postgresql/kubearchive.sql
```diff
@@ -67,7 +67,8 @@ SET default_table_access_method = heap;
 
 CREATE TABLE public.log_url (
     uuid uuid NOT NULL,
-    url text NOT NULL
+    url text NOT NULL,
+    container_name text NOT NULL
```
maruiz93, line 71: can we also add created_at and updated_at like we have in the `resource` table?

skoved: sure

skoved: done

pkg/database/fake/database_generated.go
```diff
@@ -35,8 +36,8 @@ func CreateTestLogUrls() []LogUrlRow {
 }
 
 type LogUrlRow struct {
-	Uuid types.UID
-	Url  string
+	Uuid               types.UID
+	ContainerName, Url string
```
ggallen, line 40: This look strange to me.  Why not put them both on their own line with their own types?  Is this a go standard?  I know it's possible to do it this way, but do we want to?

ggallen: >> I know it's possible to do it this way, but do we want to?
> The question of if we want to do this stylistically is different.

Yes, that's what' I'm asking.

ggallen: I've added it to the agenda for Monday.

skoved:
Updated the documentation to reflect that `CONTAINER_NAME` is now a required field in the logging ConfigMap

ggallen:
> Updated the documentation to reflect that `CONTAINER_NAME` is now a required field in the logging ConfigMap

Why? It should not be required.  The sink should determine the container name itself and not rely on the user to do it.

docs/modules/ROOT/pages/integrations/logging.adoc
```diff
@@ -96,8 +96,12 @@ In this example the following happens:
 Logging URL generation in KubeArchive is controlled using a `ConfigMap` in
 the KubeArchive installation namespace.  URLs are generated using parameters
 that allow the URL to point to the specific log file associated with the
-given `Pod`. This `ConfigMap` requires a single entry, LOG_URL, whose string
-value will be interpolated using the other variables defined in the `ConfigMap`.
+given `Pod`. This `ConfigMap` requires two entries:
+
+. LOG_URL, whose string value will be interpolated using the other variables defined in the `ConfigMap`.
+. CONTAINER_NAME, whose value is a CEL expression that will be evaluated with data from the CloudEvent to provide the
```
ggallen, line 102: `CONTAINER_NAME` should not be required. If the sink needs the container name, it knows where to get it from.  It should not depend on the user providing it.

This is completely separate from what is contained in the logging `ConfigMap`.

ggallen (changes requested):

pkg/logurls/logurls.go
```diff
@@ -82,11 +84,11 @@ func generateSubstitutionMaps(m map[string]interface{}) []map[string]string {
 	return vmaps
 }
 
-func interpolate(val string, env map[string]string, r *regexp.Regexp) string {
+func interpolate(val string, env map[string]string, r *regexp.Regexp) models.LogTuple {
 	matches := r.FindAllStringSubmatch(val, -1)
 	if matches == nil {
 		// Finished, nothing more to substitute.
-		return val
+		return models.LogTuple{ContainerName: env[containerName], Url: val}
```
ggallen, line 91: The sink should pull the container name value from the resource JSON, not from the evaluated `CONTAINER_NAME` variable.

pkg/logurls/logurls.go
```diff
@@ -12,17 +12,19 @@ import (
...
+	"github.com/kubearchive/kubearchive/pkg/models"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
 
 const (
-	VARIABLE_REGEX string = "\\{([A-Za-z0-9_]+)\\}"
-	LOG_URL        string = "LOG_URL"
+	variableRegex string = "\\{([A-Za-z0-9_]+)\\}"
+	logURL        string = "LOG_URL"
+	containerName string = "CONTAINER_NAME"
```
ggallen, line 22: This should not be needed.

docs/modules/ROOT/pages/integrations/logging.adoc
skoved: how does the sink know which log url is for which container?

pkg/database/fake/database_generated.go
skoved: moved `ContainerName` to its own line per discussion today

pkg/logurls/logurls.go
skoved: How will interpolate know which container name the url is for with out looking at the evaluated `CONTAINER_NAME`?

ggallen: The log URL code should pass back all the information needed (i.e. the URL and the container name), but it should *not* depend on the `CONTAINER_NAME` variable for the value.  There is no guarantee the user will set it correctly and then bad URLs will result.

I think instead what we should do, and what is probably quickest, is to say that `CONTAINER_NAME` is a predefined variable name and set it to the CEL express to representing the list of container names. Then it can be evaluated and used as needed if a user needs it for generating the URL.

So I think what can be done is this:

- Add `CONTAINER_NAME` to the map as `CONTAINER_NAME: "cel:spec.containers.map(m, m.name)"`, overwriting whatever setting there might be specified in the `ConfigMap`.
- Document this behavior, and note that URLs will be generated for each container in the `Pod`.  This may already be noted in the documentation.

skoved: done

docs/modules/ROOT/pages/integrations/logging.adoc
skoved: I reverted this change since KubeArchive will prove the `CONTAINER_NAME` CEL expression

cmd/operator/api/v1alpha1/kubearchiveconfig_webhook.go
```diff
@@ -60,7 +60,6 @@ func (kac *KubeArchiveConfig) ValidateUpdate(old runtime.Object) (admission.Warn
 		return nil, fmt.Errorf("Cannot update KubeArchiveConfig resource named '%s'.", kubearchiveResourceName)
 	}
```
ggallen, line 62: There is no need for a change in this file.  Why?

cmd/sink/logs/logs.go
```diff
@@ -32,6 +34,9 @@ var (
 	loggingCmSelector = corev1.ConfigMapKeySelector{
 		Key: "logging-configmap-name",
 	}
+	containerName = map[string]string{
```
ggallen, line 37: Why is this a `map` and not just a constant string?

cmd/operator/api/v1alpha1/kubearchiveconfig_webhook.go
skoved: golangci-lint was saying that this file was not `goimport`ed

cmd/sink/logs/logs.go
skoved: This was left over from some previous messing around. Fixed now

cmd/operator/api/v1alpha1/kubearchiveconfig_webhook.go
ggallen: I don't understand. The only change I am seeing here is the removal of a blank line.

ggallen: How could this ever have passed before?

rh-hemartin: The golangci-lint runs on a patch:

https://github.com/kubearchive/kubearchive/actions/runs/11593292755/job/32277025444#step:6:31

So probably that caused an omission of some kind of the double space.

skoved: I'll submit a separate PR for removing the extra newline in this file. I want to get this PR merged

maruiz93 (approved):

cmd/operator/api/v1alpha1/kubearchiveconfig_webhook.go
ggallen: No, just do it here.  It's a waste to have another PR.

ggallen (approved):

PR #609, opened 2024-10-31
Retrieve log urls

### LINKED ISSUE #539: API server should allow retrieval log URLs
The CLI and API server should allow the user to retrieve log URLs associated with archived resources.

Although logs are associated with `Pods`, the log URLs should be able to be retrieved by referencing any `Pod` owner.  For example, if the user asks for logs for a ` PipelineRun`, the API server needs to query the database to get the logs from all `Pods` that executed as part of that `PipelineRun`.

A query to retrieve resources that are owned by another resource might look like this:

```
SELECT uuid
FROM public.resource
WHERE data->'metadata'->'ownerReferences' @> '[{"uid": "2d893024-4ed4-4318-a12d-13cb7f6562bc'"}]'::jsonb;
```

The API server may need to make multiple queries to retrieve all the `Pods` and then get the log URLs.  Or a stored procedure could be created to do all the work, resulting in only a single call being required from the API server.

### Acceptance Criteria

1. The API server ~and CLI~ allows users to query and retrieve log URLs.

NOTE: the CLI part is addressed in a separate issue with its own PR

### References

- [Logging design document](https://github.com/kubearchive/kubearchive/blob/main/docs/modules/ROOT/pages/design/logging-integration.adoc)
- This is dependent upon log URL generation functionality in the sink (see https://github.com/kubearchive/kubearchive/issues/538)
- This issue relates to https://github.com/kubearchive/kubearchive/issues/492.

maruiz93:

Resolves #539

## Release Note

```release-note
Add new endpoints for the retrieval of log URLs
```

## Notes for Reviewers

I've included a new file called `query.go` in the `database` package with helper functions to encapsulate the logic for parsing the sql row results into structs or primitive data.
All the queries (new and old) are using this new functions.

There is a new interface `DBParamParser` to be implemented in the different drivers for transforming arrays into the allowed parameters for parametrized statements. This way we allow having queries of the type `SELECT * FROM ... WHERE key IN ?` being `?` a set of elements of undefined length.

The endpoints return 404 when the resource doesn't exist in the database.

For testing, you can run the integration tests locally, check the namespace used for the generated data and then do:
```
kubectl apply -f test/users/
kubectl get -n kubearchive secrets kubearchive-api-server-tls -o jsonpath='{.data.ca\.crt}' | base64 -d > ca.cr
curl -v --cacert ca.crt -H "Authorization: Bearer $(kubectl create -n test token default)" \
 https://localhost:8081/apis/batch/v1/namespaces/<ns>/cronjobs/generate-log-1/log
```

Unit tests are included but no integration tests. I will open a new issue for addressing those.

### DIFF
--- a/cmd/api/auth/authz.go
+++ b/cmd/api/auth/authz.go
@@ -6,15 +6,16 @@ package auth
 import (
 	"fmt"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/kubearchive/kubearchive/cmd/api/abort"
 	"github.com/kubearchive/kubearchive/pkg/cache"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/gin-gonic/gin"
 	apiAuthnv1 "k8s.io/api/authentication/v1"
 	apiAuthzv1 "k8s.io/api/authorization/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clientAuthzv1 "k8s.io/client-go/kubernetes/typed/authorization/v1"
 )
 
@@ -36,44 +37,59 @@ func RBACAuthorization(sari clientAuthzv1.SubjectAccessReviewInterface, cache *c
 			verb = "get"
 		}
 
-		sarSpec := apiAuthzv1.SubjectAccessReviewSpec{
-			User:   userInfo.Username,
-			Groups: userInfo.Groups,
-			ResourceAttributes: &apiAuthzv1.ResourceAttributes{
-				Namespace: c.Param("namespace"),
-				Group:     c.Param("group"),
-				Version:   c.Param("version"),
-				Resource:  c.Param("resourceType"),
-				Name:      c.Param("name"),
-				Verb:      verb,
-			},
+		sarSpecs := []apiAuthzv1.SubjectAccessReviewSpec{
+			{
+				User:   userInfo.Username,
+				Groups: userInfo.Groups,
+				ResourceAttributes: &apiAuthzv1.ResourceAttributes{
+					Namespace: c.Param("namespace"),
+					Group:     c.Param("group"),
+					Version:   c.Param("version"),
+					Resource:  c.Param("resourceType"),
+					Name:      c.Param("name"),
+					Verb:      verb,
+				},
+			}}
+
+		if strings.HasSuffix(c.Request.URL.Path, "/log") {
+			logSpec := apiAuthzv1.SubjectAccessReviewSpec{
+				User:   userInfo.Username,
+				Groups: userInfo.Groups,
+				ResourceAttributes: &apiAuthzv1.ResourceAttributes{
+					Namespace: c.Param("namespace"),
+					Version:   "v1",
+					Resource:  "pods/log",
+					Verb:      "get",
+				},
+			}
+			sarSpecs = append(sarSpecs, logSpec)
 		}
 
-		allowed := cache.Get(sarSpec.String())
-		if allowed != nil {
-			if allowed != true {
-				abort.Abort(c, "Unauthorized", http.StatusUnauthorized)
+		for _, sarSpec := range sarSpecs {
+			allowed := cache.Get(sarSpec.String())
+			if allowed != nil {
+				if allowed != true {
+					abort.Abort(c, "Unauthorized", http.StatusUnauthorized)
+					return
+				}
+				continue
+			}
+			sar, err := sari.Create(c.Request.Context(), &apiAuthzv1.SubjectAccessReview{
+				Spec: sarSpec,
+			}, metav1.CreateOptions{})
+
+			if err != nil {
+				abort.Abort(c, fmt.Sprintf("Unexpected error on SAR: %s", err.Error()), http.StatusInternalServerError)
 				return
 			}
 
-			c.Next()
-			return
-		}
+			cache.Set(sarSpec.String(), sar.Status.Allowed, cacheExpirationAuthorized)
 
-		sar, err := sari.Create(c.Request.Context(), &apiAuthzv1.SubjectAccessReview{
-			Spec: sarSpec,
-		}, metav1.CreateOptions{})
-
-		if err != nil {
-			abort.Abort(c, fmt.Sprintf("Unexpected error on SAR: %s", err.Error()), http.StatusInternalServerError)
-			return
-		}
-
-		cache.Set(sarSpec.String(), sar.Status.Allowed, cacheExpirationAuthorized)
-
-		if !sar.Status.Allowed {
-			cache.Set(sarSpec.String(), sar.Status.Allowed, cacheExpirationUnauthorized)
-			abort.Abort(c, "Unauthorized", http.StatusUnauthorized)
+			if !sar.Status.Allowed {
+				cache.Set(sarSpec.String(), sar.Status.Allowed, cacheExpirationUnauthorized)
+				abort.Abort(c, "Unauthorized", http.StatusUnauthorized)
+				return
+			}
 		}
 	}
 }
--- a/cmd/api/routers/routers.go
+++ b/cmd/api/routers/routers.go
@@ -4,6 +4,7 @@
 package routers
 
 import (
+	"errors"
 	"fmt"
 	"net/http"
 	"os"
@@ -108,6 +109,30 @@ func (c *Controller) GetNamespacedResourceByName(context *gin.Context) {
 	context.JSON(http.StatusOK, resource)
 }
 
+func (c *Controller) GetLogURLsByResourceName(context *gin.Context) {
+	group := context.Param("group")
+	version := context.Param("version")
+	apiVersion := fmt.Sprintf("%s/%s", group, version)
+	namespace := context.Param("namespace")
+	name := context.Param("name")
+
+	kind, err := discovery.GetAPIResourceKind(context)
+	if err != nil {
+		abort.Abort(context, err.Error(), http.StatusInternalServerError)
+		return
+	}
+	logURLs, err := c.Database.QueryLogURLs(context.Request.Context(), kind, apiVersion, namespace, name)
+	if errors.Is(err, database.ResourceNotFoundError) {
+		abort.Abort(context, err.Error(), http.StatusNotFound)
+	}
+	if err != nil {
+		abort.Abort(context, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	context.JSON(http.StatusOK, logURLs)
+}
+
 func (c *Controller) GetAllCoreResources(context *gin.Context) {
 	version := context.Param("version")
 	limit, id, date := pagination.GetValuesFromContext(context)
@@ -167,6 +192,30 @@ func (c *Controller) GetNamespacedCoreResourceByName(context *gin.Context) {
 	}
 
 	context.JSON(http.StatusOK, resource)
+}
+
+func (c *Controller) GetLogURLsByCoreResourceName(context *gin.Context) {
+	version := context.Param("version")
+	namespace := context.Param("namespace")
+	name := context.Param("name")
+
+	kind, err := discovery.GetAPIResourceKind(context)
+	if err != nil {
+		abort.Abort(context, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	logURLs, err := c.Database.QueryLogURLs(context.Request.Context(), kind, version, namespace, name)
+	if errors.Is(err, database.ResourceNotFoundError) {
+		abort.Abort(context, err.Error(), http.StatusNotFound)
+	}
+	if err != nil {
+		abort.Abort(context, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	context.JSON(http.StatusOK, logURLs)
+
 }
 
 // Livez returns current server configuration as we don't have a clear deadlock indicator
--- a/pkg/database/database.go
+++ b/pkg/database/database.go
@@ -6,11 +6,14 @@ package database
 import (
 	"context"
 	"database/sql"
+	"errors"
 	"fmt"
 
 	"github.com/kubearchive/kubearchive/pkg/models"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
+
+var ResourceNotFoundError = errors.New("resource not found")
 
 type newDatabaseFunc func(map[string]string) DBInterface
 
@@ -20,6 +23,8 @@ type DBInfoInterface interface {
 	GetDriverName() string
 	GetConnectionString() string
 
+	GetUUIDSQL() string
+
 	GetResourcesLimitedSQL() string
 	GetResourcesLimitedContinueSQL() string
 
@@ -28,15 +33,26 @@ type DBInfoInterface interface {
 
 	GetNamespacedResourceByNameSQL() string
 
+	GetLogURLsByPodNameSQL() string
+	GetOwnedResourcesSQL() string
+	GetLogURLsSQL() string
+
 	GetWriteResourceSQL() string
 	GetWriteUrlSQL() string
 	GetDeleteUrlsSQL() string
+}
+
+// DBParamParser need to be implemented by every database driver compatible with KubeArchive
+type DBParamParser interface {
+	// ParseParams transform the given args to accepted parameters for parametrized queries
+	ParseParams(query string, args ...any) (string, []any, error)
 }
 
 type DBInterface interface {
 	QueryResources(ctx context.Context, kind, apiVersion, limit, continueId, continueDate string) ([]*unstructured.Unstructured, int64, string, error)
 	QueryNamespacedResources(ctx context.Context, kind, apiVersion, namespace, limit, continueId, continueDate string) ([]*unstructured.Unstructured, int64, string, error)
 	QueryNamespacedResourceByName(ctx context.Context, kind, apiVersion, namespace, name string) (*unstructured.Unstructured, error)
+	QueryLogURLs(ctx context.Context, kind, apiVersion, namespace, name string) ([]string, error)
 
 	WriteResource(ctx context.Context, k8sObj *unstructured.Unstructured, data []byte) error
 	WriteUrls(ctx context.Context, k8sObj *unstructured.Unstructured, logs ...models.LogTuple) error
@@ -45,8 +61,9 @@ type DBInterface interface {
 }
 
 type Database struct {
-	db   *sql.DB
-	info DBInfoInterface
+	db          *sql.DB
+	info        DBInfoInterface
+	paramParser DBParamParser
 }
 
 func NewDatabase() (DBInterface, error) {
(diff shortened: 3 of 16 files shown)


pkg/database/database.go
```diff
@@ -107,39 +114,71 @@ func (db *Database) QueryNamespacedResourceByName(ctx context.Context, kind, api
...
+		uuids = make([]string, 0)
+		for _, row := range parsedRows {
+			if row.Kind == "Pod" {
+				podUuids = append(podUuids, row.Uuid)
+			} else {
+				uuids = append(uuids, row.Uuid)
+			}
+		}
+		i = len(uuids)
+	}
```
skoved, line 150: I think the looping logic could be made more clear by having it all in the `for` line

```suggestion
	for i:= len(uuids); i > 0; i = len(uuids) {
		parsedRows, err := newQueryPerformer[uuidKind](ctx, db.db).performQueryWithArray(db.info.GetOwnedResourcesSQL(), uuids)
		if err != nil {
			return nil, err
		}
		uuids = make([]string, 0)
		for _, row := range parsedRows {
			if row.Kind == "Pod" {
				podUuids = append(podUuids, row.Uuid)
			} else {
				uuids = append(uuids, row.Uuid)
			}
		}
	}
```

pkg/database/query.go
skoved, line 1: would using [`sqlx`](https://pkg.go.dev/github.com/jmoiron/sqlx) make sense here? [`sqlx.SelectContext`](https://pkg.go.dev/github.com/jmoiron/sqlx#SelectContext) seems like it would do what `QueryPerformer` is does and we wouldn't have to maintain the logic ourselves

skoved (changes requested):
The lookups for log urls seem to me like the most complex set of queries KubeArchive has to do so far, so thanks for working on this! There's a couple things I noticed but this otherwise looks good so far

ggallen:
I started to review this and while it probably works it seems overly complicated.  The process of generating the log URLs appears to lend itself to recursion.  I think most of your changes can probably be replaced by something like this:
```
type uuidKind struct {
        Uuid string 
        Kind string
}

func (db *Database) QueryLogURLs(ctx context.Context, kind, apiVersion, namespace, name string) ([]string, error) {
        []string urls 
        string uuid = db.getUUID(ctx, kind, apiVersion, namespace, name)
        return getLogUrls(ctx, { Uuid: uuid, Kind: kind }, urls)
}

func (db *Database) getLogURLs(ctx context.Context, uuid uuidKind, urls []string) ([]string error) {
        if uuid.Kind == 'Pod' {
                return db.getLogURLs(uuid), nil
        } else {
                owners, _ := db.getOwnedResources(ctx, uuid.Uuid) // Return UUID/kind of resources that uuid owns.
                for owner := range owners { 
                        urls = append(urls, getLogURLs(ctx, owner, urls))
                }
                return urls, nil
        }
}
Note that this is pseudo code and has to be cleaned up, but I think it logically does what is needed.

File: cmd/api/routers/routers.go [hunk 2]
```diff
@@ -108,6 +108,27 @@ func (c *Controller) GetNamespacedResourceByName(context *gin.Context) {
 	context.JSON(http.StatusOK, resource)
 }
 
+func (c *Controller) GetLogsURLByResourceName(context *gin.Context) {
```
rh-hemartin, line 111: Just for consistency.

```suggestion
func (c *Controller) GetLogURLsByResourceName(context *gin.Context) {
```

rh-hemartin (approved):
Overall I think this is good. The `queryPerformer` is handy, but we are burying deeper the `db` connection. I'm willing to assume that tradeoff right now. We need to move forward, so I would revisit this after a while and see if and how we can simplify it.

pkg/database/query.go
maruiz93: > On a more general note, why do we have to treat these queries different than any other queries?

Until now we needed to retrieve always the same stuff from all the selects, in other words, all the `SELECT` queries that we had the same fields: `uuid`, `created_at` and `data`

With the LOG_URL retrieval I am introducing new queries that retrieve different things:
1. A query that retrieves the `kind` and `uuid`, for the query that selects all the directly owned resources by a specific uuid
2. A query that retrieves the `url` field

So now we have 3 type of queries with regards to the selected fields, and that's why I introduced this new `QueryPerformer` parametrized struct.

If I instantiate a `QueryPerformer[string]` means that when I perform the query I would receive a slice of strings, if I instantiate a `QueryPerformer[int]` means that when I perform the query I would receive a slice of ints, and if I instantiate a `QueryPerformer[struct]` means that when I perform the query I would receive a slice of instances of that struct.

> would using [sqlx](https://pkg.go.dev/github.com/jmoiron/sqlx) make sense here? [sqlx.SelectContext](https://pkg.go.dev/github.com/jmoiron/sqlx#SelectContext) seems like it would do what QueryPerformer is does and we wouldn't have to maintain the logic ourselves

I think we could take a look at it but I have the feeling that it wouldn't be that simple, specially regarding passing arrays as arguments, but I would rather investigate that in a separate issue. BTW, note that I included a note on top on the "worst" function (the one that does the reflection) of the `query.go` file pointing to [an issue](https://github.com/golang/go/issues/61637) in sql/db library for implementing that logic that is already available in `sqlx` as you mention.

maruiz93:
> I started to review this and while it probably works it seems overly complicated. The process of generating the log URLs appears to lend itself to recursion. I think most of your changes can probably be replaced by something like this:
> 
> ```
> type uuidKind struct {
>         Uuid string 
>         Kind string
> }
> 
> func (db *Database) QueryLogURLs(ctx context.Context, kind, apiVersion, namespace, name string) ([]string, error) {
>         []string urls 
>         string uuid = db.getUUID(ctx, kind, apiVersion, namespace, name)
>         return getLogUrls(ctx, { Uuid: uuid, Kind: kind }, urls)
> }
> 
> func (db *Database) getLogURLs(ctx context.Context, uuid uuidKind, urls []string) ([]string, error) {
>         if uuid.Kind == 'Pod' {
>                 return db.getLogURLs(uuid), nil
>         } else {
>                 owners, _ := db.getOwnedResources(ctx, uuid.Uuid) // Return UUID/kind of resources that uuid owns.
>                 for owner := range owners { 
>                         urls = append(urls, getLogURLs(ctx, owner, urls))
>                 }
>                 return urls, nil
>         }
> }
> ```
> 
> Note that this is pseudo code and has to be cleaned up, but I think it logically does what is needed.

The problem I see with this is that you are assuming that there is only one jump between the resource given and the pods (e.g., a `TaskRun` and a `Pod` or a `ReplicaSet` and a `Pod` or a `Job` and a `Pod`) but you could have several levels (e.g. a `Deployment` owns a `ReplicaSet` that owns `Pod`s or a `CronJob` owns `Job`s that owns `Pod`s) and that's why we need some kind of recursion because the number of jumps could be n.

Also, note that with this proposal I am trying to reduce the number of queries to the minimum:

* If a pod is given just one query is being done
* If something different that a pod is given, then
    * A query is performed to get the uuid of the resource
    * n queries are performed to get the owned resources given a uuid, being n the number of jumps between the resource kind and the owned resources in the "ownership chain" (e.g, for a `Deployment` or a `CronJob` n=2, for a `ReplicaSet` or a `Job` n=1).
    * A query is performed to get the urls given the pod uuids

Considering this the total number of queries performed if the given resource is not a pod is: 2+n and I wouldn't expect to have n greater than 3.

ggallen:
> The problem I see with this is that you are assuming that there is only one jump between the resource given and the pods (e.g., a `TaskRun` and a `Pod` or a `ReplicaSet` and a `Pod` or a `Job` and a `Pod`) but you could have several levels (e.g. a `Deployment` owns a `ReplicaSet` that owns `Pod`s or a `CronJob` owns `Job`s that owns `Pod`s) and that's why we need some kind of recursion because the number of jumps could be n.

No, I'm not making that assumption at all. What I have coded handles all the cases you describe.  And I think it's significantly less code.

> If a pod is given just one query is being done

I think this is the only difference in the number of queries, and it's minimal.  It's only more in the case where you query logs for a `Pod` directly.  I would actually suggest "hiding" the `GetUUID` call behind the other calls. There's really no need to expose it to the users.

pkg/database/query.go
ggallen: So are we planning to refactor all the other queries to use `QueryPerformer`?

ggallen:
So here's a code version using recursion and your code. I haven't compiled it, but it should work and I think it's simpler than your loop.
```
type uuidKind struct {
	Uuid string `json:"uuid"`
	Kind string `json:"kind"`
}

func (db *Database) QueryLogURLs(ctx context.Context, kind, apiVersion, namespace, name string) ([]string, error) {
        []string urls
	err := db.db.QueryRowContext(ctx, db.info.GetUUIDSQL(), kind, apiVersion, namespace, name).Scan(&uuid)
	if err != nil {
		return nil, err
	}
        return queryLogURLs(ctx context.Context, {Uuid: uuid, Kind: kind }, urls)
}

func (db *Database) queryLogURLs(ctx context.Context, uuid uuidKind, []string urls) ([]string, error) {
	if uuid.Kind == 'Pod' {
	    return strQueryPerformer.performQuery(db.info.GetLogURLsSQL(), uuid.Uuid)
        } else {
		parsedRows, err := newQueryPerformer[uuidKind](ctx, db.db).performQuery(db.info.GetOwnedResourcesSQL(), uuid.Uuid)
		if err != nil {
			return nil, err
		}
                for _, row := range parsedRows {
			urls = append(urls, queryLogURLs(ctx, row, urls))
                }
                return urls, nil
        }
}
```
I note that you are correct that there will be more queries here as I am querying the log URLs for each `Pod` while you are getting them all in one call at the end.

But this recursion could easily be modified to simply get the `Pod` UUIDs instead and then make the same call at the end to get the log URLs.

maruiz93:
Still working on the PR for the test stuff, but I addressed some suggestions and also dealt with the arrays for the parametrized queries, a thing that is tightly coupled to the driver used. I created a new interface called `ParamParser` that the different driver database should implement for specifying how to deal with parameters in the form of array (e.g. for the `select ... IN (<array>)`)

Please take a look :)

pkg/database/query.go
```diff
@@ -0,0 +1,138 @@
...
+	"fmt"
+	"reflect"
+	"strings"
+
+	"github.com/lib/pq"
+)
+
+type queryPerformer[T any] struct {
+	zeroVal T
+	ctx     context.Context
```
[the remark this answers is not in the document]
maruiz93: Very good point that I wasn't aware of. I've already changed it.

pkg/database/query.go
```diff
@@ -0,0 +1,138 @@
...
+
+func newQueryPerformer[T any](ctx context.Context, db *sql.DB) queryPerformer[T] {
+	var zeroVal T
+	return queryPerformer[T]{zeroVal, ctx, db}
+}
+
+func (q queryPerformer[T]) performQueryWithArray(query string, args []string) ([]T, error) {
+	// FIXME This need to be compatible with other drivers not just with postgres
+	return q.parseRows(q.db.QueryContext(q.ctx, query, pq.Array(args)))
+}
```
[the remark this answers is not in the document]
maruiz93: Done

pkg/database/database.go
maruiz93: great point, as I was thinking of a "while" translation I didn't thought of adding the initialize and the "post-loop" change on the `for`.

pkg/database/mariadb.go
```diff
@@ -49,6 +49,23 @@ func (info MariaDBDatabaseInfo) GetNamespacedResourceByNameSQL() string {
 	return "SELECT JSON_VALUE(data, '$.metadata.creationTimestamp'), uuid, data FROM resource WHERE kind=? AND api_version=? AND namespace=? AND name=?"
 }
 
+func (info MariaDBDatabaseInfo) GetUUIDSQL() string {
+	return "SELECT uuid FROM resource WHERE kind=?AND api_version=? AND namespace=? AND name=?"
```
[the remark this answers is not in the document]
maruiz93: good catch! Done!

cmd/api/routers/routers.go
maruiz93: done! and changed in both new entrypoint methods

pkg/database/database.go
```diff
@@ -107,39 +127,81 @@ func (db *Database) QueryNamespacedResourceByName(ctx context.Context, kind, api
...
+		return nil, err
+	}
+	uuids = append(uuids, uuid)
+	var podUuids []string
+	type uuidKind struct {
+		Uuid string `json:"uuid"`
+		Kind string `json:"kind"`
+	}
+	// This loops for each jump existing in the ownerReference chain from the first resource until the pod
+	// The slice uuids is modified inside the loop
```
ggallen, line 151: Loop counters should generally not be modified within the loop. This tells me something is wrong.
```suggestion
	pq := db.newParamQuery(db.info.GetLogURLsSQL())
	pq.addStringArrayParam(podUuids)
	return strQueryPerformer.performQuery(ctx, pq)
```

maruiz93: ok, so look at it in a different way, note that golang doesn't have the `while` reserve word, `while` is performed with the `for`. So, I am doing things `while` I still have `uuid`s of  resources different of a `Pod`, but I am not iterating over an array/slice. In other words, I am not modifying the counter of a loop, I am doing things while I still have uuids, and I am removing uuids in the process and adding new ones

pkg/database/query.go
```diff
@@ -0,0 +1,138 @@
...
+
+func (q queryPerformer[T]) performQuery(query string, args ...string) ([]T, error) {
+	return q.parseRows(q.db.QueryContext(q.ctx, query, toSliceOfAny(args)...))
+}
```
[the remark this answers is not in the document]
maruiz93: done

maruiz93:
I still need to add new tests but I want to see the output of the CI

pkg/database/database.go
ggallen: You seem intent on ignoring my suggestion to use recursion to solve this problem, and I'm not sure why.

The fact that what you are trying to do isn't supported easily in golang indicates to me that you should consider other options.  What I am proposing works and I think is far more readable. I have been looking at this code for the past couple of days and I find it very confusing.  The fact that you have to explain in such detail also indicates that it's either overly complex or just unclear.

You seem to be focused on making the fewest number of calls possible, which is an admirable goal. But to me the tradeoff with the complexity and the lack of readability of the code makes it not worth it. Here are two examples I can cite:

1. The first call to special-case if the request is for a `Pod`.  Yeah, you can do that and you make one less call. But to me a more general solution that just always gets the uuid and then goes from there is easier to read and more concise. Everything is handled the same.  I'm not sure the tradeoff is worth it for the number of times logs will be requested in general versus the number of times logs are requested specifically for a `Pod`.
2. Batching uuids together for calls. I understand this a little more, but again I'm not sure it's worth the tradeoff. It makes no difference in your implementation, and it's clear why you are doing it. But I can tell you I didn't notice this batching the first N times I read the code, and as  a result I thought the code was broken, which again points to readability. Again, I question whether this is worth it, but recursion can also support doing it that way as well.

So I will meet you halfway and give you #2 but not number #1.  Here's yet another solution with that in mind:

```suggestion
type uuidKind struct {
	Uuid string `json:"uuid"`
	Kind string `json:"kind"`
}

func (db *Database) QueryLogURLs(ctx context.Context, kind, apiVersion, namespace, name string) ([]string, error) {
        []string PodUuids
        string owner
	err := db.db.QueryRowContext(ctx, db.info.GetUUIDSQL(), kind, apiVersion, namespace, name).Scan(&owner)
	if err != nil {
		return nil, err
	}
        podUuids, err := getPodUUIDs(ctx, [ owner ], podUuids)
	pq := db.newParamQuery(db.info.GetLogURLsSQL())
	pq.addStringArrayParam(podUuids)
	return strQueryPerformer.performQuery(ctx, pq)
}

func (db *Database) getPodUUIDs(ctx context.Context, []string owners, []string podUuids) ([]string, error) {
	if len(uuids) == 0 {
	    return podUuids, nil
        } else {
		pq := db.newParamQuery(db.info.GetOwnedResourcesSQL())
		pq.addStringArrayParam(owners)
		parsedRows, err := newQueryPerformer[uuidKind](db.db).performQuery(ctx, pq)
		if err != nil {
			return nil, err
		}

                var uuids string[]
                for _, row := range parsedRows {
			if row.Kind == 'Pod' {
				podUuids = append(podUuids, row.Uuid)
			} else {
				uuids = append(uuids, row.Uuid))
			}
                }
                return getPodUuids(ctx, uuids, podUuids)
        }
}
```
I don't like it as well as my previous solution, and I think it's uglier because it does the batching as you were. But I still think it's much more concise and readable that what you have.

> In other words, I am not modifying the counter of a loop, I am doing things while I still have uuids, and I am removing uuids in the process and adding new ones

I see your point here, but I think it's semantics. You aren't actually touching `i`. You are however assigning `i` a value rather than the normal increment.  It looks odd.

As I said, I don't like it, but this is even better IMO. You don't need `i` at all:
```
	for ; len(uuids) > 0;) {
```

maruiz93: > You seem intent on ignoring my suggestion to use recursion to solve this problem, and I'm not sure why.

I am sorry, it wasn't my intent to ignore you and I had a comment written these days that I thought I sent but I didn't press the "Comment" button in Github -.-'
I'll try to address the points that I did there and also what you mention in here

> The fact that what you are trying to do isn't supported easily in golang

I am not sure why you are saying this, it is supported, it's just the way golang handles the `while` of other languages.

> You seem to be focused on making the fewest number of calls possible

Indeed I am. 
Let's say you have a large complex PipelineRun, with lots of TaskRuns, let's say 15. 
If I batch the uuids, I will only do 2 queries to get the owned resources, if I go one by one as you suggested at some point, you will have 16 queries, one per PipelineRun and one per TaskRun to find its pods. 
I really believe we should prioritize optimization in the number of queries for this use case.

> So I will meet you halfway and give you `#2` but not `#1`

I could live with this, even though it makes more sense to me trying to optimize the number of queries in general.

> Regarding recursion

What I mentioned in my lost comment was that I could do recursion while keeping the logic that I have. My idea was actually pretty similar to your current suggestion but I was going to say let's do it in a separate MR.
One other thing that I was considering for the recursion is to make it a [tail-recursion](https://medium.com/@meeusdylan/tail-recursion-in-go-fb5cf69a0f26) as it is more efficient. 
Your previous suggestion weren't tail recursion functions but this one is, so it looks not good to me.

> Regarding readability

I guess that although this is kind of subjective, recursion is usually conceived as a more elegant and readable solution than iteration, and even though I chose iteration for my first approach, I don't have anything against recursion (I actually enjoy designing recursive algorithms while keeping tail-recursion in mind)

> Regarding the usage of `i` in the iteration definition

True, I actually like the `for ; len(uuids) > 0; {` that you suggest, it is cleaner.

> In summary

Now that you took the time to do a suggestion for the tail-recursive function that keeps batch queries, tomorrow I'll extract this part of the code to a recursive function following your suggestion.
I may want to try to keep the happy just-one-query path for the cases where pods are given, which is also very go-stylish as it means an early return in the function. We can discuss it tomorrow.

ggallen: > True, I actually like the for ; len(uuids) > 0; { that you suggest, it is cleaner.

Actually, I just looked again and you can drop the semi-colons:

````
for len(uuids) > 0 {
        ...
}
```

ggallen: If you *really* want to optimize, you could consider using a stored procedure here instead. Although stored procedures and individual calls perform about the same, the risk of a network issue during the calls is significantly higher for individual calls rather than the one stored procedure call.   :-)

I am *not* proposing this change, just pointing out that it might actually be more performant.

maruiz93: hmm, I am not against it if the performance isn't enough, but at first I believe that being able to receive intermediate results in the API may end up in delivering more information to the user of potential problems.

For now, I updated this logic from iteration to recursion as you suggested

maruiz93:
Added unit tests and the extra sar request for the `pod/logs` resource

cmd/api/auth/authz.go
```diff
@@ -36,44 +37,59 @@ func RBACAuthorization(sari clientAuthzv1.SubjectAccessReviewInterface, cache *c
...
+					Verb:      verb,
+				},
+			}}
 
-		allowed := cache.Get(sarSpec.String())
-		if allowed != nil {
-			if allowed != true {
-				abort.Abort(c, "Unauthorized", http.StatusUnauthorized)
-				return
+		if strings.HasSuffix(c.Request.RequestURI, "log") {
```
rh-hemartin, line 54:
```suggestion
		if strings.HasSuffix(c.Request.URL.Path, "/log") {
```

The `http.Request` docs recommend not using `.RequestURI` but `.URL` instead. Anyway, I also suggest you add a leading `/` so this don't trigger on resources that end with `log`. I don't think there will be many, but it can happen.

pkg/database/query.go
```diff
@@ -0,0 +1,154 @@
...
+	"database/sql"
+	"errors"
+	"fmt"
+	"reflect"
+	"strings"
+)
+
+type paramQuery struct {
+	query         string
+	arguments     []interface{}
```
rh-hemartin, line 17: You can use `any`, which sometimes is "plain english" than `interface{}`:


```suggestion
	arguments     []any
```

However it is not a big deal, at the end of the day both are used widely.

rh-hemartin (approved):

cmd/api/auth/authz.go
maruiz93: good point! Changes applied

pkg/database/query.go
maruiz93: I changed my mind several times of using the alias over the native name or the other way around but at the end I have both and I see that are using the alias in other parts of the code so I'll change to the alias `any` for consistency.

pkg/database/query.go
skoved: > I think we could take a look at it but I have the feeling that it wouldn't be that simple, specially regarding passing arrays as arguments, but I would rather investigate that in a separate issue

That's ok with me. My understanding is that the array that you pass in is supposed to be empty and gets populated with the data from each row of the query. I'll open an issue to investigate using sqlx

skoved (approved):

PR #612, opened 2024-11-04
Unify labels for KubeArchive components

### LINKED ISSUE #499: Unify labels for the different KubeArchive components
We aren't including any `labels` in the api-server or sink resources and we are including several for the operator:
* `app.kubernetes.io/name`
* `app.kubernetes.io/instance`
* `app.kubernetes.io/component`
* `app.kubernetes.io/created-by`
* `app.kubernetes.io/part-of`

Following the [recommended labels docs](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/) I believe we should unify all the YAML k8s resource definition to include the following:

* `app.kubernetes.io/name` : the same as in `metadata.name`
* `app.kubernetes.io/component`: one of the following: `api-server` | `sink` | `operator`
* `app.kubernetes.io/part-of`: `kubearchive`

Meaning, we won't have `app.kubernetes.io/instance` because that one should be unique nor `app.kubernetes.io/created-by` which is now [deprecated](https://kubernetes.io/docs/reference/labels-annotations-taints/#app-kubernetes-io-created-by-deprecated).

It would also be nice to have `app.kubernetes.io/version` with the value extracted from the file `VERSION` created by `release.sh`

mafh314:

Resolves #499 

## Release Note

```release-note
NONE
```

### DIFF
--- a/charts/kubearchive/templates/api_server/api_server.yaml
+++ b/charts/kubearchive/templates/api_server/api_server.yaml
@@ -2,8 +2,13 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: {{ tpl .Values.apiServer.name . }}
+  name: "{{ tpl .Values.apiServer.name . }}"
   namespace: {{ .Release.Namespace }}
+  labels:
+    app.kubernetes.io/name: "{{ tpl .Values.apiServer.name . }}"
+    app.kubernetes.io/component: api-server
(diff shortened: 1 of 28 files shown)


rh-hemartin (changes requested):
This PR needs to change the `hack/release.sh` script as well.

maruiz93 (approved):

rh-hemartin (approved):

PR #637, opened 2024-11-12
Add log url retrieval integration test

### LINKED ISSUE #630: Add log url retrieval integration tests
Add integration tests for querying the `/log` endpoints of pods and other resources

maruiz93:

Resolves #630

## Release Note

```release-note
NONE
```

## Notes for Reviewers

I've added retries on the log url retrieval because if there aren't, even though the cronjob are archived the pods could not be so the log urls could be missing.

### DIFF
--- a/test/integration/integration_test.go
+++ b/test/integration/integration_test.go
@@ -28,6 +28,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"knative.dev/eventing/test/lib/resources"
 )
 
 func TestMain(m *testing.M) {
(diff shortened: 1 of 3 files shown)


test/integration/integration_test.go
```diff
@@ -316,6 +316,47 @@ func TestArchiveAndRead(t *testing.T) {
...
+			fmt.Printf("Couldn't read Body, %s\n", errReadAll)
+			return errReadAll
+		}
+
+		if response.StatusCode != http.StatusOK {
+			fmt.Printf("The HTTP status returned is not OK, %s\n", response.Status)
+			return errors.New(response.Status)
+		}
+		var logUrls []string
+		json.Unmarshal([]byte(body), &logUrls)
```
mafh314, line 349: Hey @maruiz93 I think that the `json.Unmarshal` could return an error that we should manage. Just like in line 305.

test/integration/integration_test.go
```diff
@@ -316,6 +316,47 @@ func TestArchiveAndRead(t *testing.T) {
...
+		if len(logUrls) < 1 {
+			return errors.New("could not retrieve the cronjob pod log urls")
+		}
```
rh-hemartin, line 352: This is not checking properly what it is suposed to be. I'm printing the logUrls and its an empty slice. I guess that there is an error and its not handled, as Andrea suggested.

rh-hemartin: One solution I got:

```suggestion
		type Response struct {
			URLs []string `json:"-"`
		}
		var data Response
		err := json.Unmarshal(body, &data)
		if err != nil {
			return err
		}
		if len(data.URLs) < 1 {
			return ...
	    }
```

rh-hemartin: You can use an anonymous struct instead:

```
logUrls := struct{ URLs []string }{}
```

I'm not sure why `[]string` is not working as expected.

maruiz93: the urls are empty because they are not configured so the urls in the log_url tables are empty... I could add some configuration to also test how the sink introduce the urls but it's not what I did

rh-hemartin: Hm... then just ensure that the endpoint does not return 404. Now its misleading, and its not properly coded. It tells "if there are less than one log result then fail, if there are one or more then the test is successful"

ggallen: How could you ever have a negative number here?  Shouldn't it just be `== 0`?

maruiz93: > How could you ever have a negative number here? Shouldn't it just be == 0?

yes, I will change it, I just didn't put too much thought on it, but I agree it's more readable `== 0`  as "if it's empty" than `< 1` as "it should have at least one"

maruiz93: @ggallen I've made the change to `==0`

@rh-hemartin I've included the config of logging to the sink, so now the urls aren't empty

test/integration/integration_test.go
maruiz93: Done!

rh-hemartin (approved):

mafh314 (approved):

PR #639, opened 2024-11-12
enable additional checks in `golangci-lint`

### LINKED ISSUE #638: enable useful govet analyzers in golangci-lint
`golangci-lint` enables `govet` by default. Right now we are just using the default checks, but I think the following checks would be useful:

- loopclosure - Check references to loop variables from within nested functions
- nilness - Check for redundant or impossible nil comparisons
- shadow - Check for possible unintended shadowing of variables

skoved:

Resolves #638 

## Release Note

```release-note
NONE
```

## Notes for Reviewers

The following `govet` analyzers are now enabled:
- the default set of `govet` analyzers (were already enabled before this PR)
- loopclosure: Check references to loop variables from within nested functions
- nilness: Check for redundant or impossible nil comparisons
- shadow: Check for possible unintended shadowing of variables

The following linters are now enabled:
- nilnil: Checks that there is no simultaneous return of nil error and an invalid value

Made some fixes so that all `golangci-lint` linters pass

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -11,6 +11,15 @@ linters:
     - goimports
     - gosec
     - nilerr
+    - nilnil
     - usestdlibvars
     - wastedassign
     - unparam
+linters-settings:
+  govet:
+    enable-all: false
+    disable-all: false
+    enable:
+      - loopclosure
+      - nilness
+      - shadow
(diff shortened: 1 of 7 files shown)


maruiz93 (approved):

rh-hemartin (approved):

skoved:
rebasing from `main` to allow this to get merged

PR #640, opened 2024-11-12
Add sqlx for structscan support

### LINKED ISSUE #634: Investigate Replacing `db.QueryPerformer` with `sqlx.SelectContext`
[`sqlx.SelectContext`](https://pkg.go.dev/github.com/jmoiron/sqlx#SelectContext) will populate a slice of any type based on a `SELECT` sql statement and attempt to convert each row of the database into a struct. From what I can see, the type of the slice needs to be compatible with the fields that you are returning from the database. The [`usage`](https://pkg.go.dev/github.com/jmoiron/sqlx#readme-usage) section has an example using `sqlx.Select` with a Person struct.

maruiz93:

Resolves #634 

## Release Note

```release-note
NONE
```

## Notes for Reviewers

I kept the queryPerformer struct because of the variadic args handled with the `DBParamParser` interface of the `paramQuery` struct.

I also used string instead of `sql.RawBytes` (an alias of `[]byte`) because its usage is discouraged given that some drivers override its value after the `rows.Next()` execution.
More info about this on:
* https://github.com/DATA-DOG/go-sqlmock/issues/176
* https://jmoiron.github.io/sqlx/#query

### DIFF
--- a/pkg/database/database.go
+++ b/pkg/database/database.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/jmoiron/sqlx"
 	"github.com/kubearchive/kubearchive/pkg/models"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
@@ -61,7 +62,7 @@ type DBInterface interface {
 }
 
 type Database struct {
-	db          *sql.DB
+	db          *sqlx.DB
 	info        DBInfoInterface
 	paramParser DBParamParser
 }
@@ -140,7 +141,7 @@ func (db *Database) QueryLogURLs(ctx context.Context, kind, apiVersion, namespac
 	}
 
 	var uuid string
-	err := db.db.QueryRowContext(ctx, db.info.GetUUIDSQL(), kind, apiVersion, namespace, name).Scan(&uuid)
+	err := db.db.GetContext(ctx, &uuid, db.info.GetUUIDSQL(), kind, apiVersion, namespace, name)
 
 	if errors.Is(err, sql.ErrNoRows) {
 		return []string{}, ResourceNotFoundError
@@ -163,8 +164,8 @@ func (db *Database) QueryLogURLs(ctx context.Context, kind, apiVersion, namespac
 func (db *Database) getOwnedPodsUuids(ctx context.Context, ownersUuids, podUuids []string) ([]string, error) {
 
 	type uuidKind struct {
-		Uuid string `json:"uuid"`
-		Kind string `json:"kind"`
+		Uuid string
+		Kind string
 	}
 
 	if len(ownersUuids) == 0 {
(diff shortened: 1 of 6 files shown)


pkg/database/database.go
```diff
@@ -190,9 +191,9 @@ func (db *Database) getOwnedPodsUuids(ctx context.Context, ownersUuids, podUuids
 
 func (db *Database) performResourceQuery(ctx context.Context, pq *paramQuery) ([]*unstructured.Unstructured, int64, string, error) {
 	type resourceFields struct {
-		Date  string       `json:"created_at"`
-		Id    int64        `json:"id"`
-		Bytes sql.RawBytes `json:"data"`
+		Date     string `db:"created_at"`
+		Id       int64  `db:"id"`
+		Resource string `db:"data"`
```
skoved, line 196: Since `Resource` needs to be converted to `[]byte` anyway I think would make sense to just get `Resource` as `[]byte`

```suggestion
		Resource []byte `db:"data"`
```

pkg/database/database.go
```diff
@@ -209,7 +210,7 @@ func (db *Database) performResourceQuery(ctx context.Context, pq *paramQuery) ([
 	lastRow := parsedRows[len(parsedRows)-1]
 
 	for _, parsedRow := range parsedRows {
-		r, err := models.UnstructuredFromByteSlice([]byte(parsedRow.Bytes))
+		r, err := models.UnstructuredFromByteSlice([]byte(parsedRow.Resource))
```
skoved, line 213: If the type of `Resource` is []byte no conversion is needed

```suggestion
		r, err := models.UnstructuredFromByteSlice(parsedRow.Resource)
```

rh-hemartin (approved):
I guess this is transparent for OpenTelemetry, check if the database continue to emit metrics and traces please.

pkg/database/database.go
maruiz93: So, there is an issue regarding the retrieval of bytes because they are kind of "managed" by the driver and sometimes could be overwritten from row to row retrieval depending on the driver.

Please take a look at the links I posted on "Notes for Reviewers".

skoved: my understanding is that the sql driver treats `[]bytes` and `RawBytes` differently. From the docs for `row.Scan`:

> If a dest argument has type *[]byte, Scan saves in that argument a copy of the corresponding data. The copy is owned by the caller and can be modified and held indefinitely

maruiz93: oh! then let me check it out then! I thought that they were the same as `RawBytes` is an alias of `[]bytes` thanks for pointing it out :)

maruiz93: right! it works! Updated :)

pkg/database/database.go
maruiz93: updated!

skoved (approved):

maruiz93:
> > I guess this is transparent for OpenTelemetry, check if the database continue to emit metrics and traces please.
> 
> I didn't check it but it is transparent, sqlx db struct is a wrapper of the sql db and I extracted how to do it from the `otelsql` library docs: https://github.com/nhatthm/otelsql?tab=readme-ov-file#jmoironsqlx
> 
> I will do the test though

tested

image

pkg/database/database.go
skoved: I've been confused by this too

```
type RawBytes []byte // type declaration. RawBytes and []byte are treated as different types and explicit conversion is required between the two types

type RawBytes = []byte // type alias. RawBytes is just another way to refer to []byte. No conversion is required because RawBytes is []byte
```

[This blog post](https://go.dev/blog/alias-names) from the Go devs help me understand the difference better

```
type RawBy

PR #650, opened 2024-11-15
Split filters of SQL queries and add a builder

### LINKED ISSUE #483: Refactor database integrations.
From @maruiz93.

I was just playing a little more with the code to see if I could encapsulate a little more the responsibilities of each "part" of the database integration, and I came up with this :

https://github.com/maruiz93/kubearchive/tree/issue-453-manon-2

With that approach I have 3 elements for each database integration:

- A DBCreator that depends on the environment and implements GetDriverName and GetConnectionString
- A  DBStatements that implements the different SQL statements
- A DBClient that implements the logic to interact with the DB for each business logic action and depends on a DBStatements and a sql.DB

Each one is registered in the init. So the NewDatabase go through the following process:

1. Get the environment
2. Retrieve the creator initializer from the registered ones
3. Instantiate the creator passing the environment
4. Create a connection with the DBCreator
5. Retrieve the statements from the registered ones
6. Retrieve the db client initializer from the registered ones
7. Instantiate the db client passing the connection and the statements

This is the way I managed to get the most "naive" constructor I can and left the less logic that I could for the integration piece.

### LINKED ISSUE #649: Create a query builder
We are currently exposing the complete SQL statement for each query but this is not very maintainable when we start adding filters that are applicable to several queries (like labels or annotations)

### Acceptance Criteria

* Separate the SQL statements for `SELECT` queries into selector and filters and create a builder to build the query with the different pieces

maruiz93:

Resolves #483, resolves #649

## Release Note

```release-note
NONE
```


pkg/database/mariadb.go
```diff
@@ -10,75 +10,116 @@ import (
...
 
 	_ "github.com/go-sql-driver/mysql"
+	"github.com/jmoiron/sqlx"
 	"github.com/kubearchive/kubearchive/pkg/models"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
 
 func init() {
 	RegisteredDatabases["mariadb"] = NewMariaDBDatabase
+	RegisteredDBCreators["postgresql"] = NewMariaDBCreator
```
skoved, line 20: I think you meant to change the key to `mariadb`
```suggestion
	RegisteredDBCreators["mariadb"] = NewMariaDBCreator
```

maruiz93:
Superseded by #663

Issue #240, opened 2024-07-23
Sink should receive cloudevent messages that it missed if it restarts


skoved:
If the sink crashes and restarts, it should receive any cloudevents that were sent while it was unavailable. This will require KubeArchive to have either a Broker or Channel between the ApiServerSource and the sink before this can be worked on.

Triggers and Subscriptions support a `DeadLetterSink` which is a sink to send cloudevents to when a clouldevent could not be delivered to the `Sink`. We need to figure out how to route messages using the `DeadLetterSink` when the sink becomes ready again.

skoved:
The knative folks gave me [this article](https://developers.redhat.com/articles/2023/03/08/configuring-knative-broker-apache-kafka) as their recommendation for handling cloudevent delivery failure. To do this, we need KubeArchive to create a Broker to sit in between the ApiServerSource and the sink. Then we need to create a second Broker to act as the `deadLetterSink` for the first Broker. It will receive all events from the primary broker that failed to be delivered.

By default, knative-eventing will use `MTChannelBasedBroker` as the default Broker type and it uses `InMemoryChannel` as the default channel type. As the name implies, the clouldevents for this type of Channel are stored in memory and are ephemeral. This is fine for us when we are testing and doing development, but is not recommended for production use cases. They recommend instead to use native broker implementations, like the Kafka Broker.

I will create the following PRs as the initial steps for this issue

- [ ] update the kubearchive deployment so that a Broker is created in the `kubearchive` namespace and update the operator code to configure the ApiServerSource to send events to the broker
- [ ] add documentation with instructions for configuring KubeArchive to use be able to use a Kafka Broker

skoved:
The next thing to do is figure out how to get the cloudevents from the `deadLetterSink` broker back to the sink, when the sink becomes available again

rh-hemartin:
> The next thing to do is figure out how to get the cloudevents from the deadLetterSink broker back to the sink, when the sink becomes available again

Shouldn't this happen automatically? The dead letter sink broker has a delivery section with the same information as the normal broker, so I expect it to send the messages automatically with an exponential backoff, retry 4 times, etc.

rh-hemartin:
To me this looks like an integration, like a database. If providing a memory broker is not recommended for production I think we shouldn't provide it then request users change it. I would not provide it directly as set it as a requirement for the user.

skoved:
[...]

PR #656, opened 2024-11-18
add instructions for configuring kafka brokers

skoved:

Related to #240 

## Release Note

```release-note
NONE
```

### DIFF
--- a/docs/modules/ROOT/pages/getting-started/installation.adoc
+++ b/docs/modules/ROOT/pages/getting-started/installation.adoc
@@ -30,6 +30,50 @@ psql -h database.example.com -U username ... -f kubearchive.sql
 This file creates a PostgreSQL schema, a table for KubeArchive and other stuff KubeArchive
 requires.
 
+== Configure Knative Eventing
+
+By default, Knative Eventing use `MTChannelBasedBroker` as the class for any Broker that does not explicitly define its
+class and uses `InMemoryChannel` as the backing channel implementation. Knative Eventing strongly recommends using a
+native broker implementation for production workloads instead. We recommend the Apache Kafka Broker.
+
+. If you need a Kafka cluster, follow the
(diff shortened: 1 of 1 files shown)


docs/modules/ROOT/pages/getting-started/installation.adoc
```diff
@@ -30,6 +30,50 @@ psql -h database.example.com -U username ... -f kubearchive.sql
...
+        brokerClass: Kafka
+        apiVersion: v1
+        kind: ConfigMap
+        name: kafka-broker-config
+        namespace: knative-eventing
+EOF
+----
+
+Knative Eventing supports other Broker types as well. You can learn more about them
+link:https://knative.dev/docs/eventing/brokers/broker-types/[here]
```
rh-hemartin, line 75: I think this is too much information. If you are willing to reduce to information then I would go with:

```suggestion
KubeArchive uses a memory based  broker to avoid missing Kubernetes resource updates. If this broker is not performant enough for your use case, check
link:https://knative.dev/docs/eventing/brokers/broker-types/[Knative's Eventing supported broker types].
We recommend you use the Kafka broker.
```

If the change is more complicated that this I think it needs its own page then.

ggallen: Let's not recommend anything.  We can say we use a Kafka broker, I suppose.

I agree with less information here, but I think we can say we use the defaul `MTChannelBasedBroker` but that Knative recommends one of the other brokers in production.

skoved: I shortened this section mostly following @rh-hemartin suggestion but moved in the bit about other broker types into an admonition and followed @ggallen suggestion about not recommending anything and instead worded it as knative eventing recommends

rh-hemartin (approved):

maruiz93 (approved):

skoved:
pulled latest commits from main so I can merge

PR #658, opened 2024-11-18
fix how operator determines sink name

### LINKED ISSUE #589: Fix service name on ApiServerSource
When KubeArchive creates the ApiServerSource, reacting to a KubeArchiveConfig creation, the service name set as a sink on the ApiServerSource is wrong. KubeArchive assumes the service name starts with the same word as the namespace its running on, so if KubeArchive is installed on `mynamespace` the Service assigned as sink in the ApiServerSource is named `mynamespace-sink`. The sink Service name is independent from the namespace is located at, so its always `kubearchive-sink`.

**Acceptance Criteria**

* The APIServerSource created points to the service called `kubearchive-sink` instead of `<namespace>-sink`.

**Notes**

* While we discussed that KubeArchive should be installed on the namespace `kubearchive` and we agreed, there are some clusters that don't allow namespace that start with `kube` so users are forced to change the namespace where KubeArchive is installed.

skoved:

Resolves #589 

## Release Note

```release-note
fix bug where operator uses the wrong service name for the sink when deployed in a namespace that is not kubearchive
```

## Notes for Reviewers

make sure that the operator will use the name `kubearchive-sink` even if KubeArchive is not deployed in the `kubearchive` namespace

### DIFF
--- a/cmd/operator/internal/controller/kubearchiveconfig_controller.go
+++ b/cmd/operator/internal/controller/kubearchiveconfig_controller.go
@@ -42,8 +42,8 @@ const (
 var (
 	k9eNs         = os.Getenv("KUBEARCHIVE_NAMESPACE")
 	a13eName      = k9eNs + "-a13e"
-	k9eSinkName   = k9eNs + "-sink"
-	k9eBrokerName = k9eNs + "-broker"
+	k9eSinkName   = "kubearchive-sink"
+	k9eBrokerName = "kubearchive-broker"
 )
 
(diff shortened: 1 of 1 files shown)


skoved:
Still working on testing that this works properly

skoved:
tested successfully and it works!

mafh314 (approved):

rh-hemartin (approved):

skoved:
rebasing from main so I can merge

Issue #641, opened 2024-11-13
Merge API handlers


rh-hemartin:
Currently we have eight handlers related with retrieving resources from the database in the API:

```
	apisGroup.GET("/:group/:version/:resourceType", controller.GetAllResources)
	apisGroup.GET("/:group/:version/namespaces/:namespace/:resourceType", controller.GetNamespacedResources)
	apisGroup.GET("/:group/:version/namespaces/:namespace/:resourceType/:name", controller.GetNamespacedResourceByName)
	apisGroup.GET("/:group/:version/namespaces/:namespace/:resourceType/:name/log", controller.GetLogURLsByResourceName)

	apiGroup.GET("/:version/:resourceType", controller.GetAllCoreResources)
	apiGroup.GET("/:version/namespaces/:namespace/:resourceType", controller.GetNamespacedCoreResources)
	apiGroup.GET("/:version/namespaces/:namespace/:resourceType/:name", controller.GetNamespacedCoreResourceByName)
	apiGroup.GET("/:version/namespaces/:namespace/:resourceType/:name/log", controller.GetLogURLsByCoreResourceName)
```

The handlers of the URLS `/group/version/resourceType`, `/version/resourceType`, `/group/version/namespaces/namespace/resourceType` and `/version/namespaces/namespace/resourceType` are very similar. In order to simplify a little bit our code we could aggregate those handlers into one or two.

If we aggregate by database query, we can aggregate into two: resources and namespaced resources, as those use different queries `QueryResources` or `QueryNamespacedResources` resources respectively. This would be accomplished by merging the difference between core and non-core resources, which is the version/apiVersion creation. On the the core endpoints we pass the version as is to the database engine to be used as apiVersion on the DB. On the non-core endpoints we use group and version as apiVersion on the DB. A simple if clause could merge both:

```
        group := context.Param("group")
	version := context.Param("version")
	apiVersion := version
	if group != "" {
		apiVersion = fmt.Sprintf("%s/%s", group, version)
	}
        resources, lastId, lastDate, err = c.Database.QueryResources(context.Request.Context(), kind, apiVersion, limit, id, date)
```

If we aggregate also the namespaced endpoints we would need an if/else clause to use different queries, something like this:

```
        namespace := context.Param("namespace")
	if namespace != "" {
		resources, lastId, lastDate, err = c.Database.QueryNamespacedResources(context.Request.Context(), kind, apiVersion, namespace, limit, id, date)
	} else {
[...]

PR #663, opened 2024-11-20
Split filters of SQL queries and add a builder

### LINKED ISSUE #483: Refactor database integrations.

### LINKED ISSUE #649: Create a query builder

maruiz93:

Resolves #483, resolves #649

## Release Note

```release-note
NONE
```

## Notes for Reviewers

After the discussion on #649 and on the design meeting I included `go-sqlbuilder` which end up being
flexible enough to be used even with DBMS specific functionalities (like JSON support).

In this proposal I am including a Documentation page for the Design Proposals and the goals, so if you disagree with the "design" part more than the "implementation" part, feel free to add reviews on that part.

This PR superseeds #650 in which a custom sqlbuilder was used instead of delegating on a library.

It still applies that with this changes some of the database functions look very similar and they could be merged as part of the effort for #641 .


cmd/api/pagination/pagination.go
```diff
@@ -28,8 +28,10 @@ const (
 // GetValuesFromContext is a helper function for routes to retrieve the
 // information needed. This is kept here, so it's close to the function
 // that sets these values in the context (Middleware)
-func GetValuesFromContext(context *gin.Context) (string, string, string) {
-	return context.GetString(limitKey), context.GetString(continueIdKey), context.GetString(continueDateKey)
+func GetValuesFromContext(context *gin.Context) (int, string, string) {
+	var limit int
+	limit, _ = strconv.Atoi(context.GetString(limitKey))
```
rh-hemartin, line 33: This does make sense as if the limit is present on the context is a valid integer, because we already check on the middleware funciton. However add a comment here explaining why the error is ignored.

maruiz93: I just realized that it would make more sense to just do `context.GetInt()` and store it as an integer. PTAL.

pkg/database/database.go
```diff
@@ -62,19 +37,33 @@ type DBInterface interface {
...
 	env, err := newDatabaseEnvironment()
 	if err != nil {
 		return nil, err
 	}
+
+	var creator facade.DBCreator
+	if c, ok := RegisteredDBCreators[env[DbKindEnvVar]]; ok {
+		creator = c(env)
+	} else {
+		panic(fmt.Sprintf("No database registered with name %s", env[DbKindEnvVar]))
```
rh-hemartin, line 59: We should return `nil, fmt.Errorf("no database registered with name '%s'", env[DbKindEnvVar])` instead.

pkg/database/database.go
```diff
@@ -62,19 +37,33 @@ type DBInterface interface {
...
+
+	conn := establishConnection(creator.GetDriverName(), creator.GetConnectionString())
+
 	var database DBInterface
-	if f, ok := RegisteredDatabases[env[DbKindEnvVar]]; ok {
-		database = f(env)
+	if init, ok := RegisteredDatabases[env[DbKindEnvVar]]; ok {
+		database = init(conn)
 	} else {
 		panic(fmt.Sprintf("No database registered with name %s", env[DbKindEnvVar]))
```
rh-hemartin, line 68: No changes here, but noticed the other part approach. Golang is about returning errors, so panicking here does not look something we want to do.

pkg/database/database.go
```diff
@@ -282,13 +298,12 @@ func (db *Database) WriteUrls(ctx context.Context, k8sObj *unstructured.Unstruct
 	}
 
 	for _, log := range logs {
-		_, execErr := tx.ExecContext(
-			ctx,
-			db.info.GetWriteUrlSQL(),
-			k8sObj.GetUID(),
+		q, qargs := db.Inserter.UrlInserter(
```
rh-hemartin, line 301: I think it makes more sense this way, given we don't want to shadow the parent `query` and `args`.
```suggestion
		logQuery, logArgs := db.Inserter.UrlInserter(
```

rh-hemartin (approved):
Sorry for the late review.

I liked the way the Database was working before, in which it did not know much. Now it needs to know more because it builds the queries itself (before this was delegated to the `DBInfoInterface`.

I tried to implement this without deleting `DBInfoInterface` and having the old methods with a new implementation (building the query instead of returning a hardcoded string) but at the time of query creation it needs the arguments too, so that meant more argument passing here and there.

I think this implementation fits our current needs, but I'm worried about how it will grow when we introduce labelSelector, fieldSelector and ownerSelector. I think that will require another rework to this code, as number of arguments will grow even more and probably will be wise to move to a struct to pass around arguments. However I'm not sure, I sometimes need to implement to know how something will look like.

I'm good with these changes, I don't like them a lot but I don't see any other option forward for now. When our queries grow yet again and we go into a more stable scenario, we will probably find a less layered way to organize this.

maruiz93:
> I'm worried about how it will grow when we introduce labelSelector, fieldSelector and ownerSelector.
I am not sure of following your train of thought. I did this refactor thinking of having less work (and code) when including `labelSelector`, `fieldSelector` and `ownerSelector` (as they are going to be plugged into several queries).

> I liked the way the Database was working before, in which it did not know much. Now it needs to know more because it builds the queries itself (before this was delegated to the `DBInfoInterface`.
Now that the `DBInfoInterface` is chunked in different pieces, I find it easier to manage and we also remove duplication between `postgresql.go` and `mariadb.go`. Also, we had a problem before for variadic arguments that now is solved by the sqlbuilder (e.g. for the `in` operand in filters) that now is solved in a transparent way, and we don't need to know how are variables written for a specific database (`%` for MariaDB and `$1` for PostgreSQL), so I believe that the Database needs to know less, not more.

>  I think that will require another rework to this code, as number of arguments will grow even more and probably will be wise to move to a struct to pass around arguments.
So, the way that I see it is that if the api needs to apply, let's say, a "name filter", it knows that it needs to provide the thing to filter (in this case the `name`) and get in return the "sql piece" to be added to the query, so it makes sense to me having the arguments involved in the specific DBMS part (the old `DBInfoInterface`). 
When you say that the arguments "will grow", they will grow because we will have more filters based on different things, but the amount of arguments for each filter should be small. For example, a `labelSelector` will need a set of key, value that is probably going to be encapsulated as one arg, and the same for fieldSelector and the `ownerSelector`.


However, you can provide more input when I create the PR for adding `labelSelector`.

docs/modules/ROOT/pages/design/database-interface.adoc
```diff
@@ -0,0 +1,105 @@
...
+Encapsulate the different `WHERE` SQL clauses that can be used in `SELECT`, `DELETE` or `UPDATE` statements.
+
+`DBSorter`::
+Encapsulate the different `ORDER BY` clauses.
+
+`DBInserter`::
+Encapsulate the different `INSERT` SQL statements.
+
+`DBDeleter`::
+Encapsulate the different `DELETE` SQL statement without the `WHERE` clauses.
```
ggallen, line 47: Can or should we force `DBDeleter` to have a `WHERE` clause?  It seems dangerous not to.

We might also want to do it for `DBSelector`, but that's not quite as dangerous as `DBDeleter`.

maruiz93: So, the idea is that the SQL statements are separated so the `WHERE` clauses can be plugged in when needed.
Then, the same `WHERE` clauses can be used in a `DBSelector` and in a `DBDeleter` (and `DBUpdater` if we have one in the future). So the idea I had here was the following: using a sqlbuilder doesn't restrict you for writing any type of query, and the responsability of writing good queries rely in the `database.go` functions that specify the different parts that are used in the query.

From my POV, we should have tests for checking that we don't mess things up.

ggallen: My thinking was just having the `DBDeleter` requiring a `WHERE` clause be passed to it, so you could still use the same `WHERE` in multiple places if you like.

But I'm OK with not changing things.

ggallen (approved):

docs/modules/ROOT/pages/design/database-interface.adoc
maruiz93: I see... the problem I see with that is that considering `go-sqlbuilder` architecture to create a `WHERE` clause  you need a sqlbuilder (i.e, a `DBSelector` a `DBDeleter` and a `DBUpdater`) but not the other way around... so I don't know how to accomplish your suggestion without doing complicated stuff

ggallen: Yeah, leave it. I've already approved.

pkg/database/database.go
maruiz93: Updated

PR #665, opened 2024-11-20
Sink mounts kubearchive-logging ConfigMap

### LINKED ISSUE #657: Sink should mount and use kubearchive-logging ConfigMap.
As part of #647, the name of the logging `ConfigMap` for KubeArchive is `kubearchive-logging`. The sink should mount this `ConfigMap` and uses the values contained in it to generate the log URLs.

An empty `kubearchive-logging` `ConfigMap` is added to the KubeArchive installation as part of #647.

**Acceptance Criteria:**

- Sink mounts the `kubearchive-logging` `ConfigMap` as part of its deployment.
- Log URLs are generated only if the `LOG_URL` key is supplied in the `ConfigMap`.

skoved:

Resolves #657 

## Release Note

```release-note
NONE
```


cmd/sink/logs/logs.go
```diff
@@ -7,91 +7,80 @@ import (
...
-	if err != nil {
-		slog.Warn(
-			"error while getting ConfigMap",
-			"ConfigMap",
-			fmt.Sprintf("%s/%s", configCmRef.Namespace, configCmRef.Name),
-			"error",
-			err.Error(),
-		)
-		return emptyCm
+func loggingConfigFromFiles(files map[string]string) (map[string]string, error) {
```
maruiz93, line 42: Can we move this logic to `pkg`? With the recent changes on the logging proposal, the api will also need to load this configmap information

maruiz93 (changes requested):

cmd/sink/logs/logs.go
```diff
@@ -7,91 +7,80 @@ import (
...
+		}
 	}
-	var exists bool
-	loggingCmRef.Name, exists = kaConfigCm.Data[loggingCmSelector.Key]
+	return conf, nil
+}
+
+func getKubeArchiveLoggingConfig() map[string]string {
+	loggingDir, exists := os.LookupEnv(loggingDirEnvVar)
 	if !exists {
```
ggallen, line 57: This should always be set since it required for the deployment.  Why isn't this an error?

cmd/sink/logs/logs.go
```diff
@@ -7,91 +7,80 @@ import (
...
-		)
-		return emptyCm
+		slog.Error("Could not read logging config", "path", loggingDirEnvVar, "error", err.Error())
+		return make(map[string]string)
 	}
-	return loggingCm
+	loggingConf, err := loggingConfigFromFiles(files)
+	if err != nil {
+		slog.Error("Could not get value for logging config", "error", err.Error())
+		return make(map[string]string)
```
ggallen, line 69: If an empty map is returned here, does that mean an attempt will be made to generate URLs?  I would assume we would return nil and no attempt would be made.

cmd/sink/logs/logs.go
skoved: I moved `LoggingConfigFromFiles` and `FilesInDir` into the `pkg/files`

cmd/sink/logs/logs.go
skoved: This now returns an error PTAL

cmd/sink/logs/logs.go
skoved: I've changed how the return so it now returns a `nil` map and an error. PTAL

pkg/files/testdata/filesWithDir/dir/file
skoved, line 1: This file is empty because git ignores directories with no files in them

cmd/sink/logs/logs.go
```diff
@@ -5,93 +5,53 @@ package logs
...
 	logMap map[string]interface{}
 }
 
-func NewUrlBuilder(ctx context.Context, kubeClient kubernetes.Interface) (*UrlBuilder, error) {
-	if kubeClient == nil {
-		return &UrlBuilder{}, fmt.Errorf("Cannot create UrlBuilder with nil kubernetes client")
+func NewUrlBuilder() (*UrlBuilder, error) {
+	loggingConf, err := getKubeArchiveLoggingConfig()
+	if err != nil {
+		return &UrlBuilder{}, err
```
ggallen, line 49: Why return an empty `UrlBuilder`?  This is going to allow calls to `WriteLogs` even when logging isn't configured.  

I think the check to call `Writelogs` should check:
```
if logUrlBuilder != nil &&  strings.ToLower(obj.GetKind()) == "pod" {
```

pkg/files/testdata/filesWithDir/dir/file
maruiz93: we can rename it to `.gitkeep` which I believe is the standard naming for pushing empty directories to git repos

pkg/files/testdata/filesWithDotDot/..one
```diff
@@ -0,0 +1 @@
+ignore
```
maruiz93, line 1: To be more specific with the test you can create this and the other files with `..` prefix as symlinks. If I understand correctly that is the scenario that KubeArchive faces, isn't it?

skoved: it is. I changed `..one`, `..two`, and `..three` to symlinks to their respective files without the `..`

cmd/sink/logs/logs.go
skoved: done

pkg/files/files.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"fmt"
+	"os"
+	"strings"
+)
+
+func FilesInDir(path string) (map[string]string, error) {
+	files := make(map[string]string)
+	dir, err := os.ReadDir(path)
```
ggallen, line 14: On the same theme as my other comments, an empty `ConfigMap` (which is the default) is going to result in an empty map being returned here. This in turn will mean attempts will be made to create log URLs.

I think in this case, if there are file in the `ConfigMap`, then nil should be returned here as well.

Is there a test using the default `ConfigMap` that verifies that no log URLs are generated?  I want to make sure we don't event *attempt* to generate them if the configuration isn't there and correct.

skoved: My bad, I missed that case initially.

I'd prefer returning an empty map instead of `nil` because a nil map cannot be modified. I think that it makes sense to add the logic to prevent log urls from being generated should be put in `cmd/sink/logs/logs.go` since this is functionality specific to the sink. The code in this file will be shared with the Api Server since it will also need to mount the `ConfigMap`. I can add a test to make sure that we don't attempt to add log urls to the database if the ConfigMap is empty

ggallen: > I'd prefer returning an empty map instead of nil because a nil map cannot be modified.

Why would you ever want to modify this map?

And the check to see if the URLs should be generated should be:
```
map != nil && map[LOG_URL] != nil
```

skoved: > Why would you ever want to modify this map?

I don't know and that is the point! This function can be called in other places and we don't know what they will do with it

> And the check to see if the URLs should be generated should be:

There is already a check in `UrlBuilder.Urls` to not generate urls if the map is empty

https://github.com/kubearchive/kubearchive/blob/69a39e802f7d1d9d03054ba103cadc6807963a55/cmd/sink/logs/logs.go#L74


I've added an additional check to see if `map[LOG_URL]` exists

ggallen: >  I think that it makes sense to add the logic to prevent log urls from being generated should be put in cmd/sink/logs/logs.go since this is functionality specific to the sink.

There is already code in `main.go` to handle this, and this is why I am trying to get `sink.logUrlBuilder` to be nil in all these cases.
```
	// only write logs for k8s resources likes pods that have them and UrlBuilder is configured
	if sink.logUrlBuilder != nil && strings.ToLower(obj.GetKind()) == "pod" {
		sink.writeLogs(ctx, obj)
	}
```

skoved: > Why would you ever want to modify this map?

I don't know where this will be called in the future an why since this is now it's own package. Additionally, I don't think it is worth adding extra logic to handle this when `len(map) == 0` catches both nil and empty map

> And the check to see if the URLs should be generated should be:

map != nil && map[LOG_URL] != nil


the check for log url cannot look like that. First of all, the zero value for strings in Go is `""` not `nil`. Secondly `map`'s type is not `map[string]string` it is `map[string]` interface so a type conversion is required. I've updated the logic and added a test to ensure that log urls are not generated if there is no `LOG_URL` field.

pkg/files/testdata/filesWithDir/dir/file
skoved: done

pkg/files/files.go
skoved: I addressed the issues we discussed earlier. PTAL again

maruiz93 (approved):
LGTM, but why `..gitkeep` instead of `.gitkeep`?

skoved:
> LGTM, but why `..gitkeep` instead of `.gitkeep`?
it is part of a test for empty configuration, but the logic won't treat the configuration as empty if the file does not start with `..`

PR #670, opened 2024-11-21
Remove Prometheus metrics from the operator

### LINKED ISSUE #325: Replace kube-rbac-proxy
Currently we are using the container "kube-rbac-proxy" for the operator metrics endpoint. [That container will be removed on March 2025](https://book.kubebuilder.io/reference/metrics). We need to replace it, or remove it if we are not using it.

**Acceptance Criteria**

* Replace (or remove) the mechanism used by the Controller to protect its metrics endpoint.

[1]: https://book.kubebuilder.io/reference/metrics

rh-hemartin:

Resolves #325 

## Release Note

```release-note
NONE
```

## Notes for Reviewers

* Given that we are using OpenTelemetry metrics and not Prometheus metrics, instead of removing the RBAC proxy which is meant to protect the Prometheus endpoint that the operator offered we just remove the whole Prometheus endpoint.


ggallen:
Isn't there another issue dealing with the RBAC proxy?

rh-hemartin:
> Isn't there another issue dealing with the RBAC proxy?

I don't think we have another GitHub issue that references the RBAC proxy.

mafh314 (approved):

maruiz93 (approved):

PR #671, opened 2024-11-21
Add license checker

### LINKED ISSUE #462: Continuos Integration - Check license headers
We are currently applying license headers to the code that is part of this project, mainly bash scripts and go files.

It's easy to miss this when creating a new file so we should have this check in our CI.

There are projects like [license-checker](https://github.com/kt3k/license-checker) that allows configuring what to check in a configuration that can be added the project.

rh-hemartin:

Resolves #462 

## Release Note

```release-note
NONE
```

### DIFF
--- a/.github/workflows/documentation-build.yaml
+++ b/.github/workflows/documentation-build.yaml
@@ -1,3 +1,5 @@
+# Copyright KubeArchive Authors
+# SPDX-License-Identifier: Apache-2.0
 ---
 name: Documentation Build
 run-name: Documentation Build
--- a/.github/workflows/documentation-release.yaml
+++ b/.github/workflows/documentation-release.yaml
@@ -1,3 +1,5 @@
+# Copyright KubeArchive Authors
+# SPDX-License-Identifier: Apache-2.0
 ---
 name: Documentation Release
 run-name: Documentation Release
--- a/.github/workflows/static_code_analysis.yml
+++ b/.github/workflows/static_code_analysis.yml
@@ -54,3 +54,17 @@ jobs:
         with:
           yamllint_strict: false
           yamllint_comment: false
+  license-checker:
+    name: License Checker
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository code
+        uses: actions/checkout@v4
+      - name: Install Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '21'
+      - name: Install License Checker
+        run: npm install --global @kt3k/license-checker
+      - name: Run License Checker
+        run: license_checker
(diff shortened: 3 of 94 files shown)


maruiz93 (approved):

ggallen:
I think we should have a date in our copyright.

maruiz93:
> I think we should have a date in our copyright.

Which kind of date? I see that other popular projects under the same license like bitnami don't put any date: https://github.com/bitnami/charts/blob/main/.licenserc.yaml

ggallen:
> > I think we should have a date in our copyright.
> 
> Which kind of date? I see that other popular projects under the same license like bitnami don't put any date: https://github.com/bitnami/charts/blob/main/.licenserc.yaml

I was looking generally and it seems like copyrights should normally have the year.

https://www.copyrighted.com/blog/copyright-date#:~:text=Does%20the%20copyright%20need%20to,public%2C%20aiding%20in%20legal%20matters.

Also, both Tekton and Knative have years.

- https://github.com/knative/serving/blob/main/pkg/activator/activator.go
- https://github.com/tektoncd/pipeline/blob/main/pkg/apis/config/events.go

The license is different, but I still think we should probably have a date.

maruiz93:
> I was looking generally and it seems like copyrights should normally have the year.

I see, but that will imply having files with 2024 year and new files created next year 2025? Or is it ok if we keep 2024 for all the files including the ones that will be created next year?

ggallen:
> > I was looking generally and it seems like copyrights should normally have the year.
> 
> I see, but that will imply having files with 2024 year and new files created next year 2025? Or is it ok if we keep 2024 for all the files including the ones that will be created next year?

I don't think it matters much. I think we could just always use 2024 as that is when KubeArchive was created.  The year just starts the clock ticking for copyright purposes.  We could also use a range moving forward, and when we update copyrights we could use 2024-2025 if we wanted, or whatever the current year is.

But I'd just go with 2024 for everything.

rh-hemartin:
> I think we should have a date in our copyright.

I think we shouldn't have date in our copyright.

On the OpenSSF badge program (recommended by the CNCF https://contribute.cncf.io/maintainers/security/security-guidelines/#5-openssf-best-practices-badging) they do not require date:

```
The project MUST include a copyright statement in each source file, identifying the copyright holder (e.g., the [project name] contributors). {Met justification} [[copyright_per_file](https://www.bestpractices.dev/en/criteria#2.copyright_per_file)]
```

As justification, they link to a article by the Linux Foundation in which they recommend against adding dates:

https://www.linuxfoundation.org/blog/blog/copyright-notices-in-open-source-software-projects

```
By using a common format, the project avoids having to maintain lists of names of the authors or copyright holders, years or ranges of years, and variations on the (c) symbol. 
```

We can flip a coin if we do not agree :rofl:

maruiz93:
@rh-hemartin reason of not having dates is quite backed up so enough reason for me to keep the license as is :)

mafh314 (approved):

PR #674, opened 2024-11-22
Do not deserialize resources from the database

### LINKED ISSUE #504: Investigate the impact of deserializing JSON in the API answer
Currently when we return an answer from the API the process queries the database and then deserializes the JSON into a `unstructured.Unstructured` struct using `json.Unmarshal`.  This process happens for all objects returned, but we don't do anything with the objects. Investigate the cost of this unmarshalling using benchmarks with Go.

**Acceptance Criteria**

* Use Go benchmarking (and maybe other more rudimentary tools) to see the performance of serializing vs not-deserializing.
* Don't change the API answer code yet, we just want information.

Notes:

* It does not even need to be in our code, maybe a script that does it and that it mimics what do we do could be enough.
* Make sure its the closest possible so we get a reasonable comparison.
* The best scenario is to run this in our code, so we get the "real" comparison.
* This may be OK with queries that are limited in size, but probably is worth optimizing for very large queries.

rh-hemartin:

Resolves #504 

## Release Note

```release-note
NONE
```

### DIFF
--- a/charts/kubearchive/templates/api_server/api_server.yaml
+++ b/charts/kubearchive/templates/api_server/api_server.yaml
@@ -35,7 +35,8 @@ spec:
               # See recommendations at https://tip.golang.org/doc/gc-guide#Memory_limit
               memory: 230Mi
             limits:
-              cpu: 200m
+              # Currently the API is sitting at 630m on performance tests
+              cpu: 700m
               memory: 256Mi
           volumeMounts:
             - name: tls-secret
--- a/pkg/database/database.go
+++ b/pkg/database/database.go
@@ -50,9 +50,9 @@ type DBParamParser interface {
 }
 
 type DBInterface interface {
-	QueryResources(ctx context.Context, kind, apiVersion, limit, continueId, continueDate string) ([]*unstructured.Unstructured, int64, string, error)
-	QueryNamespacedResources(ctx context.Context, kind, apiVersion, namespace, limit, continueId, continueDate string) ([]*unstructured.Unstructured, int64, string, error)
-	QueryNamespacedResourceByName(ctx context.Context, kind, apiVersion, namespace, name string) (*unstructured.Unstructured, error)
+	QueryResources(ctx context.Context, kind, apiVersion, limit, continueId, continueDate string) ([]string, int64, string, error)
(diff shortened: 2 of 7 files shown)


pkg/database/database.go
```diff
@@ -210,10 +210,7 @@ func (db *Database) performResourceQuery(ctx context.Context, pq *paramQuery) ([
 	lastRow := parsedRows[len(parsedRows)-1]
 
 	for _, parsedRow := range parsedRows {
-		r, errConversion := models.UnstructuredFromByteSlice(parsedRow.Resource)
-		if errConversion != nil {
-			return resources, 0, "", errConversion
-		}
+		r := string(parsedRow.Resource)
```
maruiz93, line 213: If parsing the `data` field is just going to be casting it to string I would suggest removing the whole `for` block and change the the retrieved type in the struct from `[]byte` to `string` (line 196)

maruiz93 (changes requested):

pkg/database/database.go
rh-hemartin: How do you go about removing the for loop? I need to loop over the rows to append their `Resource` to the list of resources.

maruiz93: Right, the for is needed, but you get rid of the parsing operation :) 
LGTM

maruiz93 (approved):

charts/kubearchive/templates/api_server/api_server.yaml
```diff
@@ -28,12 +28,13 @@ spec:
           image: {{ required "An API Server container image is required" .Values.apiServer.image }}
           resources:
             requests:
-              cpu: 200m
+              # Currently the API is sitting at 630m on performance tests
+              cpu: 700m
```
ggallen, line 32: Why change the request?  If I remember correctly, the request is just what is needed to start, but the limit it the max.  Do we always want to start with the limit?

ggallen (approved):

charts/kubearchive/templates/api_server/api_server.yaml
rh-hemartin: No real reason, I just changed both. I reverted it back to 200m because it makes sense, we won't be using 700m always, we are sitting at 630m under 100 requests per second, which I don't think we will hit often. Lets see.

PR #680, opened 2024-11-26
Add documentation style document

### LINKED ISSUE #293: Create a documentation guide
The guide should be about content, not about asciidoc markup, and it should include:

* Present over future tense
* When over once
* Division of content: getting started (focused on learning), how-to (focused on results), discussions (focused on understanding) and reference (focused on detailed specifications as configuration files, CLI flags, etc.)
* Keep it short and to the point

rh-hemartin:

Resolves #293 

## Release Note

```release-note
NONE
```

### DIFF
--- a/docs/modules/ROOT/pages/contributors/documentation.adoc
+++ b/docs/modules/ROOT/pages/contributors/documentation.adoc
@@ -1,0 +1,91 @@
+= Documentation Style
+
+When contributing documentation to KubeArchive you are expected to follow
+the guidelines listed below:
+
+* Use present tense, instead of future or past.
+* Use the active voice, instead of the passive voice.
+* Use the second and third persons, instead of first person.
+* Use positive phrases, instead of negative phrases.
+* Use cautious language, instead of making claims.
+* Do not recommend.
+* Use english abbreviations, instead of latin abbreviations.
+
+== Tense
+
+Use present tense whenever possible:
+
+NO: When configured in this way KubeArchive will delete resources.
+
+YES: When configured in this way KubeArchive deletes resources.
+
+== Voice
+
+Use active voice as much as possible, as its usually more direct than the passive:
+
+NO: This feature is configured in the ConfigMap.
+
+YES: Configure this feature in the ConfigMap.
+
+NO: Resources are deleted by KubeArchive when `deleteWhen` is configured.
+
+YES: KubeArchive deletes resources when you configure `deleteWhen`.
+
+== Persons
+
+Avoid first person and use second to interact with the reader and third to describe
+things:
+
+NO: We create the following `KubeArchiveConfig` resource.
+
+YES: Create the following `KubeArchiveConfig` resource.
+
+NO: Execute the file we provide.
+
+YES: Execute the file provided.
+
+== Positive Phrases
+
+Use positive phrases as negative ones are sometimes confusing, most notably when they
+contain two negatives:
+
+NO: Do not set the field when its value is redundant.
+
+YES: Set the field if its value is different to avoid redundancy.
+
+NO: If the feature is not configured KubeArchive does not delete resources.
+
+YES: If the feature is configured KubeArchive deletes resources.
+
+== Claims and Recommendations
+
+Use a "cautious" language when talking about some topics as performance, benefits, best
+practices or security:
+
+NO: KubeArchive improves the performance of your Kubernetes cluster.
+
+YES: KubeArchive can improve the performance of your Kubernetes cluster.
+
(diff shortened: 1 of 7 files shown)


maruiz93:
If we want to have this doc style we can use this PR to make the proper changes in the current docs. For example in  [Configuring KubeArchive](https://kubearchive.github.io/kubearchive/main/getting-started/configuration.html#_the_kubearchiveconfig_resource) section there are several sentences not compliant:

  *  Replace `KubeArchiveConfigs are limited to one per namespace and should be named kubearchive` for `KubeArchiveConfigs are limited to one per namespace and kubearchive is the only name allowed`
    
  *  Replace `Resources in the KubeArchiveConfig are defined using a selector` for `KubeArchiveConfig selector key defines the archived resources`
    
  *  Replace `The expressions are evaluated every time a resource defined by selector changes or gets deleted.` for `When the selector changes or gets deleted KubeArchive reevaluates the expressions`
    
  *  Replace `The pod will not appear in KubeArchive until it completes` for `The pod does not appear in KubeArchive until it completes`
    
  *  Replace `We use jq in these examples to reduce output.` for `The examples include jq to reduce the output`
     
  *  Replace `When resources are deleted with deleteWhen they get archived too.` for `KubeArchive also archives the resources defined with deleteWhen`
     
  *  Replace `Start watching pods to see how this pod is created, executed, completed and deleted` for `Start watching pods to see how they go through the different states: creating, running, completed, terminating and then KubeArchive gets them automatically removed`
     
  *  Replace `After the pod has been deleted from the cluster by KubeArchive, retrieve it` for `After KubeArchive removes the pod from the Kubernetes cluster, retrieve it`
  
  * Replace `KubeArchive can be used with existing applications that clean up resources` for `You can use KubeArchive with other applications that clean up resources`.
  
  * Replace `Query KubeArchive to check that only the pod that completed correctly (archived-on-deletion) was archived` for `Query KubeArchive to check the it only archives the pod that completed correctly (archived-on-deletion)`
  
I believe that in general we want to avoid using the latin expressions `e.g.`, `i.e.` so we can be also explicit about that.

maruiz93 (changes requested):
Added some suggestions

docs/modules/ROOT/pages/contributors/documentation.adoc
```diff
@@ -0,0 +1,74 @@
...
+
+Use active voice as much as possible, as its usually more direct than the passive:
+
+NO: This feature is configured in the ConfigMap.
+
+YES: Configure this feature in the ConfigMap.
+
+NO: Resources are deleted by KubeArchive when `deleteWhen` is configured.
+
+YES: KubeArchive deletes resources when you configure `deleteWhen`.
```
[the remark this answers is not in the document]
ggallen: You're right. I misread that.

rh-hemartin:
Fixed suggestions by @maruiz93 and added the latin abbreviations section, also suggested by @maruiz93

ggallen (approved):

maruiz93 (approved):

docs/modules/ROOT/pages/contributors/documentation.adoc
```diff
@@ -0,0 +1,93 @@
...
+* Use positive phrases, instead of negative phrases.
+* Use cautious language, instead of making claims.
+* Do not recommend.
+* Use english abbreviations, instead of latin abbreviations.
+
+== Tense
+
+Use present tense whenever possible:
+
+NO: When configured in this way KubeArchive will delete resources.
```
mafh314, line 18: This comment is just about preference (another option), I would add the examples as `CORRECT` and `INCORRECT` instead of YES` and `NO`

docs/modules/ROOT/pages/contributors/documentation.adoc
```diff
@@ -0,0 +1,93 @@
...
+
+YES: Used to login with ko (for example "quay.io")
+
+NO: These are style rules (syntax, indentation, naming, etc.)
+
+YES: These are style rules (syntax, indentation, naming, and so on)
+
+YES: These are style rules (syntax, indentation, naming, and other rules)
+
+NO: against the body of the cloud event (that is the resource)
```
mafh314, line 91: Line 91 and line 93 are the same example

docs/modules/ROOT/pages/contributors/documentation.adoc
rh-hemartin: I will keep it as is

mafh314 (approved):

Issue #556, opened 2024-10-15
Use gin http server to receive cloud events


skoved:
To be able to provide a health check endpoint or endpoint for another type of kubernetes probes, the sink needs to use its own http server instead of the one used in `cloudevents.NewClientHTTP`. Gin makes sense since we're already using it in the API server and we can structure the code in the sink similarly. This issue will enable work on #306

Issue #521, opened 2024-10-09
React to SIGINT and SIGTERM signals


rh-hemartin:
Currently we are not managing what happens when an interrupt signal is sent to the API or the Sink. The operator is the only that reacts because the framework `operator-sdk` provided a way in their scaffold. The Kubernetes documentation says:

> Because Pods represent processes running on nodes in the cluster, it is important to allow those processes to gracefully terminate when they are no longer needed (rather than being abruptly stopped with a KILL signal and having no chance to clean up).

https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination

So its kind of "recommended" that you cleanup after you are sent the signal to stop. We open database connections and web connections, so we should cleanup as well. This can be seen in KubeArchive because we don't log the closure of the DB connection that should happen, as the pod gets terminated instantaneously. Handling the signal properly the closure of the DB will occurs and line printed.

**Acceptance Criteria**

* The Sink handles SIGINT and SIGTERM signals.
* The API handles SIGINT and SIGTERM signals.

**Notes**

* See an example of how to achieve this in `SetupSignalHandler` from the controller runtime package
* Check this article: https://grafana.com/blog/2024/02/09/how-i-write-http-services-in-go-after-13-years/
* Or just search "golang signal handling http server" and see what you find

PR #683, opened 2024-11-28
Handle SIGTERM on the API

rh-hemartin:

Related to #521 and #556 

## Release Note

```release-note
The KubeArchive API now gracefully shuts down if the process receives a TERM or INTERRUPT signal (SIGTERM or SIGINT respectively).
```

## Notes for Reviewers

* Handling here the API, as the sink is not migrated to Gin yet (#556)
* SIGINT is not emitted by Kubernetes, but its the interrupt signal so I think is just normal that we handle it as well.
* To test delete the API pod. In the logs you should see the different messages.

### DIFF
--- a/cmd/api/main.go
+++ b/cmd/api/main.go
@@ -4,10 +4,13 @@
 package main
 
 import (
+	"context"
 	"fmt"
 	"log/slog"
 	"net/http"
 	"os"
+	"os/signal"
+	"syscall"
 	"time"
 
 	"github.com/Cyprinus12138/otelgin"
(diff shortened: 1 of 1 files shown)


cmd/api/main.go
```diff
@@ -117,18 +120,49 @@ func main() {
...
+	err = httpServer.Shutdown(ctx)
 	if err != nil {
-		slog.Error("Could not run server on localhost", "error", err.Error())
+		slog.Error("Error shutting down the server", "error", err.Error())
 		os.Exit(1)
 	}
+
+	// This blocks until the context expires
+	<-ctx.Done()
+	slog.Debug("Shutdown reached timeout")
```
maruiz93, line 165: I have a question regarding this line. In the happy path of a server shutdown without errors, are we waiting the 5 seconds timeout instead of finishing earlier?

rh-hemartin: I checked the implementation and it looks like the server returns as fast as it can, so the happy path is as fast as possible.

maruiz93: nice!!

maruiz93 (approved):

mafh314 (approved):

rh-hemartin:
> I've tested this and I saw in the api deployment logs the "Shutting down server" message :) @rh-hemartin I have an small suggestion about the comments that you added, should those follow the style guide as well? for example this phrase: `We do not accept bodies yet, because we are read-only, so we set a`

I don't think we need to follow the style guide. They are documentation, but the documentation style guide is intended for user facing documentation at least.

Issue #631, opened 2024-11-11
Add a container filter for the log_url retrieval


maruiz93:
In `kubectl` the logs can be filtered by container name. 
In KubeArchive we have the logs stored by pod uuid and container name, so they can be also filtered.

### Acceptance Criteria

* Include query parameter in the `/log` endpoints that allow this filtering the same way that the [Kubernetes API does](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.25/#read-log-pod-v1-core)

PR #687, opened 2024-12-02
Add logging backend integration

maruiz93:

Resolves #

## Release Note

```release-note
NONE
```

## Notes for Reviewers

Still working on tests and also need to merge with the changes on #663 and #665.
Instead of adding the configmap read,as that is going to be added on #665 I tested my changes with
hardcoded values shown in `api/main.go`


ggallen:
> Kubernetes API Implementation [linked here](https://github.com/kubernetes/kubernetes/blob/v1.25.5/pkg/routes/logs.go#L44) uses [http.ServeFile](https://pkg.go.dev/net/http#ServeFile) that ends up using [copyBuffer](https://cs.opensource.google/go/go/+/refs/tags/go1.23.3:src/io/io.go;l=407) function that streams the content from a reader to a writer on byte chunks

Will streaming with with Elasticsearch, which I believe returns a JSON list of maps?

cmd/api/logging/logging.go
```diff
@@ -0,0 +1,108 @@
...
+	// FIXME For now the queries to the logging backend server are done insecurely
+	client := http.Client{
+		Transport: otelhttp.NewTransport(&http.Transport{
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: true}})}
+	user := loggingConfig[userKey]
+	password := loggingConfig[passwordKey]
+	jsonPathParser, err := jp.ParseString(loggingConfig[jsonPathKey])
+	if user == "" || password == "" || err != nil {
+		return func(c *gin.Context) {
+			abort.Abort(c, fmt.Errorf("invalid jsonPath expression in logging config: %s", err.Error()),
```
ggallen, line 39: This error doesn't make sense if it's actually the `USERNAME` or `PASSWORD`  that is empty.

cmd/api/routers/routers.go
```diff
@@ -89,7 +89,7 @@ func (c *Controller) GetResourceLogs(context *gin.Context) {
 		apiVersion = fmt.Sprintf("%s/%s", group, version)
 	}
 
-	logURLs, err := c.Database.QueryLogURLs(
+	logURL, err := c.Database.QueryLogURL(
```
ggallen, line 92: Where is the container name here?  Or I assume you are going to default to one of the containers?

But don't we have to allow a user to specify which container they want?

pkg/database/database.go
```diff
@@ -27,7 +28,7 @@ var RegisteredDBCreators = make(map[string]newDBCreatorFunc)
 type DBInterface interface {
 	QueryResources(ctx context.Context, kind, apiVersion, namespace,
 		name, continueId, continueDate string, limit int) ([]string, int64, string, error)
-	QueryLogURLs(ctx context.Context, kind, apiVersion, namespace, name string) ([]string, error)
+	QueryLogURL(ctx context.Context, kind, apiVersion, namespace, name string) (string, error)
```
ggallen, line 31: The same question again about getting the log for a specific container.

pkg/database/fake/database_generated.go
```diff
@@ -73,15 +73,15 @@ func (f *Database) queryResources(_ context.Context, kind, version, _, _ string,
 	return f.filterResourcesByKindAndApiVersion(kind, version)
 }
 
-func (f *Database) QueryLogURLs(ctx context.Context, kind, apiVersion, namespace, name string) ([]string, error) {
+func (f *Database) QueryLogURL(ctx context.Context, kind, _, _, _ string) (string, error) {
 	if kind == "Pod" {
-		return []string{f.logUrl[0].Url}, f.err
+		return f.logUrl[0].Url, f.err
```
ggallen, line 78: Why is `Pod` differentiated here?

pkg/database/fake/database_generated.go
```diff
@@ -73,15 +73,15 @@ func (f *Database) queryResources(_ context.Context, kind, version, _, _ string,
...
 	}
 	var urls []string
 	for _, l := range f.logUrl {
 		urls = append(urls, l.Url)
 	}
-	return urls, f.err
+	return urls[0], f.err
```
ggallen, line 84: Why have this loop?  Why not just return `f.logUrl[0]`?

pkg/database/fake/database_generated.go
```diff
@@ -73,15 +73,15 @@ func (f *Database) queryResources(_ context.Context, kind, version, _, _ string,
 	return f.filterResourcesByKindAndApiVersion(kind, version)
 }
 
-func (f *Database) QueryLogURLs(ctx context.Context, kind, apiVersion, namespace, name string) ([]string, error) {
+func (f *Database) QueryLogURL(ctx context.Context, kind, _, _, _ string) (string, error) {
```
ggallen, line 76: Is it possible that a `Pod` may not have any logs?  If so, that needs to be checked here as `f.logUrl[0]` will then panic.

cmd/api/routers/routers.go
maruiz93: Yes, we need to allow the user to specify which container they want but I think we also should work when no container is specified and default to return one of them (replicating the `kubectl logs` command behavior). 

For introducing the container filtering there is a separate issue - https://github.com/kubearchive/kubearchive/issues/631

pkg/database/database.go
maruiz93: Same answer: 

Yes, we need to allow the user to specify which container they want but I think we also should work when no container is specified and default to return one of them (replicating the kubectl logs command behavior).

For introducing the container filtering there is a separate issue - https://github.com/kubearchive/kubearchive/issues/631

pkg/database/fake/database_generated.go
maruiz93: Note that this module is just for testing purposes to be used as a "fake" database interface implementation. In this case, the pods always have logs, so f.logUrl[0] shouldn't fail.

pkg/database/fake/database_generated.go
maruiz93: Right, with the latest changes this now doesn't make sense, I need to just return `f.logUrl[0]`

pkg/database/fake/database_generated.go
maruiz93: Right, with the latest changes this now doesn't make sense, I need to just return `f.logUrl[0]`

maruiz93: fixed

pkg/database/fake/database_generated.go
maruiz93: fixed

cmd/api/logging/logging.go
maruiz93: fixed

maruiz93:
> > Kubernetes API Implementation [linked here](https://github.com/kubernetes/kubernetes/blob/v1.25.5/pkg/routes/logs.go#L44) uses [http.ServeFile](https://pkg.go.dev/net/http#ServeFile) that ends up using [copyBuffer](https://cs.opensource.google/go/go/+/refs/tags/go1.23.3:src/io/io.go;l=407) function that streams the content from a reader to a writer on byte chunks
> 
> Will streaming with with Elasticsearch, which I believe returns a JSON list of maps?

I've created a separate issue to address logging integration tests with elasticsearch. If we need to change the approach could be done when addressing that issue.

cmd/api/logging/logging.go
```diff
@@ -0,0 +1,162 @@
...
+func LogRetrieval(loggingCreds map[string]string, loggingCredsErr error) gin.HandlerFunc {
+	user := loggingCreds[userKey]
+	password := loggingCreds[passwordKey]
+	if user == "" || password == "" || loggingCredsErr != nil {
+		loggingCredsErr = fmt.Errorf(
+			"logging secret values unset. To configure logging update the kubearchive-logging Secret")
+		return func(c *gin.Context) {
+			abort.Abort(c, loggingCredsErr, http.StatusBadRequest)
+		}
+	}
```
rh-hemartin, line 60: I think this piece should execute before `Controller.GetLogURL` to avoid executing it as it reaches for the database. This way if the credentials are not specified, we don't even reach the database to read log information.

cmd/api/logging/logging.go
```diff
@@ -0,0 +1,162 @@
...
+		var logsReturned bool
+		for readErr == nil {
+			parsedLine, errParseLine := parseLine(line, jsonPathParser)
+			if errParseLine != nil {
+				abort.Abort(c, errParseLine, http.StatusInternalServerError)
+				return
+			}
+			if parsedLine != "" {
+				logsReturned = true
+				c.JSON(http.StatusOK, parsedLine+"\n")
```
rh-hemartin, line 116: Why do we add a jumpline here? I guess its because it failed without it, but why did it fail to create a JSON from a string and needed a jumpline? Or this is because of a representation in your console?

cmd/api/logging/logging.go
```diff
@@ -0,0 +1,162 @@
...
+		logUrl := c.GetString("logURL")
+		jsonPath := c.GetString("jsonPath")
+
+		if logUrl == "" {
+			abort.Abort(c, fmt.Errorf("no log URL found"), http.StatusNotFound)
+			return
+		}
+		slog.Info("Retrieving logs", "logURL", logUrl)
+
+		var jsonPathParser jp.Expr
```
rh-hemartin, line 77: What happens if this never gets populated? What does it match by default?

cmd/api/logging/logging.go
maruiz93: It's for readability, the logs come in JSON separated with "\n", so I parse each line and return it again separated with a "\n". Otherwise the readability is pretty bad.

cmd/api/logging/logging.go
maruiz93: good point, I'll do some adjustments

cmd/api/logging/logging.go
maruiz93: if this value never gets populated `jsonPathParser` is nil so when the line is parsed in the `parseLine` function (line 128), when `jsonPathParser` is checked, if its value is nil, the whole line is returned without applying any filters.

cmd/api/logging/logging.go
maruiz93: I keep using `c.JSON` as we do in the rest of the handlers to keep sending the Content-Type as "application/json" but for the tests that I did with the log-generator (the same from the integration tests) the log output isn't in JSON format, it could be not needed.

I have no strong preference for keeping `c.JSON`, at least for now and the use cases we are thinking of. I can change it to `c.String` if you do.

rh-hemartin: Make it `c.String` we want to avoid serialization/deserialization as much as possible and in this case the impact can be higher as there is no limit to these log lines.

cmd/api/logging/logging.go
maruiz93: Adjustments done, now the middleware is splitted up in two pieces. The first piece runs before the handler and sets the credentials in the request context, the second piece just interacts with the logging backend to retrieve the logs.

This means that now if the credentials aren't set, the handler to get the log url from the database won't be executed.

cmd/api/logging/logging.go
maruiz93: Done!

cmd/api/logging/logging.go
```diff
@@ -0,0 +1,178 @@
...
+		if jsonPath != "" {
+			jsonPathParser, errJsonPath = jp.ParseString(jsonPath)
+			if errJsonPath != nil {
+				abort.Abort(c, fmt.Errorf("invalid jsonPath %s for url %s", jsonPath, logUrl),
+					http.StatusInternalServerError)
+				return
+			}
+		}
+
+		request, errReq := http.NewRequest(http.MethodGet, logUrl, nil)
```
skoved, line 105: Since your `http.Client` is using `otelhttp.Transport` you should use `http.NewRequestWithContext` so that the Otel trace is passed along.

```suggestion
		request, errReq := http.NewRequestWithContext(c.Request.Context(), http.MethodGet, logUrl, nil)
```

pkg/database/database.go
```diff
@@ -121,46 +141,55 @@ func (db *Database) QueryLogURLs(ctx context.Context, kind, apiVersion, namespac
...
+		return "", "", err
 	}
+	if len(pods) == 0 {
+		return "", "", ResourceNotFoundError
+	}
+
+	// Get the most recent pod from owned by the provided resource
+	sort.Slice(pods, func(i, j int) bool {
+		return pods[i].Date > pods[j].Date
+	})
```
skoved, line 162: the docs for [`sort.Slice`](https://pkg.go.dev/sort#Slice) says that [`slices.SortFunc`](https://pkg.go.dev/slices#SortFunc) is generally faster. It might be worth it instead

skoved (commented):
looks good! There were a few things that I noticed but I think they're relatively small

cmd/api/logging/logging.go
```diff
@@ -0,0 +1,178 @@
...
+			if errParseLine != nil {
+				abort.Abort(c, errParseLine, http.StatusInternalServerError)
+				return
+			}
+			if parsedLine != "" {
+				logsReturned = true
+				c.String(http.StatusOK, parsedLine+"\n")
+			}
+			line, readErr = readLine(reader)
+		}
```
ggallen, line 135: I think this might be more clearly written like this:

```
		var logsReturned bool
		for {
			line, readErr = readLine(reader)
            if readErr != nil {
                break
            }
			parsedLine, errParseLine := parseLine(line, jsonPathParser)
			if errParseLine != nil {
				abort.Abort(c, errParseLine, http.StatusInternalServerError)
				return
			}
			if parsedLine != "" {
				logsReturned = true
				c.String(http.StatusOK, parsedLine+"\n")
			}
		}
```

cmd/api/logging/logging.go
```diff
@@ -0,0 +1,178 @@
...
+		return string(line), nil
+	}
+
+	var jsonLine any
+	var errJson error
+	var result string
+
+	jsonLine, errJson = oj.Parse(line)
+	if errJson != nil {
+		return result, errJson
```
ggallen, line 155: Why return `result` on error and not `nil`?

cmd/api/logging/logging.go
```diff
@@ -0,0 +1,178 @@
...
+		isPrefix            = true
+		err           error = nil
+		tmpLine, line []byte
+	)
+	for isPrefix && err == nil {
+		tmpLine, isPrefix, err = reader.ReadLine()
+		line = append(line, tmpLine...)
+	}
+	return line, err
+}
```
ggallen, line 178: Any reason not to simply use `bufio.ReadBytes('\n')` rather than writing your own?

integrations/logging/splunk/patch-jsonpath.yaml
```diff
@@ -0,0 +1,7 @@
+# Copyright KubeArchive Authors
+# SPDX-License-Identifier: Apache-2.0
+---
+data:
+  POD_ID: "cel:metadata.uid"
+  LOG_URL: "https://splunk-single-standalone-service.splunk-operator.svc.cluster.local:8089/services/search/jobs/export?search=search%20%2A%20%7C%20spath%20%22kubernetes.pod_id%22%20%7C%20search%20%22kubernetes.pod_id%22%3D%22{POD_ID}%22%20%7C%20spath%20%22kubernetes.container_name%22%20%7C%20search%20%22kubernetes.container_name%22%3D%22{CONTAINER_NAME}%22%20%7C%20sort%20time%20%7C%20table%20%22message%22&output_mode=json"
+  LOG_URL_JSONPATH: "$.result.message"
```
ggallen, line 7: Do we really want this as part of our "base" install?  Logging isn't required, and not everyone will be using Splunk even if they configure logging.

pkg/database/database.go
```diff
@@ -121,46 +141,55 @@ func (db *Database) QueryLogURLs(ctx context.Context, kind, apiVersion, namespac
...
+	sb.Where(db.Filter.UuidsFilter(sb.Cond, []string{pods[0].Uuid}))
+	logUrls, err := logQueryPerformer.performQuery(ctx, sb)
+	if err != nil {
+		return "", "", err
+	}
+	if len(logUrls) >= 1 {
+		return logUrls[0].LogURL, logUrls[0].JsonPath, nil
+	} else {
+		return "", "", ResourceNotFoundError
+	}
```
ggallen, line 174: Why not return `nil` here rather than `""`?  It seems odd.

rh-hemartin (approved):

cmd/api/logging/logging.go
maruiz93: > Why is loggingCredsError passed in here? It's re-used and ends up hiding the real error.

I think you have made a good point here, I am going to log this errors and also output them in the abort error message concatenated with the current message: `logging secret values unset. To configure logging update the kubearchive-logging Secret`

> I believe these are going to panic if GetKubeArchiveLoggingCredentials fails.

When a map doesn't have the requested values, it doesn't panic in go, it just retrieves the nil value for that object which for strings is an empty string.

> Is abort the right thing to do here?

I think so. The idea is that if you have the logging credentials unset, and then you ask for logs in the log endpoint, it should say `Bad Request` with the explanation of why is this a bad request (because the configuration doesn't allow to request logs)

Note that `abort` is a function defined in the API for correctly log and output an HTTP error (status code > 299) as an endpoint response.

cmd/api/logging/logging.go
```diff
@@ -0,0 +1,178 @@
...
+
+// LogRetrieval retrieves a middleware function that checks the logging config and when set
+// expects to find a log url in the context from where retrieve and parse logs
+// It should be called when user, password and logURL are set in the context.
+func LogRetrieval() gin.HandlerFunc {
+	// FIXME For now the queries to the logging backend server are done insecurely. Needed for the current test env.
+	// #nosec G402
+	client := http.Client{
+		Transport: otelhttp.NewTransport(&http.Transport{
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: true}})}
```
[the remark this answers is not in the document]
maruiz93: great point

charts/kubearchive/templates/logging/secret.yaml
```diff
@@ -8,5 +8,5 @@ metadata:
   namespace: {{ .Release.Namespace }}
 type: Opaque
 data:
-  password: Y2hhbmdlbWU=  # notsecret
-  user: YWRtaW4=
+  PASSWORD: Y2hhbmdlbWU=  # notsecret
+  USER: YWRtaW4=
```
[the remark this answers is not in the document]
maruiz93: I wanted to keep the same case that we are using in the kubearchive-logging ConfigMap. See [docs](https://kubearchive.github.io/kubearchive/main/design/logging-integration.html#generating_log_urls) and [hardcoded expected keys](https://github.com/kubearchive/kubearchive/blob/main/cmd/sink/logs/logs.go#L23-L23)

cmd/api/logging/logging.go
maruiz93: `result` is a string, so it can't be `nil` (not all types can be `nil` in golang, the default/empty value of a string is `""` (empty string), while for a bool is `false` and for an int is `0`. More details [here](https://go.dev/tour/basics/12)

BTW, at that point in the function, `result` is going to be `""` (string zero value). If you think that explicitly returning an empty string makes things clear I can replace it.

cmd/api/logging/logging.go
maruiz93: I made the decission based on [this stackoverflow answer](https://stackoverflow.com/questions/8757389/reading-a-file-line-by-line-in-go#answer-8758113). The points I considered were:
* If the last log line doesn't end with a `\n` won't be parsered by `bufio.ReadBytes('\n')` or `bufio.ReadString('\n')`
* `bufio.ReadLine` doesn't return the full line when it's too large that's why I need to do some extra process to the returned results of `reader.ReadLine()`

integrations/logging/splunk/patch-jsonpath.yaml
maruiz93: I see this patch the same way I see the database integrations. Our "base" install doesn't include them, if the user wants to use a PostgreSQL DB, they shouldn't install it on the cluster from our script `integrations/database/postgresql/install.sh` they should have their own database and the provide the proper kubearchive-logging Secret and ConfigMap.

The integration scripts are mainly used for test and development environments.

The same happens with splunk. If the customer uses splunk, they won't be using the splunk deployed with `integrations/logging/splunk/install.sh`. This is only for test and development.

Also, note that by default this integration isn't included and that there is a separate script for elasticsearch logging integration (also for test and development) in `integrations/logging/elasticsearch/install.sh`

pkg/database/database.go
maruiz93: because the returned type is a string. A string in golang can't be `nil`, its zero value is an empty string.

cmd/api/logging/logging.go
maruiz93: I think that your proposal is based on doing the "explicit early return" recommended by golang in general terms, but I also think that you might find the `for readErr == nil` confusing because it's a `while` hidden under the name of a `for`. It's just something I wanted to mention because I needed to do that "change of mind" when programming in go.

By the way, it's no big deal for me going with your proposal so I'll apply it.

cmd/api/logging/logging.go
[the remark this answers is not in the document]
maruiz93: Done! I've set a minute just in case, but we can tune it later or maybe even set it as configurable by an environment variable (as we do with the cache)

cmd/api/logging/logging.go
maruiz93: Done!

pkg/database/database.go
maruiz93: I haven't noticed that suggestion! Thanks for the pointers, the change is already in place :)

cmd/api/logging/logging.go
maruiz93: Change done!

skoved (approved):

cmd/api/logging/logging.go
maruiz93: Done

ggallen: >> I believe these are going to panic if GetKubeArchiveLoggingCredentials fails.
>
> When a map doesn't have the requested values, it doesn't panic in go, it just retrieves the nil value for that object which for strings is an empty string.

That's my point. I believe the map is `nil`.

charts/kubearchive/templates/logging/secret.yaml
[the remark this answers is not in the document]
ggallen: But this is a `Secret` and not a `ConfigMap`.  The keys in the `ConfigMap` are upper-case in this instance because they are just inserted into the environment for substitution, and environment variables are normally upper-case. And we are basing our substitution on bash-like environment variables.

Are these used in that way at all?

To me, when I see all upper case I think it's an environment variable or a constant (which I know isn't how go does it).

This isn't a big deal, but these types of changes are really preferences, and they take away for the real changes you are making. Those are big enough, and throwing in these kinds of changes makes it more difficult to review.

integrations/logging/splunk/patch-jsonpath.yaml
ggallen: Fair enough. I didn't notice this file was in the integrations directory.

pkg/database/database.go
ggallen: I didn't know that. Thanks.

cmd/api/logging/logging.go
ggallen: No, that's fine. I didn't know that a string couldn't be `nil`.

charts/kubearchive/templates/logging/secret.yaml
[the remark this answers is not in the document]
maruiz93: > The keys in the ConfigMap are upper-case in this instance because they are just inserted into the environment for substitution, and environment variables are normally upper-case.

The `kubearchive-logging` ConfigMap is currently being mounted as a volume in the sink [1], the same as `kubearchive-logging` Secret is being mounted as a volume in the api-server.
This means that the variables inside both of them are both being read as files and not as environment variables so we shouldn't be casing the variables as uppercase if the reason is for being env vars.

> This isn't a big deal, but these types of changes are really preferences

I'm sorry it looked like that, I should've be specific about this change in the description's PR, but it wasn't a preference choice but a "looking for consistency" choice. I don't mind having the variables with lower case, but I do care of following the same convention we have for the other ConfigMap/Secrets that we have in place being consumed identically (in this case not as env vars but as files).

FWIW, we currently have the following secrets/configmaps:
* `kubearchive-database-credentials` Secret, exposed as environment variables in both sink and api-server
* `sink-filters` ConfigMap, exposed as files in the sink
* `kubearchive-logging` ConfigMap exposed as files in the sink
* `kubearchive-logging` Secret exposed as files in the api-server

Despite the `sink-filters` which are a bit special usecase as it is filled on runtime by the sink for internal purposes, all of them are uppercased, so for this PR, it makes sense to follow that.
If you think that we should change the case convention for variables exposed as files, I'm willing to discuss it, but I think it's outside of the scope of this PR.

[1] https://github.com/maruiz93/kubearchive/blob/main/charts/kubearchive/templates/sink/sink.yaml#L41-L41

cmd/api/logging/logging.go
maruiz93: hmm, right... I'm trying to change it now (I found some issues on the related unit test so I'm trying to address both)

charts/kubearchive/templates/logging/secret.yaml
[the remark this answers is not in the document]
maruiz93: As talked in a meeting. We will keep them upper case for now. In the kubearchive-logging ConfigMap the `JSON_PATH` variable isn't used for "substitution" as in "bash" style and it was kept upper case for consistency, so we could do the same for the Secrets.

cmd/api/logging/logging.go
maruiz93: Fixed

cmd/api/logging/logging.go
maruiz93: Even if map is `nil` go accepts looking for keys on it. I guess that's because go is statically typed.

Here is a POC where you can see what I mean: https://go.dev/play/p/6nuqM9pMmlH

ggallen: That is bizarre, but it's explained [here](https://stackoverflow.com/questions/77151147/why-go-does-not-panic-when-trying-to-read-from-a-nil-map).

ggallen (approved):

PR #694, opened 2024-12-04
Merge API handlers

### LINKED ISSUE #641: Merge API handlers

rh-hemartin:

Resolves #641 

## Release Note

```release-note
NONE
```

## Notes for Reviewers

* Merged Database's `QueryNamespacedResources` and `QueryResources` so its more compact. I think the logic does not look bad in that case, let me know what do you think.

### DIFF
--- a/cmd/api/routers/routers.go
+++ b/cmd/api/routers/routers.go
@@ -31,22 +31,40 @@ type Controller struct {
 
 const listString = `{"kind": "List", "apiVersion": "v1", "metadata": {"continue": "%s"}, "items": [%s]}`
 
-// GetAllResources responds with the list of resources of a specific type across all namespaces
-func (c *Controller) GetAllResources(context *gin.Context) {
-	group := context.Param("group")
-	version := context.Param("version")
-	apiVersion := fmt.Sprintf("%s/%s", group, version)
+func (c *Controller) GetResources(context *gin.Context) {
 	limit, id, date := pagination.GetValuesFromContext(context)
-
 	kind, err := discovery.GetAPIResourceKind(context)
 	if err != nil {
 		abort.Abort(context, err, http.StatusInternalServerError)
 		return
 	}
 
-	resources, lastId, lastDate, err := c.Database.QueryResources(context.Request.Context(), kind, apiVersion, limit, id, date)
+	group := context.Param("group")
+	version := context.Param("version")
+	namespace := context.Param("namespace")
+	name := context.Param("name")
+
+	apiVersion := version
+	if group != "" {
+		apiVersion = fmt.Sprintf("%s/%s", group, version)
+	}
+
+	// We send namespace even if its an empty string (non-namespaced resources) the Database
+	// knows what to do
+	resources, lastId, lastDate, err := c.Database.QueryResources(
+		context.Request.Context(), kind, apiVersion, namespace, name, limit, id, date)
+
 	if err != nil {
 		abort.Abort(context, err, http.StatusInternalServerError)
+		return
+	}
+
+	if name != "" {
+		if len(resources) > 1 {
+			abort.Abort(context, errors.New("more than one resource found"), http.StatusInternalServerError)
+			return
+		}
+		context.String(http.StatusOK, resources[0])
 		return
 	}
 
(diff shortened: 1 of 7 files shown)


cmd/api/routers/routers.go
```diff
@@ -41,43 +41,33 @@ type List struct {
...
+	}
+
+	var resources []*unstructured.Unstructured
+	var lastId int64
+	var lastDate string
+	if namespace != "" {
+		resources, lastId, lastDate, err = c.Database.QueryNamespacedResources(context.Request.Context(), kind, apiVersion, namespace, limit, id, date)
+	} else {
+		resources, lastId, lastDate, err = c.Database.QueryResources(context.Request.Context(), kind, apiVersion, limit, id, date)
 	}
```
maruiz93, line 69: I believe that we could make a similar merge of the Database `QueryNamespacedResources` and `QueryResources`, so we could replace this if/else block to a one call. 

I am ok doing that in a separate PR though.

maruiz93 (approved):

cmd/api/routers/routers.go
rh-hemartin: I gave it a try, let me know if it looks good.

pkg/database/database.go
```diff
@@ -50,8 +50,7 @@ type DBParamParser interface {
 }
 
 type DBInterface interface {
-	QueryResources(ctx context.Context, kind, apiVersion, limit, continueId, continueDate string) ([]*unstructured.Unstructured, int64, string, error)
-	QueryNamespacedResources(ctx context.Context, kind, apiVersion, namespace, limit, continueId, continueDate string) ([]*unstructured.Unstructured, int64, string, error)
+	QueryResources(ctx context.Context, kind, apiVersion, namespace, limit, continueId, continueDate string) ([]*unstructured.Unstructured, int64, string, error)
```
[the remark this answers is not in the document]
rh-hemartin: Good idea, I also implemented something that includes the name resources. Let me  know what you think.

pkg/database/database.go
```diff
@@ -86,49 +84,32 @@ func (db *Database) Ping(ctx context.Context) error {
...
-
-	if continueId != "" && continueDate != "" {
-		pq = db.newParamQuery(db.info.GetNamespacedResourcesLimitedContinueSQL())
-		pq.addStringParams(kind, apiVersion, namespace, continueDate, continueId, limit)
-	} else {
+	if namespace != "" {
 		pq = db.newParamQuery(db.info.GetNamespacedResourcesLimitedSQL())
 		pq.addStringParams(kind, apiVersion, namespace, limit)
+
+		if name != "" {
```
ggallen, line 96: Can you only query using a `name` if you've used a `namespace` too?

rh-hemartin: Yes, to retrieve by name you need to specify namespace as we don't support cluster wide resources.

And yes, `limit` is not compatible with `name`. Retrieving by name always retrieve one element or it fails if there is more than one element.

cmd/api/routers/routers.go
```diff
@@ -31,171 +31,64 @@ type Controller struct {
...
 
-	resource, err := c.Database.QueryNamespacedResourceByName(context.Request.Context(), kind, apiVersion, namespace, name)
-	if err != nil {
-		abort.Abort(context, err, http.StatusInternalServerError)
-		return
-	}
+	// We send namespace even if its an empty string (non-namespaced resources) the Database
+	// knows what to do
+	resources, lastId, lastDate, err := c.Database.QueryResources(context.Request.Context(), kind, apiVersion, namespace, name, limit, id, date)
```
maruiz93, line 54: We should be mindful with long lines,  line break suggestion and also added a couple of fixes in the comment
```suggestion
	// We send name and namespace even if it's an empty string (non-namespaced resources) the Database
	// knows what to do
	resources, lastId, lastDate, err := c.Database.QueryResources(context.Request.Context(), 
	    kind, apiVersion, namespace, name, limit, id, date)
```

pkg/database/database.go
```diff
@@ -50,9 +50,7 @@ type DBParamParser interface {
 }
 
 type DBInterface interface {
-	QueryResources(ctx context.Context, kind, apiVersion, limit, continueId, continueDate string) ([]string, int64, string, error)
-	QueryNamespacedResources(ctx context.Context, kind, apiVersion, namespace, limit, continueId, continueDate string) ([]string, int64, string, error)
-	QueryNamespacedResourceByName(ctx context.Context, kind, apiVersion, namespace, name string) (string, error)
+	QueryResources(ctx context.Context, kind, apiVersion, namespace, name, limit, continueId, continueDate string) ([]string, int64, string, error)
```
maruiz93, line 53: Be mindful to the line length:
```suggestion
	QueryResources(ctx context.Context, 
	    kind, apiVersion, namespace, name, limit, continueId, continueDate string) ([]string, int64, string, error)
```

pkg/database/fake/database_generated.go
```diff
@@ -102,8 +84,17 @@ func (f *Database) QueryLogURLs(ctx context.Context, kind, apiVersion, namespace
 	return urls, f.err
 }
 
-func (f *Database) QueryNamespacedResources(ctx context.Context, kind, version, namespace, limit, continueId, continueDate string) ([]string, int64, string, error) {
-	resources := f.filterResourcesByKindApiVersionAndNamespace(kind, version, namespace)
+func (f *Database) QueryResources(ctx context.Context, kind, version, namespace, name, limit, continueId, continueDate string) ([]string, int64, string, error) {
```
maruiz93, line 87: line break:
```suggestion
func (f *Database) QueryResources(ctx context.Context, 
    kind, version, namespace, name, limit, continueId, continueDate string) ([]string, int64, string, error) {
```

maruiz93 (approved):

ggallen (approved):

cmd/api/routers/routers.go
rh-hemartin: Done

PR #707, opened 2024-12-12
add json_path to log_url table

### LINKED ISSUE #689: Store JSONPath in along the log url in the database
### Context

We currently store the `url` in the `log_url` table and avoid building the url on retrieval instead of on insertion because that prevents having wrong old urls. E.g, if the logging backend system changes, the new pods will have the new log urls while the old ones will keep the old system, so the information is still "correct". 
Now that we have a new element in place, the JSONPath to be able to parse the logs retrieved from the logging system, following the same logic we should store it in the database.

Another advantage of this design is that the `kubearchive-logging` configmap is only coupled to the sink but not to both sink and api.

### Acceptance Criteria

* Store the JSONPath along the `url` in a new column of the `log_url` database table.

skoved:

Resolves #689 

## Release Note

```release-note
add json_path column to the log_url table
sink writes the value for LOG_URL_JSONPATH from the logging ConfigMap to the json_path column for the log_url table
```

## Notes for Reviewers

UrlBuilder stores the value for `LOG_URL_JSONPATH` from the logging ConfigMap. Sink calls the `UrlBuilder.GetJsonPath()` method when writing log urls to the database.

### DIFF
--- a/integrations/database/mariadb/kubearchive.sql
+++ b/integrations/database/mariadb/kubearchive.sql
@@ -59,6 +59,7 @@ CREATE TABLE `log_url` (
   `uuid` uuid NOT NULL,
   `url` text NOT NULL,
   `container_name` text NOT NULL,
+  `json_path` text,
   `created_at` timestamp NOT NULL DEFAULT current_timestamp(),
   `updated_at` timestamp NOT NULL DEFAULT current_timestamp()
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb3 COLLATE=utf8mb3_general_ci;
--- a/integrations/database/postgresql/kubearchive.sql
+++ b/integrations/database/postgresql/kubearchive.sql
@@ -88,6 +88,7 @@ CREATE TABLE public.log_url (
     uuid uuid NOT NULL,
     url text NOT NULL,
     container_name text NOT NULL,
+	json_path text,
     created_at timestamp without time zone DEFAULT now() NOT NULL,
     updated_at timestamp without time zone DEFAULT now() NOT NULL
 );
--- a/pkg/database/facade/inserter.go
+++ b/pkg/database/facade/inserter.go
@@ -16,15 +16,15 @@ type DBInserter interface {
 		clusterDeletedTs sql.NullString,
 		data []byte,
 	) *sqlbuilder.InsertBuilder
-	UrlInserter(uuid, url, containerName string) *sqlbuilder.InsertBuilder
+	UrlInserter(uuid, url, containerName, jsonPath string) *sqlbuilder.InsertBuilder
 }
 
 type PartialDBInserterImpl struct{}
 
-func (PartialDBInserterImpl) UrlInserter(uuid, url, containerName string) *sqlbuilder.InsertBuilder {
+func (PartialDBInserterImpl) UrlInserter(uuid, url, containerName, jsonPath string) *sqlbuilder.InsertBuilder {
 	ib := sqlbuilder.NewInsertBuilder()
 	ib.InsertInto("log_url")
-	ib.Cols("uuid", "url", "container_name")
-	ib.Values(uuid, url, containerName)
+	ib.Cols("uuid", "url", "container_name", "json_path")
+	ib.Values(uuid, url, containerName, jsonPath)
 	return ib
 }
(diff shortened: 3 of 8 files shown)


integrations/database/mariadb/kubearchive.sql
```diff
@@ -59,6 +59,7 @@ CREATE TABLE `log_url` (
   `uuid` uuid NOT NULL,
   `url` text NOT NULL,
   `container_name` text NOT NULL,
+  `json_path` text NOT NULL,
```
ggallen, line 62: This should not be `NOT NULL`. It's possible that some logging system will not need it.

@maruiz93 , take note.

integrations/database/postgresql/kubearchive.sql
```diff
@@ -88,6 +88,7 @@ CREATE TABLE public.log_url (
     uuid uuid NOT NULL,
     url text NOT NULL,
     container_name text NOT NULL,
+	json_path text NOT NULL,
```
ggallen, line 91: This should not be NOT NULL. It's possible that some logging system will not need it.

@maruiz93 , take note.

pkg/database/facade/inserter.go
```diff
@@ -16,15 +16,15 @@ type DBInserter interface {
 		clusterDeletedTs sql.NullString,
 		data []byte,
 	) *sqlbuilder.InsertBuilder
-	UrlInserter(uuid, url, containerName string) *sqlbuilder.InsertBuilder
+	UrlInserter(uuid, url, containerName, jsonPath string) *sqlbuilder.InsertBuilder
```
ggallen, line 19: Need to make sure that a `null` doesn't get written to the database as the string "null".

ggallen (approved):

maruiz93 (approved):

pkg/database/facade/inserter.go
skoved: since `UrlInserter` accepts a jsonPath as a string, the closest thing to `null` that go will let you pass is empty string. I tested with the following logging config (note that there is not `LOG_URL_JSONPATH` defined)

```
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: kubearchive-logging
  namespace: kubearchive
data:
  # NOTE: CONTAINER_NAME is provided at URL generation time by KubeArchive.
  POD_ID: "cel:metadata.uid"
  LOG_URL: "https://localhost:8089/services/search/jobs/export?search=search%20%2A%20%7C%20spath%20%22kubernetes.pod_id%22%20%7C%20search%20%22kubernetes.pod_id%22%3D%22{POD_ID}%22%20%7C%20spath%20%22kubernetes.container_name%22%20%7C%20search%20%22kubernetes.container_name%22%3D%22{CONTAINER_NAME}%22%20%7C%20sort%20time%20%7C%20table%20%22message%22&output_mode=json"
```
when I look at the `log_url` table, I see that the rows have an empty string for the `json_path` column, not `"null"`

integrations/database/mariadb/kubearchive.sql
skoved: done

integrations/database/postgresql/kubearchive.sql
skoved: done

skoved:
@maruiz93 sorry for accidentally requesting a review

maruiz93 (approved):

integrations/database/mariadb/kubearchive.sql
ggallen: How about just `DEFAULT ''` rather than `NOT NULL`.  I guess it really doesn't matter too much.

But to me `NOT NULL` implies I'd better pass it a real JSON path.  And that shouldn't be required.

ggallen: And what does "done" mean here?  You didn't change anything.

skoved: I removed `NOT NULL` from the jsonPath line. I can add `DEFAULT ''` to be safe though

ggallen: No need. Sorry, I always look at the old code so I didn't see your change.

skoved:
pulled commits from main so I can merge

PR #721, opened 2024-12-19
Database.WriteUrls does not transact with invalid arguments

### LINKED ISSUE #720: `Database.WriteUrls` should not perform db operations if the k8s object is null or `[]model.LogTuple` is empty
`Database.WriteUrls` writes new log urls to the database in two steps. First it deletes all rows in the `log_url` table with the same `uuid` as `k8sObj`. Then it adds all every element of `logs` to the table. The current implementation has 2 issues

1. `k8sObj` is a `*unstructured.Unstructured` and no check for `k8sObj == nil` is done which means that we can crash on null pointer when calling `k8sObj.GetUID()`.
2. if `len(logs) == 0`, any logs already stored in the table will be deleted (maybe this should be the behavior)

These scenarios can be prevented by check for these conditions at the beginning of the method

skoved:

Resolves #720

## Release Note

```release-note
Fixed bug where KubeArchive crashed under some circumstances and fixed a bug where KubeArchive could delete all logs related to a resource.
```

## Notes for Reviewers

Databases.WriteUrls checks if k8sObj is nil If this condition is true, it returns an error and
does not perform any database transactions. sink.writeLogs checks if len(urls) == 0 and logs
that no urls were generated and does not call Database.WriteUrls when this condition is true

### DIFF
--- a/pkg/database/database.go
+++ b/pkg/database/database.go
@@ -263,13 +263,20 @@ func (db *Database) WriteResource(ctx context.Context, k8sObj *unstructured.Unst
 	return nil
 }
 
-// WriteUrls deletes urls for k8sObj before writing urls to prevent duplicates
+// WriteUrls deletes urls for k8sObj before writing urls to prevent duplicates. If logs is empty or nil all urls for
+// k8sObj will be deleted from the database and will not be replaced
 func (db *Database) WriteUrls(
 	ctx context.Context,
 	k8sObj *unstructured.Unstructured,
 	jsonPath string,
 	logs ...models.LogTuple,
 ) error {
+	// The sink performs checks before WriteUrls is called, which currently make it not possible for this check to
+	// evaluate to true during normal program execution. This check is here purely as a safeguard.
+	if k8sObj == nil {
+		return errors.New("Cannot write log urls to the database when k8sObj is nil")
+	}
+
 	tx, err := db.DB.BeginTx(ctx, nil)
 	if err != nil {
 		return fmt.Errorf(
--- a/cmd/sink/main.go
+++ b/cmd/sink/main.go
@@ -91,6 +91,10 @@ func (sink *Sink) writeLogs(ctx context.Context, obj *unstructured.Unstructured)
 		slog.Error("Could not build log urls for object", "objectID", obj.GetUID(), "err", err)
 		return
 	}
+	if len(urls) == 0 {
+		slog.Info("No log urls were generated for object", "objectID", obj.GetUID())
+		return
+	}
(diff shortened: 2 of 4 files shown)


mafh314 (approved):

rh-hemartin:
Hey @skoved the release notes are directly aimed towards users and are rendered in the following way: https://github.com/kubearchive/kubearchive/releases, so I would rewrite your notes as:

```
Fixed bug where KubeArchive crashed under some circumstances and fixed a bug where KubeArchive could delete all logs related to a resource.
```

pkg/database/database.go
```diff
@@ -240,6 +240,13 @@ func (db *Database) WriteUrls(
 	jsonPath string,
 	logs ...models.LogTuple,
 ) error {
+	if k8sObj == nil {
+		return errors.New("Cannot write log urls to the database when k8sObj is nil")
+	}
+	if len(logs) == 0 {
+		return errors.New("Cannot write log urls to the database when no logs are provided")
+	}
```
rh-hemartin, line 248: The check before and this one are very domain specific, not that related with the database itself, because they are preconditions to continue with the database operations. What do you think about moving them to the parent caller, `Sink.WriteUrls`? We already have a precondition there in which the creation of URLs should be successful anyway, so we keep the business preconditions together.

pkg/database/database.go
```diff
@@ -240,6 +240,13 @@ func (db *Database) WriteUrls(
 	jsonPath string,
 	logs ...models.LogTuple,
 ) error {
+	if k8sObj == nil {
+		return errors.New("Cannot write log urls to the database when k8sObj is nil")
+	}
```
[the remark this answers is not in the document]
skoved: I agree with you that based on where we call `WriteUrls`, I agree that I don't think it's possible for it be called with `k8sObj == nil`, this is more of a defensive thing in case changes are made in the future. I'll add a comment explaining that this check can't and should not return true during normal execution like you suggest

pkg/database/database.go
skoved: I think that it makes sense to move this check to the sink side. I'll add a comment to `Database.WriteUrls` to mention that if logs is empty or `nil`, any existing logs will be removed and will not be replaced. I do think that the we should keep the `k8sObj == nil` check though

skoved:
updated the commit message to match the changes made to this PR

rh-hemartin (approved):

skoved:
pulling commits from main so that I can merge this

maruiz93:
Sorry about the problems with the integration tests, seems that some timeouts need to be extended, should be fixed in https://github.com/kubearchive/kubearchive/pull/741

PR #756, opened 2025-01-17
Set sslmode as required for PostgreSQL

### LINKED ISSUE #755: Allow SSL connection for PostgreSQL
Currently we have `sslmode=disable` on our connection string to PostgreSQL. If the server wants us to use TLS we can't accept that, because of the `disable`. There are different modes explained in the PostgreSQL documentation[1] and we probably want to use `allow` which means "I don't care about security, but I will pay the overhead of encryption if the server insists on it.". This way we can accommodate for installations that want a higher level of security by encrypting the communication between the database and its clients.

**Acceptance Criteria**

* Change `sslmode` on the PostgreSQL string connection from `disable` to ~`allow`~`require`

[1]: https://www.postgresql.org/docs/current/libpq-ssl.html#LIBPQ-SSL-PROTECTION

rh-hemartin:

Resolves #755 

## Release Note

```release-note
KubeArchive now requires PostgreSQL databases to be protected with TLS. It can be with a self-signed certificate as the certificate is not verified, just required.
```

### DIFF
--- a/integrations/database/postgresql/postgres.yaml
+++ b/integrations/database/postgresql/postgres.yaml
@@ -7,6 +7,11 @@ metadata:
   name: kubearchive
 spec:
   instances: 1
+  postgresql:
+    # https://www.postgresql.org/docs/current/auth-pg-hba-conf.html
+    # columns: connection-type database user address auth-options
+    pg_hba:
+      - hostnossl  all  all  all  reject
   bootstrap:
     initdb:
       database: kubearchive
--- a/docs/modules/ROOT/pages/getting-started/installation.adoc
+++ b/docs/modules/ROOT/pages/getting-started/installation.adoc
@@ -11,7 +11,7 @@ This document explains how to install and configure KubeArchive in your Kubernet
 
 == Prerequisites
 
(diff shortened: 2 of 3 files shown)


integrations/database/postgresql/postgres.yaml
```diff
@@ -7,6 +7,9 @@ metadata:
   name: kubearchive
 spec:
   instances: 1
+  postgresql:
+    pg_hba:
+      - hostnossl  all  all  all  reject
```
mafh314, line 12: Hey @rh-hemartin could you add a comment here with something like this?: (to know which value applies to each parameter)
# DATABASE  USER  ADDRESS  METHOD

rh-hemartin: I introduced a comment, check it out.

mafh314: Thanks!

rh-hemartin:
@mafh314 approve the PR if you find it correct

maruiz93 (approved):

mafh314 (approved):

PR #770, opened 2025-01-22
move cmd/api/abort to pkg/abort

skoved:
## Release Note

```release-note
NONE
```

## Notes for Reviewers

As part of the sink refactor to use a gin http server, it would be helpful to share this code with the api server. I'm trying to break up the refactor into smaller pieces that can be considered on their own so that I don't end up with one massive PR at the end


skoved:
closing because after additional work on the refactor, it does not seem to be the case that cmd/api/abort is needed by the sink

PR #776, opened 2025-01-23
Move logging integration test to its own func

### LINKED ISSUE #768: Logging integration test should be in its own test
After https://github.com/kubearchive/kubearchive/issues/659 the integration test `TestArchiveAndRead` was extended to also test the pod logs retrieval. This generated the need of extending the timeout of the tests to make the integration tests resillient enough.

If we move this part to its own test we will probably reduce the timeout needed and will have the integration tests executed in less time while keeping everything tested.

*Acceptance Criteria*:
* Move the logging test of out of `TestArchiveAndRead` integration test and create its own.

rh-hemartin:

Resolves #768 

## Release Note

```release-note
NONE
```

## Notes for Reviewers

* Moved some tests around since files were getting very long.

### DIFF
--- a/.github/workflows/go_test_integration.yml
+++ b/.github/workflows/go_test_integration.yml
@@ -54,6 +54,7 @@ jobs:
         run: |
           export KO_DOCKER_REPO="kind.local"
           bash hack/quick-install.sh
+          bash integrations/logging/splunk/install.sh
       - name: Test Go code
         run: |
           export KO_DOCKER_REPO="kind.local"
--- a/test/integration/database_test.go
+++ b/test/integration/database_test.go
@@ -1,0 +1,134 @@
+// Copyright KubeArchive Authors
+// SPDX-License-Identifier: Apache-2.0
+//go:build integration
+
+package main
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/avast/retry-go/v4"
+	"github.com/kubearchive/kubearchive/test"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+// TestDatabaseConnection verifies the database connection retries using the Sink component.
+func TestDatabaseConnection(t *testing.T) {
+	clientset, _, err := test.GetKubernetesClient()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	dynclient, err := test.GetDynamicKubernetesClient()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Fence database to make it unavailable - https://cloudnative-pg.io/documentation/1.24/fencing/
+	clusterResource := schema.GroupVersionResource{Group: "postgresql.cnpg.io", Version: "v1", Resource: "clusters"}
+	resource, err := dynclient.Resource(clusterResource).Namespace("postgresql").Get(context.Background(), "kubearchive", metav1.GetOptions{})
+	if err != nil {
+		t.Fatal(err)
+	}
+	annotations := resource.GetAnnotations()
+	annotations["cnpg.io/fencedInstances"] = "[\"*\"]"
+	resource.SetAnnotations(annotations)
+
+	resource, err = dynclient.Resource(clusterResource).Namespace("postgresql").Update(context.Background(), resource, metav1.UpdateOptions{})
+	if err != nil {
+		t.Fatal(err)
+	}
+	t.Logf("Fenced database")
+
+	// restart sink pod - replicas = 0
+	deploymentScaleSink, err := clientset.AppsV1().Deployments("kubearchive").GetScale(context.Background(), "kubearchive-sink", metav1.GetOptions{})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	scaleSink := *deploymentScaleSink
+	scaleSink.Spec.Replicas = 0
+	usSink, err := clientset.AppsV1().Deployments("kubearchive").UpdateScale(context.Background(), "kubearchive-sink", &scaleSink, metav1.UpdateOptions{})
+	if err != nil {
+		t.Fatal(err)
+	}
+	t.Logf("Changing sink to %d replicas", scaleSink.Spec.Replicas)
+	t.Log(*usSink)
+
+	t.Logf("Waiting 5 seconds for kubearchive-sink to scale down...")
+	time.Sleep(5 * time.Second)
+
+	// restart sink pod - replicas = 1
+	deploymentScaleSink, err = clientset.AppsV1().Deployments("kubearchive").GetScale(context.Background(), "kubearchive-sink", metav1.GetOptions{})
+	if err != nil {
+		t.Fatal(err)
+	}
+	scaleSink = *deploymentScaleSink
+	scaleSink.Spec.Replicas = 1
+	usSink, err = clientset.AppsV1().Deployments("kubearchive").UpdateScale(context.Background(), "kubearchive-sink", &scaleSink, metav1.UpdateOptions{})
+	if err != nil {
+		t.Fatal(err)
+	}
+	t.Logf("Changing sink to %d replicas", scaleSink.Spec.Replicas)
+	t.Log(*usSink)
+
+	// wait to sink pod ready and generate connection retries with the database
+	t.Logf("Waiting for sink to be up, and to generate retries with the database")
+	retryErr := retry.Do(func() error {
+		logs, err := test.GetPodLogs(clientset, "kubearchive", "kubearchive-sink")
+		if err != nil {
+			return err
+		}
+
+		t.Logf("Logs\n%s", logs)
+		if strings.Contains(logs, "connection refused") {
+			return nil
+		}
+
+		return fmt.Errorf("Sink pod didn't try to connect to the database yet")
+	})
+
+	if retryErr != nil {
+		t.Fatal(retryErr)
+	}
+
+	// Unfence database
+	resource, err = dynclient.Resource(clusterResource).Namespace("postgresql").Get(context.Background(), "kubearchive", metav1.GetOptions{})
+	if err != nil {
+		t.Fatal(err)
+	}
+	annotations = resource.GetAnnotations()
+	delete(annotations, "cnpg.io/fencedInstances")
+	resource.SetAnnotations(annotations)
+
+	resource, err = dynclient.Resource(clusterResource).Namespace("postgresql").Update(context.Background(), resource, metav1.UpdateOptions{})
+	if err != nil {
+		t.Fatal(err)
+	}
+	t.Logf("Unfenced database")
+
+	retryErr = retry.Do(func() error {
+		logs, err := test.GetPodLogs(clientset, "kubearchive", "kubearchive-sink")
+		if err != nil {
+			return nil
+		}
+
+		t.Logf("Logs:\n%s", logs)
+		if strings.Contains(logs, "Successfully connected to the database") {
+			return nil
+		}
+
+		return errors.New("Sink pod didn't connect successfully to the database yet")
+	})
+
+	if retryErr != nil {
+		t.Fatal(retryErr)
+	}
+}
--- a/test/integration/logging_test.go
+++ b/test/integration/logging_test.go
@@ -1,0 +1,141 @@
+// Copyright KubeArchive Authors
+// SPDX-License-Identifier: Apache-2.0
+//go:build integration
+
+package main
+
+import (
+	"context"
+	"crypto/tls"
+	"errors"
+	"fmt"
+	"net/http"
+	"os/exec"
+	"strings"
+	"testing"
+
+	"github.com/avast/retry-go/v4"
+	"github.com/kubearchive/kubearchive/test"
+	authenticationv1 "k8s.io/api/authentication/v1"
+	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestLogging(t *testing.T) {
+	namespaceName := fmt.Sprintf("test-%s", test.RandomString())
+	clientset, _, errClient := test.GetKubernetesClient()
+	if errClient != nil {
+		t.Fatal(errClient)
+	}
+
+	pods, err := clientset.CoreV1().Pods("kubearchive").List(context.Background(), metav1.ListOptions{
+		LabelSelector: "app=kubearchive-api-server",
+		FieldSelector: "status.phase=Running",
+	})
+	fmt.Println(fmt.Sprintf("Pod to forward: %s", pods.Items[0].Name))
+	if err != nil {
+		t.Fatal(err)
+	}
+	var portForward chan struct{}
+	var errPortForward error
+	retryErr := retry.Do(func() error {
+		portForward, errPortForward = test.PortForward([]string{"8081:8081"}, pods.Items[0].Name, "kubearchive")
+		if errPortForward != nil {
+			return errPortForward
+		}
+		return nil
+	}, retry.Attempts(3))
+
+	if retryErr != nil {
+		t.Fatal(retryErr)
+	}
+
+	defer close(portForward)
+
+	_, errNamespace := clientset.CoreV1().Namespaces().Create(context.Background(), &corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: namespaceName,
+		},
+	}, metav1.CreateOptions{})
+	if errNamespace != nil {
+		t.Fatal(errNamespace)
+	}
+
+	t.Cleanup(func() {
+		errNamespace = clientset.CoreV1().Namespaces().Delete(context.Background(), namespaceName, metav1.DeleteOptions{})
+		if errNamespace != nil {
+			t.Fatal(errNamespace)
+		}
+	})
+
+	_, roleBindingErr := clientset.RbacV1().RoleBindings(namespaceName).Create(context.Background(), &rbacv1.RoleBinding{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "view-default-test",
+		},
+		Subjects: []rbacv1.Subject{
+			{
+				Kind:      "ServiceAccount",
+				Name:      "default",
+				Namespace: namespaceName,
+			},
+		},
+		RoleRef: rbacv1.RoleRef{
+			APIGroup: "rbac.authorization.k8s.io",
+			Kind:     "ClusterRole",
+			Name:     "view",
+		},
+	},
+		metav1.CreateOptions{})
+	if roleBindingErr != nil {
+		t.Fatal(roleBindingErr)
+	}
+
+	token, tokenErr := clientset.CoreV1().ServiceAccounts(namespaceName).CreateToken(context.Background(), "default", &authenticationv1.TokenRequest{}, metav1.CreateOptions{})
+	if tokenErr != nil {
+		fmt.Printf("could not create a token, %s", tokenErr)
+		t.Fatal(tokenErr)
+	}
+
+	client := http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{
+				InsecureSkipVerify: true,
+			},
+		},
+	}
+
+	// Install the log-generator.
+	namespaceCmd := fmt.Sprintf("--namespace=%s", namespaceName)
+	cmd := exec.Command("bash", "../log-generators/cronjobs/install.sh", namespaceCmd, "--num-jobs=1")
+	output, errScript := cmd.CombinedOutput()
+	if errScript != nil {
+		fmt.Println("Could not run the log-generator: ", errScript)
+		t.Fatal(errScript)
+	}
+	fmt.Println("Output: ", string(output))
+
+	url := fmt.Sprintf("https://localhost:8081/apis/batch/v1/namespaces/%s/cronjobs/generate-log-1/log", namespaceName)
+	retryErr = retry.Do(func() error {
+		body, err := test.GetLogs(&client, token.Status.Token, url)
+		if err != nil {
+			return err
+		}
+
+		if len(body) == 0 {
+			return errors.New("could not retrieve the pod log")
+		}
+		fmt.Println("Successfully retrieved logs")
+
+		bodyString := string(body)
+		if len(strings.Split(bodyString, "\n")) != 1025 {
+			return fmt.Errorf("expected 1025 lines, currently '%d'. Trying again...", len(strings.Split(bodyString, "\n")))
+		}
+
+		return nil
+	}, retry.Attempts(20))
+
+	if retryErr != nil {
+		t.Fatal(retryErr)
+	}
+}
--- a/test/main.go
+++ b/test/main.go
@@ -5,8 +5,10 @@ package test
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"log/slog"
 	"math/rand"
 	"net/http"
@@ -26,6 +28,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
 
 const (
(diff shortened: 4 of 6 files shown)


integrations/logging/splunk/cluster-output.yaml
```diff
@@ -18,3 +18,8 @@ spec:
     index: main
     format:
       type: json
+    buffer:
+      timekey: 10s
+      timekey_wait: 1s
+      retry_exponential_backoff_base: 1.5
+      retry_max_interval: 30s
```
maruiz93, line 25: Can we add some comments or links to where these parameters are explained?

test/integration/logging_test.go
```diff
@@ -0,0 +1,236 @@
+// Copyright KubeArchive Authors
+// SPDX-License-Identifier: Apache-2.0
+//go:build integration
```
maruiz93, line 3: If we are adding this test in a separate file called `logging_test` I think that we should restructure the rest of the tests, maybe something like this:
- Move `TestKubeArchiveDeployments` and `TestDatabaseConnection` to its own file and call it something like `resiliency_test.go`
- Rename `integration_test.go` to `read_archived_test.go

test/integration/logging_test.go
```diff
@@ -0,0 +1,236 @@
...
+
+	podName := list.Items[0].GetName()
+	url = fmt.Sprintf("https://localhost:8081/api/v1/namespaces/%s/pods/%s/log", namespaceName, podName)
+	request, errRequest := http.NewRequest("GET", url, nil)
+	if errRequest != nil {
+		fmt.Printf("could not create a request, %s", errRequest)
+		t.Fatal(errRequest)
+	}
+	request.Header.Set("Authorization", fmt.Sprintf("Bearer %s", token.Status.Token))
+	fmt.Printf("HTTP request: %s\n", request.URL.RequestURI())
```
maruiz93, line 203: I'd suggest moving the helper function of `getUrl()` in `integration_test.go` to `test/main.go` so it's available to every integration test (same as the PortForward function) and use it here too.

It will imply not doing the "parsing" of the data to an `Unstructured` struct in the function but I think the refactor is worth it.

test/integration/logging_test.go
```diff
@@ -0,0 +1,236 @@
...
+				corev1.Container{
+					Name:    "fedora",
+					Command: []string{"echo", "Hello World"},
+					Image:   "quay.io/fedora/fedora:latest",
+				},
+			},
+		},
+	}
+
+	_, createPodErr := clientset.CoreV1().Pods(namespaceName).Create(context.Background(), pod, metav1.CreateOptions{})
```
maruiz93, line 170: This is the simplest case of the logging retrieval, retrieving the logs of a pod, and although I think that this is a nice test to have, I miss being able to also test the more "complex" case of retrieving the logs from an owner of the pod (like we were doing with the `CronJob`  when dealing with the cronjob-generator)

Another alternative could be to create a deployment with one pod, that's simpler than the cronjob and if we ask for the logs of the deployment it would be the same as askinf for the pods of the cronjob, a two-level ownership jump.

test/integration/logging_test.go
```diff
@@ -0,0 +1,236 @@
...
+			Namespace: namespaceName,
+		},
+		Spec: corev1.PodSpec{
+			RestartPolicy: corev1.RestartPolicyNever,
+			Containers: []corev1.Container{
+				corev1.Container{
+					Name:    "fedora",
+					Command: []string{"echo", "Hello World"},
+					Image:   "quay.io/fedora/fedora:latest",
+				},
```
maruiz93, line 165: I'd suggest taking advantage of the image that we are using in the cronjob-generator script for having container with more "realistic" logs than just a one line.

```suggestion
				corev1.Container{
					Name:    "apache-log-generator",
					Command: []string{"echo", "Hello World"},
					Image:   "quay.io/kubearchive/mingrammer/flog",
					ImagePullPolicy: PullPolicy.PullIfNotPresent,
                                       Args: []string{"-n","1024","-d","5ms"}
				},
```

After that, and if we end up only having one pod from where to retrieve the logs, an additional assertion could be to test if the lines retrieved are 1024:

```
	assert.Equal(t, 1024, len(strings.Split(data, "\n"))
```

maruiz93 (changes requested):

integrations/logging/splunk/install.sh
```diff
@@ -12,7 +12,7 @@ cd ${SCRIPT_DIR}
 
 NAMESPACE="splunk-operator"
 
-kubectl apply -f https://github.com/splunk/splunk-operator/releases/download/2.6.0/splunk-operator-cluster.yaml \
+kubectl apply -f https://github.com/splunk/splunk-operator/releases/download/2.7.0/splunk-operator-cluster.yaml \
```
maruiz93, line 15: Another addition that could be done to this script is to ensure that the splunk-single-standalone is ready, that way the wait part is done here instead of on the tests.

The splunk-single-standalone is a statefulset with a [OnDelete](https://kubernetes.io/docs/tutorials/stateful-application/basic-stateful-set/#on-delete) strategy update which means that isn't covered by the `kubectl rollout status` command so I suggest to add the following line at the end of the script:

kubectl -n ${NAMESPACE} wait --for=condition=ready pod -l app.kubernetes.io/instance=splunk-single-standalone

maruiz93:
The test timed out after an hour (https://github.com/kubearchive/kubearchive/actions/runs/12932989111/job/36070521141?pr=776) so there might be some problems with the splunk deployment...

We may need to make the `splunk/install.sh` script more resilient, for example by checking if there are any errors in the fluentbit pods...

Also, once this is more fined-tune, we may be able to reduce the timeout of 1 hour to maybe 30 mins in `.github/workflows/go-test-integration.yml`

integrations/logging/splunk/cluster-output.yaml
ggallen: But there are not any comments for any of the other parameters.  Can't we just assume that anyone looking at this would be able to find the documentation for it?  Isn't that what Google is for?

integrations/logging/splunk/install.sh
ggallen: > The splunk-single-standalone is a statefulset with a OnDelete strategy update which means that isn't covered by the kubectl rollout status command

Where do you see this? Are we seeing issues where the script is done but the logging system isn't ready?  I ran this many times on my laptop tying to insure that when the script ends it is ready.

integrations/logging/splunk/cluster-output.yaml
maruiz93: well, I just asked because they are optional parameters that we are modifying to get a specific behavior in our integration tests. Of course, we can google the information, but it's really helpful having it written down and a reference for knowing why they shouldn't be changed.

For example, this happened to me when I was trying to understand the "fencing" that we do the database in one of our integration tests, it's really nice but for someone that is not familiar with it (like me) I had to google it. Next time someone sees the same for the first time or after some time, the same question will arise, so, if having a link makes things easier why not adding them?
BTW, I added a link for this in https://github.com/kubearchive/kubearchive/blob/main/test/integration/integration_test.go#L91-L91

The following will be enough for me:
```suggestion
     # Flush and Write times of the logging operator need to be shorter for the integration tests
     # https://kube-logging.dev/4.7/docs/configuration/plugins/outputs/buffer/?highlight=timekey#buffer-timekey
      timekey: 10s 
      timekey_wait: 1s
      retry_exponential_backoff_base: 1.5
      retry_max_interval: 30s
```

integrations/logging/splunk/install.sh
maruiz93: > Are we seeing issues where the script is done but the logging system isn't ready?
Yes, it can be seen in any of the integration test execution (e.g. https://github.com/kubearchive/kubearchive/actions/runs/12932989111/job/36070521141?pr=776)

```
HTTP request: /api/v1/namespaces/test-kwdgdznk/pods/fedora/log
HTTP body: {"message":"Get \"https://splunk-single-standalone-service.splunk-operator.svc.cluster.local:8089/services/search/jobs/export?search=search%20%2A%20%7C%20spath%20%22kubernetes.pod_id%22%20%7C%20search%20%22kubernetes.pod_id%22%3D%229d754de2-cc48-49cd-9681-70ea94a49484%22%20%7C%20spath%20%22kubernetes.container_name%22%20%7C%20search%20%22kubernetes.container_name%22%3D%22fedora%22%20%7C%20sort%20time%20%7C%20table%20%22message%22\u0026output_mode=json\": dial tcp 10.96.181.162:8089: connect: connection refused"}
The HTTP status returned is not OK, 500 Internal Server Error
HTTP body: {"message":"Get \"https://splunk-single-standalone-service.splunk-operator.svc.cluster.local:8089/services/search/jobs/export?search=search%20%2A%20%7C%20spath%20%22kubernetes.pod_id%22%20%7C%20search%20%22kubernetes.pod_id%22%3D%229d754de2-cc48-49cd-9681-70ea94a49484%22%20%7C%20spath%20%22kubernetes.container_name%22%20%7C%20search%20%22kubernetes.container_name%22%3D%22fedora%22%20%7C%20sort%20time%20%7C%20table%20%22message%22\u0026output_mode=json\": dial tcp 10.96.181.162:8089: connect: connection refused"}
The HTTP status returned is not OK, 500 Internal Server Error
HTTP body: {"message":"Get \"https://splunk-single-standalone-service.splunk-operator.svc.cluster.local:8089/services/search/jobs/export?search=search%20%2A%20%7C%20spath%20%22kubernetes.pod_id%22%20%7C%20search%20%22kubernetes.pod_id%22%3D%229d754de2-cc48-49cd-9681-70ea94a49484%22%20%7C%20spath%20%22kubernetes.container_name%22%20%7C%20search%20%22kubernetes.container_name%22%3D%22fedora%22%20%7C%20sort%20time%20%7C%20table%20%22message%22\u0026output_mode=json\": dial tcp 10.96.181.162:8089: connect: connection refused"}
The HTTP status returned is not OK, 500 Internal Server Error
HTTP body: {"message":"Get \"https://splunk-single-standalone-service.splunk-operator.svc.cluster.local:8089/services/search/jobs/export?search=search%20%2A%20%7C%20spath%20%22kubernetes.pod_id%22%20%7C%20search%20%22kubernetes.pod_id%22%3D%229d754de2-cc48-49cd-9681-70ea94a49484%22%20%7C%20spath%20%22kubernetes.container_name%22%20%7C%20search%20%22kubernetes.container_name%22%3D%22fedora%22%20%7C%20sort%20time%20%7C%20table%20%22message%22\u0026output_mode=json\": dial tcp 10.96.181.162:8089: connect: connection refused"}
The HTTP status returned is not OK, 500 Internal Server Error
HTTP body: {"message":"Get \"https://splunk-single-standalone-service.splunk-operator.svc.cluster.local:8089/services/search/jobs/export?search=search%20%2A%20%7C%20spath%20%22kubernetes.pod_id%22%20%7C%20search%20%22kubernetes.pod_id%22%3D%229d754de2-cc48-49cd-9681-70ea94a49484%22%20%7C%20spath%20%22kubernetes.container_name%22%20%7C%20search%20%22kubernetes.container_name%22%3D%22fedora%22%20%7C%20sort%20time%20%7C%20table%20%22message%22\u0026output_mode=json\": dial tcp 10.96.181.162:8089: connect: connection refused"}
The HTTP status returned is not OK, 500 Internal Server Error
HTTP body: {"message":"Get \"https://splunk-single-standalone-service.splunk-operator.svc.cluster.local:8089/services/search/jobs/export?search=search%20%2A%20%7C%20spath%20%22kubernetes.pod_id%22%20%7C%20search%20%22kubernetes.pod_id%22%3D%229d754de2-cc48-49cd-9681-70ea94a49484%22%20%7C%20spath%20%22kubernetes.container_name%22%20%7C%20search%20%22kubernetes.container_name%22%3D%22fedora%22%20%7C%20sort%20time%20%7C%20table%20%22message%22\u0026output_mode=json\": dial tcp 10.96.181.162:8089: connect: connection refused"}
The HTTP status returned is not OK, 500 Internal Server Error
HTTP body: {"message":"Get \"https://splunk-single-standalone-service.splunk-operator.svc.cluster.local:8089/services/search/jobs/export?search=search%20%2A%20%7C%20spath%20%22kubernetes.pod_id%22%20%7C%20search%20%22kubernetes.pod_id%22%3D%229d754de2-cc48-49cd-9681-70ea94a49484%22%20%7C%20spath%20%22kubernetes.container_name%22%20%7C%20search%20%22kubernetes.container_name%22%3D%22fedora%22%20%7C%20sort%20time%20%7C%20table%20%22message%22\u0026output_mode=json\": dial tcp 10.96.181.162:8089: connect: connection refused"}
The HTTP status returned is not OK, 500 Internal Server Error
HTTP body: {"message":"Get \"https://splunk-single-standalone-service.splunk-operator.svc.cluster.local:8089/services/search/jobs/export?search=search%20%2A%20%7C%20spath%20%22kubernetes.pod_id%22%20%7C%20search%20%22kubernetes.pod_id%22%3D%229d754de2-cc48-49cd-9681-70ea94a49484%22%20%7C%20spath%20%22kubernetes.container_name%22%20%7C%20search%20%22kubernetes.container_name%22%3D%22fedora%22%20%7C%20sort%20time%20%7C%20table%20%22message%22\u0026output_mode=json\": dial tcp 10.96.181.162:8089: connect: connection refused"}
The HTTP status returned is not OK, 500 Internal Server Error
HTTP body: {"message":"Get \"https://splunk-single-standalone-service.splunk-operator.svc.cluster.local:8089/services/search/jobs/export?search=search%20%2A%20%7C%20spath%20%22kubernetes.pod_id%22%20%7C%20search%20%22kubernetes.pod_id%22%3D%229d754de2-cc48-49cd-9681-70ea94a49484%22%20%7C%20spath%20%22kubernetes.container_name%22%20%7C%20search%20%22kubernetes.container_name%22%3D%22fedora%22%20%7C%20sort%20time%20%7C%20table%20%22message%22\u0026output_mode=json\": dial tcp 10.96.181.162:8089: connect: connection refused"}
The HTTP status returned is not OK, 500 Internal Server Error
HTTP body: {"message":"Get \"https://splunk-single-standalone-service.splunk-operator.svc.cluster.local:8089/services/search/jobs/export?search=search%20%2A%20%7C%20spath%20%22kubernetes.pod_id%22%20%7C%20search%20%22kubernetes.pod_id%22%3D%229d754de2-cc48-49cd-9681-70ea94a49484%22%20%7C%20spath%20%22kubernetes.container_name%22%20%7C%20search%20%22kubernetes.container_name%22%3D%22fedora%22%20%7C%20sort%20time%20%7C%20table%20%22message%22\u0026output_mode=json\": dial tcp 10.96.181.162:8089: connect: connection refused"}
The HTTP status returned is not OK, 500 Internal Server Error
HTTP body: {"message":"no parsed logs for the json path: $.result.message from logs in https://splunk-single-standalone-service.splunk-operator.svc.cluster.local:8089/services/search/jobs/export?search=search%20%2A%20%7C%20spath%20%22kubernetes.pod_id%22%20%7C%20search%20%22kubernetes.pod_id%22%3D%229d754de2-cc48-49cd-9681-70ea94a49484%22%20%7C%20spath%20%22kubernetes.container_name%22%20%7C%20search%20%22kubernetes.container_name%22%3D%22fedora%22%20%7C%20sort%20time%20%7C%20table%20%22message%22\u0026output_mode=json"}
The HTTP status returned is not OK, 404 Not Found
HTTP body: {"message":"no parsed logs for the json path: $.result.message from logs in https://splunk-single-standalone-service.splunk-operator.svc.cluster.local:8089/services/search/jobs/export?search=search%20%2A%20%7C%20spath%20%22kubernetes.pod_id%22%20%7C%20search%20%22kubernetes.pod_id%22%3D%229d754de2-cc48-49cd-9681-70ea94a49484%22%20%7C%20spath%20%22kubernetes.container_name%22%20%7C%20search%20%22kubernetes.container_name%22%3D%22fedora%22%20%7C%20sort%20time%20%7C%20table%20%22message%22\u0026output_mode=json"}
The HTTP status returned is not OK, 404 Not Found
HTTP body: {"message":"no parsed logs for the json path: $.result.message from logs in https://splunk-single-standalone-service.splunk-operator.svc.cluster.local:8089/services/search/jobs/export?search=search%20%2A%20%7C%20spath%20%22kubernetes.pod_id%22%20%7C%20search%20%22kubernetes.pod_id%22%3D%229d754de2-cc48-49cd-9681-70ea94a49484%22%20%7C%20spath%20%22kubernetes.container_name%22%20%7C%20search%20%22kubernetes.container_name%22%3D%22fedora%22%20%7C%20sort%20time%20%7C%20table%20%22message%22\u0026output_mode=json"}
The HTTP status returned is not OK, 404 Not Found
HTTP body: {"message":"no parsed logs for the json path: $.result.message from logs in https://splunk-single-standalone-service.splunk-operator.svc.cluster.local:8089/services/search/jobs/export?search=search%20%2A%20%7C%20spath%20%22kubernetes.pod_id%22%20%7C%20search%20%22kubernetes.pod_id%22%3D%229d754de2-cc48-49cd-9681-70ea94a49484%22%20%7C%20spath%20%22kubernetes.container_name%22%20%7C%20search%20%22kubernetes.container_name%22%3D%22fedora%22%20%7C%20sort%20time%20%7C%20table%20%22message%22\u0026output_mode=json"}
```

In this case, the first 11 retries return a 500 error because the splunk API isn't available.
You can check other executions or run the test locally to reproduce the error.

integrations/logging/splunk/cluster-output.yaml
ggallen: We've had documentation discussions before and I think we both know how each of us feels about it.  I don't care that much, we can add it.

But why not add these as well?

https://kube-logging.dev/4.7/docs/configuration/plugins/outputs/buffer/?highlight=timekey#buffer-retry_exponential_backoff_base
https://kube-logging.dev/4.7/docs/configuration/plugins/outputs/buffer/?highlight=timekey#buffer-retry_max_interval

We can add them all.

I don't see any difference between adding these and adding the ones for the logging test. Each of these options is there for some reason. Removing them may have adverse effects.

We could add links and pointers for everything. Eventually they become noise and possibly out-of-date. Where do you draw the line?

maruiz93: I guess I'm driven by the "feeling" of thinking that I will need this information in the future and I will appreciate having it written in some place and it's easier to find here than elsewhere.

For the "why not adding all the links" once you have one link it's the same place for the others. We can just put https://kube-logging.dev/4.7/docs/configuration/plugins/outputs/buffer/

and it's good enough for me

test/integration/logging_test.go
rh-hemartin: Thanks, I will switch to this one.

test/integration/logging_test.go
rh-hemartin: I agree, I'm experimenting now to make Splunk go faster, then I will switch the approach to some resource with children. Maybe a Job will be better, so we don't need to stop it before retrieving logs.

test/integration/logging_test.go
rh-hemartin: I'll see what I can do.

test/integration/logging_test.go
rh-hemartin: I'll see what I can do.

test/integration/logging_test.go
rh-hemartin: I changed to use the cron generator

test/integration/logging_test.go
rh-hemartin: I moved the function, but I didn't change the implementation. Most calls expect to parse the body into a unstructured list, so I leave it as is. The logging test has to implement its own thing to check the logs.

integrations/logging/splunk/install.sh
rh-hemartin: I added a wait for all pods, 180s.

test/integration/logging_test.go
```diff
@@ -0,0 +1,236 @@
...
+			fmt.Printf("The HTTP status returned is not OK, %s\n", response.Status)
+			return errors.New(response.Status)
+		}
+
+		if len(body) == 0 {
+			return errors.New("could not retrieve the pod log")
+		}
+		fmt.Println("successfully retrieved logs")
+		return nil
+	}, retry.Attempts(20))
```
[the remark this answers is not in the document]
rh-hemartin: I think it gets close to 10, as the first 5 or 6 are very close retries. So I'm leaving it as 20.

test/integration/logging_test.go
rh-hemartin: I divided the tests into files but not by test case, even if that matches now. The idea is that each file contains a set of related tests. I left the "TestReadAndArchive" into the current file and renamed it "TestNormalOperation".

test/main.go
```diff
@@ -205,3 +208,38 @@ func GetDynamicKubernetesClient() (*dynamic.DynamicClient, error) {
...
+	}
+
+	var data unstructured.UnstructuredList
+	err = json.Unmarshal(body, &data)
+	if err != nil {
+		fmt.Printf("Couldn't unmarshal JSON, %s\n", err)
+		return nil, err
+	}
+	fmt.Printf("The HTTP status returned is OK, %s, returned %d items\n", response.Status, len(data.Items))
+	return &data, nil
```
maruiz93, line 244: If this is extracted to a separate function, then `GetUrl` could be used in the `logging_test.go` to make it more compact, readable and easy to maintain

test/integration/logging_test.go
```diff
@@ -0,0 +1,166 @@
...
+			return errReadAll
+		}
+
+		response.Body.Close()
+
+		if response.StatusCode != http.StatusOK {
+			fmt.Printf("HTTP body: %s\n", body)
+			fmt.Printf("The HTTP status returned is not OK, %s\n", response.Status)
+			return errors.New(response.Status)
+		}
```
maruiz93, line 148: this could be replaced with a call to `test.GetUrl` if the part of parsing the body to a list of `Unstructured` instances is moved out.
See the other comment in the review

.github/workflows/go_test_integration.yml
```diff
@@ -54,6 +54,7 @@ jobs:
         run: |
           export KO_DOCKER_REPO="kind.local"
           bash hack/quick-install.sh
+          bash integrations/logging/splunk/install.sh
```
maruiz93, line 57: I agree with this change, makes sense that splunk is already installed before running the tests but we miss a thing to be tested in this change, that the logging configuration changes are correctly loaded after restarting the sink/api.

I'd add two more tests to cover this (they don't need to be added in this PR, it's ok to do them in a different one):
- Modify the logging-credentials, restart the api, check that the request doesn't have permissions and set it correctly again (similar to the test where we fence the database)
- Modify the logging-configuration, restart the sink, see how the URL has been changed and doesn't retrieve what's expected and reconfigure it.

maruiz93 (commented):
I left a suggestion for refactor and a suggestion for new tests to be added (it's ok to do them in a separate PR)

test/main.go
rh-hemartin: I created `getUrl` which has the common functionality, take a look and let me know how you find it.

test/integration/logging_test.go
rh-hemartin: Related to the abocve

test/main.go
maruiz93: it looks good!

maruiz93 (approved):

test/integration/database_test.go
```diff
@@ -0,0 +1,134 @@
...
+		if err != nil {
+			return err
+		}
+
+		t.Logf("Logs\n%s", logs)
+		if strings.Contains(logs, "connection refused") {
+			return nil
+		}
+
+		return fmt.Errorf("Pod didn't try to connect to the database yet")
```
mafh314, line 95: Hey @rh-hemartin I have a small petition here, Could you add in this error that the Pod is the sink Pod?.
Thanks!

test/integration/database_test.go
```diff
@@ -0,0 +1,134 @@
...
+	usSink, err = clientset.AppsV1().Deployments("kubearchive").UpdateScale(context.Background(), "kubearchive-sink", &scaleSink, metav1.UpdateOptions{})
+	if err != nil {
+		t.Fatal(err)
+	}
+	t.Logf("Changing sink to %d replicas", scaleSink.Spec.Replicas)
+	t.Log(*usSink)
+
+	// wait to sink pod ready and generate connection retries with the database
+	t.Logf("Waiting for sink to be up, and to generate retries with the database")
+	retryErr := retry.Do(func() error {
```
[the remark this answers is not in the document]
rh-hemartin: I don't think we need to specify default values, that defeats their purpose.

mafh314:
@rh-hemartin I ran the tests locally and I'm seeing these warnings:
`
=== RUN   TestDatabaseConnection
time=2025-02-10T16:01:27.335Z level=ERROR msg="Could not enable log url creation" error="Logging Config is empty. To configure logging update the kubearchive-logging ConfigMap"
        [GIN-debug] [WARNING] Running in "debug" mode. Switch to "release" mode in production.
         - using env:	export GIN_MODE=release
         - using code:	gin.SetMode(gin.ReleaseMode)
        
        [GIN-debug] POST   /                         --> github.com/kubearchive/kubearchive/cmd/sink/routers.(*Controller).CloudEventsHandler-fm (4 handlers)
        time=2025-02-10T16:01:27.342Z level=INFO msg="Received watch event of type. Updating filters" "event type"=ADDED
--- PASS: TestDatabaseConnection (7.75s)
`

`
=== RUN   TestLogging
Pod to forward: kubearchive-api-server-5c59b956d4-dlqnr
Forwarding from 127.0.0.1:8081 -> 8081
Forwarding from [::1]:8081 -> 8081

Output:  Warning: resource namespaces/test-phfakxkj is missing the kubectl.kubernetes.io/last-applied-configuration annotation which is required by kubectl apply. kubectl apply should only be used on resources created declaratively by either kubectl create --save-config or kubectl apply. The missing annotation will be patched automatically.
namespace/test-phfakxkj configured
`

rh-hemartin:
> @rh-hemartin I ran the tests locally and I'm seeing these warnings:

The first one is because there is no `GIN_MODE` env var for the sink and it defaults to debug. In another PR I introduced it with the value "release" like the API has. That should make the first go away.

The second warning comes from the log generator cronjob installation script. It tries to create the namespace in which it will be installed, but it is already created by the test, so it yields that warning on `kubectl apply`. So its also safe.

mafh314 (approved):

PR #788, opened 2025-01-28
Make Splunk faster for integration tests

rh-hemartin:

Resolves #

## Release Note

```release-note
NONE
```

## Notes for Reviewers

* Split this from !776, to separate better the concerns.


integrations/logging/splunk/install.sh
```diff
@@ -12,6 +12,9 @@ cd ${SCRIPT_DIR}
 
 NAMESPACE="splunk-operator"
 
+set -o errexit
+set -o xtrace
```
maruiz93, line 16: I you are adding this here, I guess that you'd want to add the other changes of this file in https://github.com/kubearchive/kubearchive/pull/776/files#diff-993791378b7246b0c4d06335d979f7e5b3b9004bba12f6bda04b224bb4cc2cc6, right?:

```
kubectl wait -n ${NAMESPACE} pod --all --for=condition=ready --timeout=90s
```

.github/workflows/go_test_integration.yml
```diff
@@ -57,4 +57,4 @@ jobs:
       - name: Test Go code
         run: |
           export KO_DOCKER_REPO="kind.local"
-          go test -count=1 -v ./test/integration -tags=integration -timeout 60m
+          go test -count=1 -v ./test/integration -tags=integration
```
maruiz93, line 60: Be careful with this change. This can't be merged until we make sure the integration tests run under 10 minutes each one

maruiz93 (commented):
As we currently don't have an integration test suite that runs under 10 minutes each test, I'd suggest adding a label stating that this PR is on hold

.github/workflows/go_test_integration.yml
rh-hemartin: How many executions do you need to be comfortable merging this?

integrations/logging/splunk/install.sh
rh-hemartin: I'm not sure, I didn't want to introduce a bunch of changes... let me try this and lets see.

.github/workflows/go_test_integration.yml
maruiz93: I wasn't thinking on an amount of executions but rather having addressed #776 before

integrations/logging/splunk/install.sh
rh-hemartin: I increased it to 180s as it failed a couple of times with 90s.

maruiz93 (approved):

PR #771, opened 2025-01-22
refactor sink to use gin http server

skoved:
## Which Issue(s) this Pull Request Resolves

Resolves #556, resolves #521

## Release Note

```release-note
NONE
```

## Notes for Reviewers

Refactor the sink to use a gin http server instead of the default http server provided by the CloudEvents SDK. This will allow the sink to provide endpoints for Kubernetes container probes. I've tried to mirror the structure of the api server as close as is reasonable to make it easier to maintain

This PR is large but most of the change is moving the logic for the `sink` struct from `cmd/sink/main.go` to the `controller` struct in `cmd/sink/routers/routers.go` as well as a mock `Filters` (add associated tests for the mock) to make writing tests for `routers` easier.

Screenshot of an otel trace before the refactor
image

Screenshot of an otel trace after the refactor
image

Issue #306, opened 2024-08-01
Create health check endpoint for the sink


maruiz93:
The sink should provide readiness and liveness probes

skoved:
I have 2 concerns:

1. I wonder if it makes sense for the health check to only check the database connection. For instance, if the pod for the database goes down, should the sink get restarted as well? Wouldn't the better strategy be to have it wait for the database to come back up?

2. Since the main thread of the sink has sit in a receive loop waiting for CloudEvents, I think that the health check endpoint will need to be run on a separate go routine. This is not a big problem, but it'll have to figure out the logistics for sharing the database connection.

rh-hemartin:
>     1. I wonder if it makes sense for the health check to only check the database connection. For instance, if the pod for the database goes down, should the sink get restarted as well? Wouldn't the better strategy be to have it wait for the database to come back up?

I think you have a point here, maybe we shouldn't check the database inside `livenessProbe` . It does make sense in a `readinessProbe` so the pod is not marked as started until it gets connected to the database. I think we need to think about scenarios in which a restart could make the sink healthy again... The only thing that comes to mind is the memory usage gets too high and the application gets stuck because of that, but that can be handled directly by Kubernetes with limits.

>     2. Since the main thread of the sink has sit in a receive loop waiting for CloudEvents, I think that the health check endpoint will need to be run on a separate go routine. This is not a big problem, but it'll have to figure out the logistics for sharing the database connection.

I think its time to move to a custom server instead of using the default that comes with cloud events. At the end of the day its "just" a POST in a certain endpoint and that endpoint deserializes the cloudmessage. For consistency we could use Gin. Take a look at this example https://github.com/cloudevents/sdk-go/blob/main/samples/http/receiver-gin/main.go and give it a go. Then we could just add a new handler to the server that handles the health check.

skoved:
> I think you have a point here, maybe we shouldn't check the database inside livenessProbe . It does make sense in a readinessProbe so the pod is not marked as started until it gets connected to the database. 

I'll look into the `readinessProbe` more. On the application side, it looks similar. You just provide an endpoint for kubenetes to send a request to.

> I think we need to think about scenarios in which a restart could make the sink healthy again... The only thing that comes to mind is the memory usage gets too high and the application gets stuck because of that, but that can be handled directly by Kubernetes with limits.

I need to think more about this as well. One thing that comes to mind is that maybe we should have an expectation that the sink will respond to an HTTP request within a certain amount of time?

[...]

PR #802, opened 2025-02-04
move shared router code to pkg

skoved:

Related to #306

## Release Note

```release-note
NONE
```

## Notes for Reviewers

This moves code that was originally in the api and moves it into pkg so that it can be shared with the sink. This helps with the work on #306. I broke this off of the main bulk of the work for #306 because the rest depends of that work depends on #771


rh-hemartin (approved):

skoved:
Since the PR for adding the kubernetes probe endpoints was merged first I updated this PR

maruiz93 (approved):

PR #805, opened 2025-02-06
add liveness and readyness endpoints to the sink

### LINKED ISSUE #306: Create health check endpoint for the sink

skoved:

Resolves #306 

## Release Note

```release-note
NONE
```

## Notes for Reviewers

The readiness endpoint checks if the sink can connect to the database AND the Kubernetes API. Since the dynamic client does not have any kind of connection check method, the sink instead tries to list the ConfigMaps in the namespace it is deployed in (which it should always have the permissions to do)

### DIFF
--- a/cmd/sink/routers/routers.go
+++ b/cmd/sink/routers/routers.go
@@ -7,6 +7,8 @@ import (
 	"context"
 	"fmt"
 	"log/slog"
+	"net/http"
+	"os"
 	"strings"
 	"time"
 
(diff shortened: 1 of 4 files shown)


cmd/sink/routers/routers.go
```diff
@@ -189,3 +196,44 @@ func (c *Controller) receiveCloudEvent(ctx context.Context, event cloudevents.Ev
...
+			Kind:       "ConfigMap",
+		},
+	}
+	resource, _ := meta.UnsafeGuessKindToResource(cm.GroupVersionKind())
+	_, err = c.K8sClient.Resource(resource).Namespace(ns).List(ctx.Request.Context(), metav1.ListOptions{})
+	if err != nil {
+		abort.Abort(ctx, err, http.StatusServiceUnavailable)
+	}
+	ctx.JSON(http.StatusOK, gin.H{"message": "ready"})
+}
```
maruiz93, line 239: This readiness probe is more complete that the one in the api.
As it checks not only db ping but also the k8s API connectivity, we could extract this to the `pkg` to be reused,
WDYT?

skoved:
rebasing against main

cmd/sink/routers/routers.go
maruiz93: that's a good point, then we can leave it as is :)

maruiz93 (approved):

cmd/sink/routers/routers.go
```diff
@@ -189,3 +196,44 @@ func (c *Controller) receiveCloudEvent(ctx context.Context, event cloudevents.Ev
...
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "v1",
+			Kind:       "ConfigMap",
+		},
+	}
+	resource, _ := meta.UnsafeGuessKindToResource(cm.GroupVersionKind())
+	_, err = c.K8sClient.Resource(resource).Namespace(ns).List(ctx.Request.Context(), metav1.ListOptions{})
+	if err != nil {
+		abort.Abort(ctx, err, http.StatusServiceUnavailable)
+	}
```
rh-hemartin, line 237: Why we check that we have permissions to list ConfigMaps? I guess this may change at any time, but wouldn't be easier to check for permissions at startup instead?

skoved: I guess it kind of is a check on the permissions, but what I'm trying to check is that the sink can successfully connect to the k8s api, so that kubernetes will change the ready status of the sink if it cannot

rh-hemartin (approved):

skoved:
rebasing against main so the PR can merge

PR #808, opened 2025-02-07
Add database integration docs

### LINKED ISSUE #806: Update the version of postgresql for dev and testing
Amazon RDS stopped supporting PostgreSQL 16.2 in favor of 16.6.
Update the version of the postgresql version used for testing and development so it matches this new minimum.

### LINKED ISSUE #807: Add docs for the database integration
In the `Integrations` section of the KubeArchive docs there is documentation for the Logging Integration but we lack docs for the database integration.

### Acceptance Criteria:
* Include a page with the information about the database integration:
  * An overview
  * Configuration and Customization
  * How to add integrations with new databases

maruiz93:

Resolves #806, resolves #807 

## Release Note

```release-note
The documentation now contains a page that details the integration of KubeArchive with databases.
```

## Notes for Reviewers

The `kubearchive.sql` was modified by `pg_dump`.
I followed the instructions in https://github.com/kubearchive/kubearchive/blob/main/integrations/database/postgresql/README.md

### DIFF
--- a/docs/modules/ROOT/pages/integrations/database.adoc
+++ b/docs/modules/ROOT/pages/integrations/database.adoc
@@ -1,0 +1,197 @@
+= Database Integrations
+
+ifeval::["{page-component-display-version}" == "main"]
+:download-path: https://github.com/kubearchive/kubearchive/releases/latest/download
+endif::[]
+ifeval::["{page-component-display-version}" != "main"]
+:download-path: https://github.com/kubearchive/kubearchive/releases/download/{page-component-display-version}
+endif::[]
+
+== Overview
+
+KubeArchive requires a database to store resources.
+This document list the databases supported, how to set up the database schema
+and how to add new database integrations.
+
+=== PostgreSQL
+
+Currently, PostgreSQL is the only fully supported database engine.
+The KubeArchive integration test suite runs with the PostgreSQL version detailed in
+link:{download-path}/kubearchive.sql[the kubearchive.sql file].
+
+The PostgreSQL implementation is available
+link:https://github.com/kubearchive/kubearchive/blob/{page-component-display-version}/pkg/database/postgresql.go[here].
+
+The PostgreSQL schema is available
+link:https://github.com/kubearchive/kubearchive/blob/{page-component-display-version}/integrations/database/postgresql/kubearchive.sql[here]
+and in the
+link:{download-path}[release]
+as [filename]`kubearchive.sql`.
+
+=== MariaDB
+
+This integration is still in progress.
+
+The MariaDB implementation is available
+link:https://github.com/kubearchive/kubearchive/blob/{page-component-display-version}/pkg/database/mariadb.go[here].
+
+[#_configuration_and_customization]
+== Configuration and Customization
+
+=== Schema
+
+KubeArchive offers a [filename]`kubearchive.sql` file with the instructions to set up
+the schema as part of the installation process.
+
+[IMPORTANT]
+====
+The provided schema includes indexes.
+
+KubeArchive users should review and customize the indexes
+if needed based on the expected queries and the amount and
+distribution of the expected archived data.
+====
+
+The schema includes the creation of the database `kubearchive` owned by the `kubearchive` user.
+The schema expects the existence of a `kubearchive` database user fow owning this database.
+
+[IMPORTANT]
+====
+An account with admin privileges should run [filename]`kubearchive.sql`.
+
+[source, bash]
+----
+psql -U admin \ <1>
+-h database.example.com \ <2>
+-f kubearchive.sql
+----
+<1> The database admin account used to apply the schema. It can be different from the KubeArchive DB user.
+<2> The database URL
+====
+
+=== Database Credentials
+
+The `kubearchive-database-credentials` Secret stores the information to connect KubeArchive with the Database.
+
+The default content of this Secret when KubeArchive is installed is:
+
+[source, yaml]
+----
+kind: Secret
+type: Opaque
+metadata:
+  name: kubearchive-database-credentials
+  namespace: kubearchive
+stringData:
+  DATABASE_KIND: postgresql
+  DATABASE_PORT: "5432"
+  DATABASE_URL: "kubearchive-rw.postgresql.svc.cluster.local"
+  DATABASE_DB: "kubearchive"
+  DATABASE_USER: "kubearchive"
+  DATABASE_PASSWORD: "Databas3Passw0rd"  # notsecret
+----
+
+Update the secret with the specific values of your database and
+restart the pods accordingly to pick the new values.
+The command for changing the most common values, URL and password, is:
+
+[source, bash]
+----
+kubectl patch secret -n kubearchive kubearchive-database-credentials \
+--patch='{"stringData": {
+"DATABASE_URL": "database.example.com", <1>
+"DATABASE_PASSWORD": "password" <2>
+}}'
+----
+<1> The database URL
+<2> The database password
+
+[WARNING]
+====
+If you change the `DATABASE_USER` or the `DATABASE_DB` then you should accordingly update
+the schema provided within the KubeArchive release.
+====
+
+[IMPORTANT]
+====
+Don't forget to restart kubearchive sink and api as both access the database:
+
+[source, bash]
+----
+kubectl rollout -n kubearchive restart deployment kubearchive-sink kubearchive-api-server
+----
+====
+
+
+== Adding a New Database
+
+To add a new database integration follow the instructions:
+
+1. Create a new file under `pkg/database` named after the new database engine.
+2. Include an
+link:https://go.dev/doc/effective_go#init[init function]
+that:
+
+* Inserts in the
+link:https://github.com/kubearchive/kubearchive/blob/{page-component-display-version}/pkg/database/database.go#L27[`RegisteredDBCreators`]
+map an entry keyed with the name of the DB type with a
+link:https://github.com/kubearchive/kubearchive/blob/{page-component-display-version}/pkg/database/database.go#L24[`newDBCreatorFunc`]
+implementation as value.
+
+* Inserts in the
+link:https://github.com/kubearchive/kubearchive/blob/{page-component-display-version}/pkg/database/database.go#L26[`RegisteredDatabases`]
+map an entry keyed with the name of the DB type with a
+link:https://github.com/kubearchive/kubearchive/blob/{page-component-display-version}/pkg/database/database.go#L23[`newDatabaseFunc`]
+as value.
+
+=== The `newDBCreatorFunc`
+
+This function receives the set of environment variables with the database configuration and returns a
+link:https://github.com/kubearchive/kubearchive/blob/{page-component-display-version}/pkg/database/facade/creator.go[`DBCreator`]
+implementation.
+
+The `DBCreator` should include `env` as a parameter so the `GetConnectionString` can access them.
+
+=== The `DBInterface` implementation returned by the `newDatabaseFunc`
+
+[NOTE]
+====
+This section includes the development guidelines expected by the KubeArchive maintainers.
+====
+
+The `DBInterface` should be a struct with
+link:https://github.com/kubearchive/kubearchive/blob/main/pkg/database/database.go#L40[`Database`]
+embedded.
+
+Make sure to implement the interfaces that are part of it:
+
+* link:https://github.com/kubearchive/kubearchive/blob/{page-component-display-version}/pkg/database/facade/deleter.go[DBDeleter]
+* link:https://github.com/kubearchive/kubearchive/blob/{page-component-display-version}/pkg/database/facade/filter.go[DBFilter]
+* link:https://github.com/kubearchive/kubearchive/blob/{page-component-display-version}/pkg/database/facade/inserter.go[DBInserter]
+* link:https://github.com/kubearchive/kubearchive/blob/{page-component-display-version}/pkg/database/facade/selector.go[DBSelector]
+* link:https://github.com/kubearchive/kubearchive/blob/{page-component-display-version}/pkg/database/facade/sorter.go[DBSorter]
+
+[NOTE]
+====
+All the interfaces work with objects from the
+link:https://pkg.go.dev/github.com/huandu/go-sqlbuilder[sqlbuilder Go library].
+Check out the docs to see how to work with them.
+====
+
+[TIP]
+====
+Some interfaces, like `DBDeleter`, have a default implementation.
+
+Other interfaces, like `DBFilter`, have a partial implementation.
+
+Those implementations may have the functionality that you need.
+Check them before implementing your own.
+====
+
+If the database interaction logic changes from the implementation in the `Database` struct,
+override the implementation of the affected functions.
+For example, a database that does not support upsert queries
+needs to implement `ResourceInserter` as a series of  `SELECT` and an `INSERT` statements.
+
+Take a look at the current database integrations and
+feel free to contribute to our code adding new database integrations!
--- a/docs/modules/ROOT/nav.adoc
+++ b/docs/modules/ROOT/nav.adoc
@@ -21,6 +21,7 @@
 ** xref:reference/api.adoc[]
 
 * Integrations
+** xref:integrations/database.adoc[Database Integrations]
 ** xref:integrations/logging.adoc[Logging Integrations]
 
 * Design
--- a/docs/modules/ROOT/pages/getting-started/installation.adoc
+++ b/docs/modules/ROOT/pages/getting-started/installation.adoc
@@ -17,22 +17,25 @@ This document explains how to install and configure KubeArchive in your Kubernet
 
 == Prepare the Database
 
-. Create ada user and a password for KubeArchive to use.
-. Create a database for KubeArchive, with admin permissions granted to the user you just created.
-. As the user, execute the `kubearchive.sql` we provide for initializing the database, located at:
+. Create a database user `kubearchive`.
+. With a DB admin account, apply the schema provided in the [filename]`kubearchive.sql` file
+within the KubeArchive release:
 +
 [source,bash,subs="attributes"]
 ----
 wget {download-path}/kubearchive.sql
-psql -h database.example.com -U username ... -f kubearchive.sql
+psql -h database.example.com \
+-U admin \
+-f kubearchive.sql
 ----
-+
-This file creates a PostgreSQL schema, a table for KubeArchive and other stuff KubeArchive
-requires.
+
+More information about the schema and its customization in the
+xref:integrations/database.adoc#_configuration_and_customization[Database Integration]
+section.
 
 == Configure Knative Eventing
 
-KubeArchive uses a broker to avoid missing Kubernetes resource updates. By default it is an `MTChannelBasedBroker`, but
+KubeArchive uses a broker to avoid missing Kubernetes resource updates. By default, it is an `MTChannelBasedBroker`, but
 the broker type can be changed if desired.
 
 [IMPORTANT]
@@ -42,6 +45,7 @@ link:https://knative.dev/docs/eventing/brokers/broker-types/[other broker types]
 in production environments.
 ====
 
+[_install_and_configure_kubearchive]
 == Install and Configure KubeArchive
 
 . Apply the release YAML to install KubeArchive:
@@ -53,19 +57,21 @@ kubectl apply -n kubearchive -f {download-path}/kubearchive.yaml
 
 . Provide the credentials for the database in the `kubearchive-database-credentials` secret:
 +
-[source,bash]
+[source, bash]
 ----
-cat <<EOF > /tmp/patch.yaml
-stringData:
-  DATABASE_KIND: postgresql
-  DATABASE_PORT: <postgresql-port>
-  DATABASE_URL: <postgresql-url>
-  DATABASE_USER: <kubearchive-username>
-  DATABASE_DB: <kubearchive-database>
-  DATABASE_PASSWORD: <kubearchive-password>
-EOF
-kubectl patch -n kubearchive secret kubearchive-database-credentials --patch-file /tmp/patch.yaml
+kubectl patch secret -n kubearchive kubearchive-database-credentials \
+--patch='{"stringData": {
+"DATABASE_URL": "database.example.com", <1>
+"DATABASE_PASSWORD": "password" <2>
+}}'
 ----
+<1> The database URL
+<2> The database password
+
++
+More info about this Secret in the
+xref:integrations/database.adoc#_configuration_and_customization[Database Integration]
+section.
 
 . Restart the `kubearchive-api-server` and `kubearchive-sink` Deployments to update
 the database credentials secret:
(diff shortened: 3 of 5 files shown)


rh-hemartin:
You need to add the document to a nav document

docs/modules/ROOT/pages/integrations/database.adoc
```diff
@@ -0,0 +1,69 @@
...
+====
+The provided schema includes indexes.
+
+KubeArchive users should review and customize the indexes
+if needed based on the expected queries and the amount and
+the distribution of the expected archived data.
+====
+
+Specific instructions for the configuration of the PostgreSQL Database are available in
+xref:/getting-started/configuration.adoc#_install_and_configure_kubearchive[Install and Configure KubeArchive].
```
rh-hemartin, line 31: The instructions are not in the `configuration.adoc` document, link directly to the `installation.adoc` file.

maruiz93: Right! It's fixed now

docs/modules/ROOT/pages/integrations/database.adoc
```diff
@@ -0,0 +1,69 @@
...
+KubeArchive needs a database to store the resources to make them retrievable.
+KubeArchive currently supports integration with PostgreSQL.
+MariaDB integration is in progress.
+
+[NOTE]
+====
+
+KubeArchive integration test suite runs with PostgreSQL 16.6
+
+====
```
rh-hemartin, line 14:
```suggestion
KubeArchive requires a database to store resources. This document list the databases supported, how to setup the database schema and how to configure KubeArchive to connect to the database.
```

Then I would create a section per DB: PostgreSQL, MariaDB. Each section should have the version, as you added in a NOTE below, and then the link to the SQL file.

What do you think?

docs/modules/ROOT/pages/integrations/database.adoc
```diff
@@ -0,0 +1,69 @@
...
+if needed based on the expected queries and the amount and
+the distribution of the expected archived data.
+====
+
+Specific instructions for the configuration of the PostgreSQL Database are available in
+xref:/getting-started/instllation.adoc#_install_and_configure_kubearchive[Install and Configure KubeArchive].
+
+== Support new databases
+
+KubeArchive database integration is designed to include new databases.
```
rh-hemartin, line 35: I think this phrase can go away, this idea is already implicit in the section itself.

docs/modules/ROOT/pages/integrations/database.adoc
```diff
@@ -0,0 +1,69 @@
...
+
+To add a new database integration the following interfaces need to be implemented:
```
rh-hemartin, line 37:
```suggestion
To add support for a new database, implement the following interfaces:
```

docs/modules/ROOT/pages/integrations/database.adoc
```diff
@@ -0,0 +1,69 @@
...
+* link:https://github.com/kubearchive/kubearchive/blob/main/pkg/database/facade/filter.go[DBFilter]
+* link:https://github.com/kubearchive/kubearchive/blob/main/pkg/database/facade/inserter.go[DBInserter]
+* link:https://github.com/kubearchive/kubearchive/blob/main/pkg/database/facade/selector.go[DBSelector]
+* link:https://github.com/kubearchive/kubearchive/blob/main/pkg/database/facade/sorter.go[DBSorter]
+
+[NOTE]
+====
+Except for the `DBCreator` functions, all the functions in the interfaces receive/output objects of the
+link:https://pkg.go.dev/github.com/huandu/go-sqlbuilder[sqlbuilder Go library].
+Check out the docs to see how to work with them.
```
rh-hemartin, line 50:
```suggestion
All the interfaces, with the exception of `DBCreator`, work with objects from the
link:https://pkg.go.dev/github.com/huandu/go-sqlbuilder[sqlbuilder Go library].
Check out the docs to see how to work with them.
```

docs/modules/ROOT/pages/integrations/database.adoc
```diff
@@ -0,0 +1,69 @@
...
+====
+
+[TIP]
+====
+Some of the interfaces like `DBDeleter` include a default implementation that can be used called `DBDeleterImpl`.
+
+Other interfaces like `DBFilter` include a partial implementation called `PartialDBFilterImpl` with the
+implementations of the filters that are likely to be reused.
+
+Check if those default implementations fit your database integration needs before implementing them.
```
rh-hemartin, line 60:
```suggestion
Some interfaces, like `DBDeleter`, have a default implementation. Other interfaces, like `DBFilter`, have a partial implementation. Those implementations can be useful, check them before implementing your own.
```

docs/modules/ROOT/pages/integrations/database.adoc
```diff
@@ -0,0 +1,69 @@
...
+====
+
+If more customization is needed, override the implementation of the functions declared in
+link:https://github.com/kubearchive/kubearchive/blob/main/pkg/database/database.go#L29[DBInterface].
```
rh-hemartin, line 64: Which kind of "more customization" are we talking about?

docs/modules/ROOT/pages/integrations/database.adoc
```diff
@@ -0,0 +1,69 @@
...
+
+The PostgreSQL implementation is available in
+link:https://github.com/kubearchive/kubearchive/blob/main/pkg/database/postgresql.go[here].
```
rh-hemartin, line 67: If you liked the idea of having a section per database, this should go into the PostgreSQL section

rh-hemartin:
Change the release note from:

```
A page for database integration added in the docs.
```

to

```
The documentation now contains a page that details the integration of KubeArchive with databases.
```

docs/modules/ROOT/pages/integrations/database.adoc
maruiz93: I've changed the docs to go with this approach, take a look and see if it looks good to you

docs/modules/ROOT/pages/integrations/database.adoc
maruiz93: The sentence was incorrect, I change the instructions so now there is no "except" in the sentence.

docs/modules/ROOT/pages/integrations/database.adoc
maruiz93: done!

docs/modules/ROOT/pages/integrations/database.adoc
```diff
@@ -0,0 +1,114 @@
...
+
+== Overview
+
+KubeArchive requires a database to store resources.
+This document list the databases supported, how to set up the database schema
+and how to add new database integrations.
+
+=== PostgreSQL
+
+The only currently full supported.
```
rh-hemartin, line 11:
```suggestion
Currently PostgreSQL is the only fully supported.
```

docs/modules/ROOT/pages/integrations/database.adoc
```diff
@@ -0,0 +1,114 @@
...
+KubeArchive integration test suite runs with PostgreSQL 16.6
```
rh-hemartin, line 12:
```suggestion
The KubeArchive integration test suite runs with PostgreSQL 16.6
```

docs/modules/ROOT/pages/integrations/database.adoc
```diff
@@ -0,0 +1,114 @@
...
+
+The PostgreSQL implementation is available
+link:https://github.com/kubearchive/kubearchive/blob/main/pkg/database/postgresql.go[here].
+
```
rh-hemartin, line 16: The PostgreSQL schema is available

link:https://github.com/kubearchive/kubearchive/blob/main/integrations/database/postgresql/kubearchive.sql[here]
and
link:https://github.com/kubearchive/kubearchive/releases/latest[the latest release as `kubearchive.sql`].

docs/modules/ROOT/pages/integrations/database.adoc
maruiz93: applied a similar change to remove the passive form

docs/modules/ROOT/pages/integrations/database.adoc
```diff
@@ -0,0 +1,114 @@
...
+KubeArchive users should review and customize the indexes
+if needed based on the expected queries and the amount and
+the distribution of the expected archived data.
+====
+
+Specific instructions for the configuration of the PostgreSQL Database are available in
+xref:/getting-started/instllation.adoc#_install_and_configure_kubearchive[Install and Configure KubeArchive].
+
+== Add new database integrations
```
rh-hemartin, line 41:
```suggestion
== Adding A New Database
```

docs/modules/ROOT/pages/integrations/database.adoc
```diff
@@ -0,0 +1,114 @@
...
+
+To add a new database integration follow the instructions:
+
+1. Create a new file under `pkg/database` with the name of the new database type.
+2. Include an
+link:https://go.dev/doc/effective_go#init[init function]
+that:
+
+* Inserts in the
+link:https://github.com/maruiz93/kubearchive/blob/postgresql-version-update/pkg/database/database.go#L27[`RegisteredDBCreators`]
```
rh-hemartin, line 51: Should this point to our main repository instead?

```suggestion
link:https://github.com/kubearchive/kubearchive/blob/postgresql-version-update/pkg/database/database.go#L27[`RegisteredDBCreators`]
```

docs/modules/ROOT/pages/integrations/database.adoc
```diff
@@ -0,0 +1,114 @@
...
+====
+
+Specific instructions for the configuration of the PostgreSQL Database are available in
+xref:/getting-started/instllation.adoc#_install_and_configure_kubearchive[Install and Configure KubeArchive].
+
+== Add new database integrations
+
+To add a new database integration follow the instructions:
+
+1. Create a new file under `pkg/database` with the name of the new database type.
```
rh-hemartin, line 45:
```suggestion
1. Create a new file under `pkg/database` named after the new database engine.
```

docs/modules/ROOT/pages/integrations/database.adoc
```diff
@@ -0,0 +1,114 @@
...
+as value.
+
+=== The `newDBCreatorFunc`
+
+This function receives the set of environment variables with the database configuration and returns a
+link:https://github.com/kubearchive/kubearchive/blob/main/pkg/database/facade/creator.go[`DBCreator`]
+implementation.
+
+The recommendation is that the `DBCreator` implementation include the `env` param as parameter
+so the `GetDriverName` and the `GetConnectionString` can access them.
```
rh-hemartin, line 69:
```suggestion
The `DBCreator` implementation must include the `env` param as parameter
so the `GetDriverName` and the `GetConnectionString` can access them.
```

docs/modules/ROOT/pages/integrations/database.adoc
```diff
@@ -0,0 +1,114 @@
...
+
+=== The `DBInterface` implementation returned by the `newDatabaseFunc`
+
+[NOTE]
+====
+This section includes the development guidelines expected by the KubeArchive maintainers.
+They can be skipped, but they will be considered in the review process.
+====
```
rh-hemartin, line 77: If they are optional but they will be considered in the review process they are in fact required.

docs/modules/ROOT/pages/integrations/database.adoc
maruiz93: ops, yes, it should

docs/modules/ROOT/pages/integrations/database.adoc
```diff
@@ -0,0 +1,114 @@
...
+Other interfaces, like `DBFilter`, have a partial implementation.
+
+Those implementations can have the functionality that you need.
+Check them before implementing your own.
+====
+
+If the database interaction logic changes from the implemented in the `Database` struct,
+override the implementation of their functions.
+An example could be with a database that doesn't support upsert queries
+and need to do a transaction with a `SELECT` and an `INSERT`.
```
rh-hemartin, line 111:
```suggestion
If the database interaction logic changes from the implemented in the `Database` struct,
override the implementation of their functions. For example, a database that does not
support upsert queries needs to implement `ResourceInserter` as a series of  `SELECT` and an `INSERT` statements.
```

maruiz93: no, currently neither mariadb or postgresql reimplement those methods, they just use the "default"implementation of the `Database` struct as it is embedded there. 
It's not correct as go doesn't support the inheritance, it uses composition, but the outcome there is like both `PostgreSQLDatabase` and `MariaDBDatabase` "inherit" the methods from `Database` and they don't need to override them.

docs/modules/ROOT/pages/integrations/database.adoc
```diff
@@ -0,0 +1,114 @@
...
+Some interfaces, like `DBDeleter`, have a default implementation.
+
+Other interfaces, like `DBFilter`, have a partial implementation.
+
+Those implementations can have the functionality that you need.
+Check them before implementing your own.
+====
+
+If the database interaction logic changes from the implemented in the `Database` struct,
+override the implementation of their functions.
```
skoved, line 109:
```suggestion
If the database interaction logic changes from the implementation in the `Database` struct,
override the implementation of the affected functions.
```

skoved (commented):
Noticed a few small things

docs/modules/ROOT/pages/integrations/database.adoc
```diff
@@ -0,0 +1,114 @@
...
+====
+The provided schema includes indexes.
+
+KubeArchive users should review and customize the indexes
+if needed based on the expected queries and the amount and
+the distribution of the expected archived data.
+====
+
+Specific instructions for the configuration of the PostgreSQL Database are available in
+xref:/getting-started/instllation.adoc#_install_and_configure_kubearchive[Install and Configure KubeArchive].
```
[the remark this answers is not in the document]
maruiz93: From the user POV, I think it's interesting having also instructions (or a link to them in this case) with the instructions of how "install" the database, as it's not only applying the schema, the user need to know about how the credentials are exposed.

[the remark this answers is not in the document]
rh-hemartin: I agree, maybe we need a section about the configuration in the secret and then link it to the specific example given in the getting started pages

docs/modules/ROOT/pages/integrations/database.adoc
maruiz93: Applied the changes with a little modification:

`Currently, PostgreSQL is the only fully supported database engine.`

WDYT?

docs/modules/ROOT/pages/integrations/database.adoc
maruiz93: applied suggestion

docs/modules/ROOT/pages/integrations/database.adoc
maruiz93: applied suggestion

docs/modules/ROOT/pages/integrations/database.adoc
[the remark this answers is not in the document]
maruiz93: one way or another works for me as long as the information isn't repeated (we should have links, this makes the doc more maintainable) and the information is in both db integration and getting started sections.

If you have a strong preference of having the instructions of configuring the secret in the "database integration" section, I can add the changes.

docs/modules/ROOT/pages/integrations/database.adoc
maruiz93: suggestion applied

docs/modules/ROOT/pages/integrations/database.adoc
maruiz93: fixed

docs/modules/ROOT/pages/integrations/database.adoc
maruiz93: I've replaced it with a `should`.

You can potentially do a working `GetConnectionString` method that doesn't depend on an `env`, but you shouldn't do that because that breaks our "development guidelines".

docs/modules/ROOT/pages/integrations/database.adoc
maruiz93: I removed the second sentence.

The reason I added the note is because these guidelines may not work for every database engine. For example, when adding integration with a non-relational db like mongo.

There might be other examples, but the point is, we will review the PRs considering these guidelines whenever they can be applied.

docs/modules/ROOT/pages/integrations/database.adoc
maruiz93: suggestion applied

docs/modules/ROOT/pages/integrations/database.adoc
```diff
@@ -0,0 +1,114 @@
...
+KubeArchive offers a [filename]`kubearchive.sql` file with the DDL to create
+the schema as part of the installation.
+
+[IMPORTANT]
+====
+The provided schema includes indexes.
+
+KubeArchive users should review and customize the indexes
+if needed based on the expected queries and the amount and
+the distribution of the expected archived data.
```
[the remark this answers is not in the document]
maruiz93: suggestion applied!

docs/modules/ROOT/pages/integrations/database.adoc
maruiz93: suggestion applied

docs/modules/ROOT/pages/integrations/database.adoc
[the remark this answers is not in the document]
rh-hemartin: I don't agree that the information here and in the getting started are the same. This document feels as a reference guide (definition + brief explanation), while the getting started is more of a tutorial (don't understand, just apply and get something working). So here it makes sense that we explain the secret and its keys, while it does not make sense to have a `kubectl patch/apply` example. So explain the secret keys here and link to the getting started as an example.

docs/modules/ROOT/pages/integrations/database.adoc
rh-hemartin: I'm fine with should instead of must. However I think this should not be about what you could do, but about what we want you to do as maintainers of the tool.

docs/modules/ROOT/pages/integrations/database.adoc
```diff
@@ -0,0 +1,114 @@
...
+====
+
+[TIP]
+====
+Some interfaces, like `DBDeleter`, have a default implementation.
+
+Other interfaces, like `DBFilter`, have a partial implementation.
+
+Those implementations can have the functionality that you need.
+Check them before implementing your own.
```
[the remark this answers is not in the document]
maruiz93: suggestion applied

docs/modules/ROOT/pages/integrations/database.adoc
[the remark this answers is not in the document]
maruiz93: I've added some instructions regarding how to apply the schema and the patch to the `kubearchive-database-credentials` Secret.
PTAL

docs/modules/ROOT/pages/integrations/database.adoc
```diff
@@ -0,0 +1,188 @@
...
+----
+<1> The database admin account used to apply the schema. It can be different from the KubeArchive DB user.
+<2> The database URL
+====
+
+=== Database Credentials
+
+The `kubearchive-database-credentials` Secret stores the information to connect KubeArchive with the Database.
+
+The default content of this Secret when KubeArchive is installed has the following data:
```
rh-hemartin, line 67:
```suggestion
The default content of this Secret when KubeArchive is installed is:
```

rh-hemartin (approved):

docs/modules/ROOT/pages/integrations/database.adoc
```diff
@@ -0,0 +1,117 @@
...
+== Overview
+
+KubeArchive requires a database to store resources.
+This document list the databases supported, how to set up the database schema
+and how to add new database integrations.
+
+=== PostgreSQL
+
+Currently, PostgreSQL is the only fully supported database engine.
+The KubeArchive integration test suite runs with PostgreSQL 16.6
```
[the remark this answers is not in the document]
maruiz93: > Can we just point to the kubearchive.sql file to tell the user which version is supported? Then we don't have to change it in multiple places.

It's ok for me, I'll do the change

> Also, why are we talking about the integration test suite? Shouldn't it just say KubeArchive needs PostgreSQL version of 16.6 or later?

Hmm, I don't think so, I'm pretty sure that KubeArchive works with lower versions of PostgreSQL, it's just that we test the product with a specific one, so if the user wants to try a different one it's more "risky" but it's probably going to be ok unless they use a really old version. Maybe a couple of errors on the import of the pg_dump, but as far as I'm concerned, related to the queries we don't use any really recent functionality.

docs/modules/ROOT/pages/integrations/database.adoc
```diff
@@ -0,0 +1,117 @@
...
+=== MariaDB
+
+This integration is still in progress.
+
+The MariaDB implementation is available
+link:https://github.com/kubearchive/kubearchive/blob/main/pkg/database/mariadb.go[here].
+
+== Configuration and Customization
+
+KubeArchive offers a [filename]`kubearchive.sql` file with the DDL to create
```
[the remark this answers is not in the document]
maruiz93: In this sentence I use both `DDL` and `schema` because they are not exactly the same. DDL are the instructions used to defined a schema. Anyway, I get the point that there is no need to add acronyms and people that aren't used to database stuff may not be familiar with them, so I'll rephrase the sentence.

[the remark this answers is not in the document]
maruiz93: Changed!

docs/modules/ROOT/pages/integrations/database.adoc
```diff
@@ -0,0 +1,188 @@
...
+  DATABASE_PORT: "5432"
+  DATABASE_URL: "kubearchive-rw.postgresql.svc.cluster.local"
+  DATABASE_DB: "kubearchive"
+  DATABASE_USER: "kubearchive"
+  DATABASE_PASSWORD: "Databas3Passw0rd"  # notsecret
+----
+
+The KubeArchive user should update the secret with the specific values of their database and
+restart the pods accordingly to pick the new values.
+The command for changing the most common values, URL and password, is:
```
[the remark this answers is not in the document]
maruiz93: suggestion applied

docs/modules/ROOT/pages/integrations/database.adoc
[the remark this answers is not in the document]
ggallen: > > Also, why are we talking about the integration test suite? Shouldn't it just say KubeArchive needs PostgreSQL version of 16.6 or later?
> 
> Hmm, I don't think so

Typically database version are not backward compatible, but are forward compatible. In other words, if we say we require 16.6, I should not expect < 16.6 to work but would expect >= 16.6 to work. We should say "at least 16.6" or "16.6 or later".

And it's still not clear why we are talking about the integration test suite. KubeArchive needs a particular PostgreSQL version, not the test suite.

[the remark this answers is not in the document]
maruiz93: updated

skoved (approved):

PR #811, opened 2025-02-07
sink sends proper http response based on processing of a cloud event

### LINKED ISSUE #779: Sink should report Cloud Event delivery failure if an error occurs during event processing
When the sink receives an event, if the following happens:

- failed database connection when attempting to write
- failed to delete the resource from the cluster (if deletion is required)

the sink should send an http error result back to the sender. This lets the sender know that something went wrong and the sink did not process the event. This lets the sender mark the message as undelivered and it can act accordingly. In our use case, the Knative Trigger that sends Cloud Events to the sink will first retry sending the event and then if the retries max out, the message will be sent to the dead letter sink.

This can be done by adding `cloudevents.Result` as the return type of `routers.Controller.receiveCloudEvent`. The result can be created with [`cloudevents.NewHTTPResult`](https://pkg.go.dev/github.com/cloudevents/sdk-go/v2#NewHTTPResult)

skoved:

Resolves #779, related to #814

## Release Note

```release-note
Now KubeArchive retries to process resource updates if a failure happens.
```

## Notes for Reviewers

if the sink fails to write the resource or log urls from a cloud event to the database or if the sink fails to delete
the resource from the cluster, the sink will respond with an http 500 response. This will allow senders of Cloud
Events to respond to cloud event delivery failure. The exception to this is when the payload of the Cloud Event
is not a kubernetes resource. In this case a 422 response is sent.

### DIFF
--- a/cmd/sink/routers/result.go
+++ b/cmd/sink/routers/result.go
@@ -1,0 +1,19 @@
+// Copyright KubeArchive Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package routers
+
+import (
+	cloudevents "github.com/cloudevents/sdk-go/v2"
+	"github.com/cloudevents/sdk-go/v2/protocol"
+)
+
+func NewCEResult(statusCode int, err error) protocol.Result {
+	if statusCode >= 200 && statusCode < 400 {
+		// Cloud Event Processed successfully
+		return cloudevents.NewHTTPResult(statusCode, "cloud event processed successfully")
+	} else if err == nil {
+		return cloudevents.NewHTTPResult(statusCode, "an error occurred while processing the cloud event")
+	}
+	return cloudevents.NewHTTPResult(statusCode, "an error occurred while processing the cloud event", err.Error())
+}
--- a/cmd/sink/routers/routers.go
+++ b/cmd/sink/routers/routers.go
@@ -15,6 +15,7 @@ import (
 	ceOtelObs "github.com/cloudevents/sdk-go/observability/opentelemetry/v2/http"
 	cloudevents "github.com/cloudevents/sdk-go/v2"
 	ceClient "github.com/cloudevents/sdk-go/v2/client"
+	"github.com/cloudevents/sdk-go/v2/protocol"
 	"github.com/gin-gonic/gin"
 	"github.com/kubearchive/kubearchive/cmd/api/abort"
 	"github.com/kubearchive/kubearchive/cmd/sink/filters"
@@ -60,7 +61,7 @@ func NewController(
 	return controller, nil
 }
 
-func (c *Controller) writeLogs(ctx context.Context, obj *unstructured.Unstructured) {
+func (c *Controller) writeLogs(ctx context.Context, obj *unstructured.Unstructured) error {
 	logCtx, cancel := context.WithTimeout(ctx, time.Second*5)
 	defer cancel()
 	urls, err := c.LogUrlBuilder.Urls(logCtx, obj)
@@ -73,7 +74,7 @@ func (c *Controller) writeLogs(ctx context.Context, obj *unstructured.Unstructur
 			"name", obj.GetName(),
 			"err", err,
 		)
-		return
+		return err
 	}
 	if len(urls) == 0 {
 		slog.Info(
@@ -83,7 +84,7 @@ func (c *Controller) writeLogs(ctx context.Context, obj *unstructured.Unstructur
 			"namespace", obj.GetNamespace(),
 			"name", obj.GetName(),
 		)
-		return
+		return nil
 	}
 	err = c.Db.WriteUrls(logCtx, obj, c.LogUrlBuilder.GetJsonPath(), urls...)
 	if err != nil {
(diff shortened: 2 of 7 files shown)


rh-hemartin (changes requested):
Change the release note to something like:

```
Now KubeArchive retries to process resource updates if a failure happens.
```

skoved:
@rh-hemartin updated the release note

rh-hemartin (approved):

mafh314 (approved):

maruiz93 (approved):

cmd/sink/routers/result.go
```diff
@@ -0,0 +1,19 @@
...
+
+package routers
+
+import (
+	cloudevents "github.com/cloudevents/sdk-go/v2"
+	"github.com/cloudevents/sdk-go/v2/protocol"
+)
+
+func NewCEResult(statusCode int, err error) protocol.Result {
+	// Cloud Event Processed successfully
```
ggallen, line 12: I think this comment belongs inside the `true` clause of the `if` statement.

cmd/sink/routers/result.go
```diff
@@ -0,0 +1,19 @@
...
+	if statusCode >= 200 && statusCode < 400 {
+		return cloudevents.NewHTTPResult(statusCode, "cloud event processed successfully")
+	} else if err == nil {
+		return cloudevents.NewHTTPResult(statusCode, "an error occurred while processing the cloud event")
+	}
```
ggallen, line 17: Is this condition ever possible?  I don't think it is, and if so, the `else` clause should be removed.

cmd/sink/routers/routers.go
```diff
@@ -90,25 +93,31 @@ func (c *Controller) writeResource(ctx context.Context, obj *unstructured.Unstru
...
+	)
 	// only write logs for k8s resources likes pods that have them and UrlBuilder is configured
 	if c.LogUrlBuilder != nil && strings.ToLower(obj.GetKind()) == "pod" {
-		c.writeLogs(ctx, obj)
+		err = c.writeLogs(ctx, obj)
 	}
+	if err != nil {
+		return err
+	}
+	return nil
```
ggallen, line 110: Why not just:
```
return err
```

pkg/database/fake/database_generated.go
```diff
@@ -152,11 +161,18 @@ func (f *Database) filterResourceByKindApiVersionNamespaceAndName(kind, apiVersi
...
+		return f.err
+	}
 	f.resources = append(f.resources, k8sObj)
 	return nil
 }
 
 func (f *Database) WriteUrls(_ context.Context, k8sObj *unstructured.Unstructured, jsonPath string, logs ...models.LogTuple) error {
+	if f.urlErr != nil {
+		return f.urlErr
+	}
```
ggallen, line 174: Why even make this call if there is a known error with the database?

skoved:
rebased and resolved the merge conflict

pkg/database/fake/database_generated.go
skoved: This condition is used in the unit tests for the sink to simulate `WriteUrls` returning an error so that we can test that the sink responds with the correct http status codes

cmd/sink/routers/routers.go
skoved: not sure why I did that. I removed the if at the end of `writeResource` and replaced it with `return err`

cmd/sink/routers/result.go
skoved: you're right. Moved it inside `if`

cmd/sink/routers/result.go
ggallen: The I would split the method into two:

```
func NewCESuccessResult() protocol.Result {

func NewCEErrorResult(statusCode int, err error) protocol.Result {
```

skoved: `NewCEErrorResult` would still have the same issue, which is that error can be nil, which is an issue if it needs to call `err.Error()`

pkg/database/fake/database_generated.go
skoved: yeah, `pkg/database/fake` is a database mock for testing

cmd/sink/routers/result.go
ggallen: Yes. I am assuming that someone who calls an error routine that takes an `error` parameter will in fact pass an error in and not `nil` (which is actually to golang indication for *not* an error). That just seems like common sense. But it could be done.

Note however that you also make an assumption. A user could pass a success HTTP code *and* an error. In that case you assume the success is correct. But do you really know? 

I could actually argue that the panic (in my case) which would occur when passing in a `nil` is better than the scenario of a success HTTP code and an error. In the former it will be clear there is a problem, in the latter it will be hidden.

In fact, I think these routines should be split as I suggested.  And in the error method you can check that the error code is not success (and throw a new error if it is) and also check for `nil` for the err.

skoved: > Yes. I am assuming that someone who calls an error routine that takes an error parameter will in fact pass an error in and not nil (which is actually to golang indication for not an error). That just seems like common sense. But it could be done.

I agree with you that a developer using `NewCEErrorResult` should not pass a nil error when providing http response code < 200 or >= 400

> Note however that you also make an assumption. A user could pass a success HTTP code and an error. In that case you assume the success is correct. But do you really know?

I think this a different argument than the one that I'm making. I am **NOT** trying to prevent a developer using this function from calling this function with values that don't make sense, I'm trying to make function not crash. The constraints that I'm operating on are:

- provide the error message in the response (makes it easier of users and developers of kubearchive to debug issues)
- `nil` is a valid error value
- always return a valid `protocol.Result` (this is need for setting the http response code)

Since `nil` is valid error value and we have to call `err.Error()` to get the error message, we need to provide a `nil` check in `NewCEErrorResult`

> I could actually argue that the panic (in my case) which would occur when passing in a nil is better than the scenario of a success HTTP code and an error. In the former it will be clear there is a problem, in the latter it will be hidden.

I think to the extent that we can, we should try to avoid scenarios where we could cause a panic by de referencing nil. We should not want kubearchive to crash

ggallen: Fair points. I'm not advocating for a panic either, just pointing out that from a debugging standpoint it might be preferable.

skoved:
rebased from main and added an missing error

skoved:
rebasing to get the PR to merge

maruiz93:
> I updated the PR so that if `event.Data` cannot be unmarshalled into an `unstructed.Unstructured`, the sink will send an HTTP 422 response

Can you update the Note for Reviewers accordingly?

skoved:
> > I updated the PR so that if `event.Data` cannot be unmarshalled into an `unstructed.Unstructured`, the sink will send an HTTP 422 response
> 
> Can you update the Note for Reviewers accordingly?

done!

PR #825, opened 2025-02-12
add design doc for the cloud event delivery pipeline

skoved:

Resolves #

## Release Note

```release-note
The documentation now contains a page that details how KubeArchive handles Cloud Events
```

## Notes for Reviewers

This document is explaining how KubeArchive is already configured, but I think it is important to have it all written down somewhere. The only part that is missing is how to get cloud events from the dead letter sink to the KubeArchive sink, which is still being investigated.


docs/modules/ROOT/images/ce-pipeline.png
maruiz93, line 1: It's not very important, but I think it would look better to have a plain white background.

If you did this using miro, you can do it by using the frame feature.

docs/modules/ROOT/pages/design/cloudevents.adoc
```diff
@@ -0,0 +1,125 @@
...
+If the sink becomes unreachable or unable to process cloud events, the dead letter sink will start to receive cloud
+events from the primary broker.
+
+The sink needs to be able to receive cloud events from the dead letter sink after recovering from a error state, but the
+process for doing this is still TBD.
+
+=== Triggers
+
+KubeArchive creates a trigger that subscribes the KubeArchive sink to all cloud events stored by the primary broker. It
+is configured as follows:
```
maruiz93, line 84: If I understand correctly, based on the attached diagram, KubeArchive will also create a second trigger that subscribes the KubeArchive sink to all cloud events stored by the dead letter sink, right?

Shouldn't we add some paragraph explaining that?

maruiz93 (commented):
A little cosmetic suggestion and a question about the subscription to the dead letter sink, but looks good :)

docs/modules/ROOT/pages/design/cloudevents.adoc
```diff
@@ -0,0 +1,125 @@
...
+KubeArchive uses Cloud Events that are generated by a Knative Eventing ApiServerSource to discover when resources on
+the kubernetes cluster have changed. These Cloud Events contain the json representation of the Kubernetes resource that
+has changed. The KubeArchive Sink (sink) receives these messages and determines, using CEL expression from the
+KubeArchiveConfig, if the data contained in the Cloud Event should be written to the database and potentially deleted
+from the cluster. While the sink is processing a Cloud Event, it is possible at many points for an error to occur,
+especially when the sink has to make a request of the network, particularly, to the Kubernetes API or the database. This
+page proposes how KubeArchive set up its Cloud Event delivery pipeline and how it should consume Cloud Events to achieve
+the following goals:
+
+* KubeArchive receives all Cloud Events from the Knative ApiServerSource
```
rh-hemartin, line 13: I don't think this should be a goal, as its stated that KubeArchive already receives CloudEvents from the Knative Eventing ApiServerSource.

docs/modules/ROOT/pages/design/cloudevents.adoc
```diff
@@ -0,0 +1,125 @@
...
+* Cloud Events sent to KubeArchive can be replayed
+* Cloud Events received by KubeArchive, that cannot be processed successfully, are reported back to the sender as
+  delivery failure
+* Cloud Events sent to the sink will attempt to be redelivered if the initial delivery fails
+* Cloud Events that could not be delivered are replayed to KubeArchive automatically when KubeArchive becomes available
+  again
```
rh-hemartin, line 19: I think we can generalize and compact these, what about:

```suggestion
* Resource updates can be rejected by KubeArchive if a problem happens during processing
* Resource updates are sent again to KubeArchive if it does not provide a response back
```

docs/modules/ROOT/pages/design/cloudevents.adoc
```diff
@@ -0,0 +1,125 @@
...
+
+== Design
+
+KubeArchive needs to use the functionality provided by Knative Eventing, Brokers and Triggers as well as features
+provided by the Cloud Events Specification to achieve an event driven architecture, where events are not lost and events
+can be replayed safely without worries about side effects.
```
rh-hemartin, line 25:
```suggestion
We decided to use an Event Driven architecture for KubeArchive. Reusing Knative Eventing's components the system can be fault tolerant and scale well with multiple event consumers.
```

docs/modules/ROOT/pages/design/cloudevents.adoc
```diff
@@ -0,0 +1,125 @@
...
+
+=== Knative Eventing Brokers
+
+Instead of the ApiServerSource sending Cloud Events directly to the sink, the Cloud Events should instead be sent to a
+Knative Eventing Broker. Brokers can receive, persist, and send messages. Using the broker allows messages to be
+sent somewhere that they can be stored long term and is separate from the KubeArchive sink and Knative Eventing
+ApiServerSource, so if either are restarted, the Cloud Events won't be lost. KubeArchive will use two brokers. One
+primary broker (called broker) and secondary broker (called dead letter sink). The default configuration of
+Knative Eventing will have these brokers be a created as In Memory Brokers, but changing the Knative Eventing
+configuration can allow this broker to be created, for example, as a Kafka Broker instead
```
rh-hemartin, line 40:
```suggestion
Instead of sending directly events directly to KubeArchive's sink, we decided to send the events to a Knative Eventing Broker. These can persist events in case of consumer or producer failure. By default KubeArchive uses two in-memory brokers: primary and dead letter sink. The primary broker receives the events from the ApiServerSource and sends them to the dead letter sink broker if its unable to deliver them to KubeArchive.
```

docs/modules/ROOT/pages/design/cloudevents.adoc
```diff
@@ -0,0 +1,125 @@
...
+    deadLetterSink:
+      ref:
+        apiVersion: eventing.knative.dev/v1
+        kind: Broker
+        name: kubearchive-dls
+----
+
+If the initial delivery fails, the broker will attempt to deliver the cloud event 4 additional times. When attempting to
+retry delivering the event, the broker will use an exponential backoff algorithm to calculate the amount of time to wait
+between retry attempts, and if all retry attempts fail, the broker will send the cloud event to the dead letter sink.
```
rh-hemartin, line 69:
```suggestion
If the initial delivery fails, the broker attempts to deliver the cloud event 4 additional times using an exponential backoff algorithm. If all retry attempts fail, the broker sends the Cloud Event to the dead letter sink broker.
```

docs/modules/ROOT/pages/design/cloudevents.adoc
```diff
@@ -0,0 +1,125 @@
...
+
+==== Dead Letter Sink
+
+The dead letter sink is a second broker that stores cloud events that could not be delivered to the KubeArchive sink. If
+the sink has received every message from the broker, then there will be no cloud events stored in the dead letter sink.
+If the sink becomes unreachable or unable to process cloud events, the dead letter sink will start to receive cloud
+events from the primary broker.
```
rh-hemartin, line 76:
```suggestion
The dead letter sink stores the Cloud Events that could not be delivered to the KubeArchive sink. This broker does not contain any message if the connectivity between the Primary Broker and the KubeArchive sink is correct.
```

docs/modules/ROOT/pages/design/cloudevents.adoc
rh-hemartin, line 84:
```suggestion
KubeArchive creates a couple of Knative Eventing Triggers to receive events from the primary and dead letter sink brokers. It is configured as follows:
```

docs/modules/ROOT/pages/design/cloudevents.adoc
```diff
@@ -0,0 +1,125 @@
...
+      apiVersion: v1
+      kind: Service
+      name: kubearchive-sink
+----
+
+This configuration tells the primary broker to send cloud events to the url for the `kubearchive-sink` service on the
+`/` route using the `HTTP POST` method.
+
+KubeArchive will need to create a trigger so the sink can subscribe to cloud events from the dead letter sink, but how
+this will be configured is TBD.
```
rh-hemartin, line 101:
```suggestion
This configuration tells the primary broker to target the `kubearchive-sink` Service. By default events are sent to the root endpoint (`/`) using the `POST` method.
```

docs/modules/ROOT/pages/design/cloudevents.adoc
```diff
@@ -0,0 +1,125 @@
...
+
+=== HTTP Status Codes
+
+The sink receives cloud events from Knative Eventing over HTTP on the `/` route using the `HTTP POST` method. After
+receiving the cloud event, the sink has to process the cloud event to determine if the resource contained in the cloud
+event needs to be written to the database and if the resource needs to be deleted from the cluster. All cloud events
+must be successfully processed for KubeArchive to function as intended. To make sure that this happens, KubeArchive will
+respond to cloud events with the following HTTP response codes:
```
rh-hemartin, line 109:
```suggestion
The sink receives Cloud Events from the ApiServerSource over HTTP on the `/` route using a `POST` method. After
receiving the event, the sink processes it to determine if the resource change needs to be written to the database and if the resource needs to be deleted from the cluster. The HTTP response code KubeArchive sends to the `POST` request have different meanings:
```

docs/modules/ROOT/pages/design/cloudevents.adoc
ggallen: Why not just say?
```
If the initial delivery fails, the broker will try to resend the Cloud Event
using the the configured `retry`and `backoffPolicy` parameters.
These parameters may be changed by the user, see the Knative
documentation for details. If all retry attempts fail, the broker sends
the Cloud Event to the dead letter sink broker.
```
I don't see any good reason to include the actual default values in the text. What if they change?

docs/modules/ROOT/pages/design/cloudevents.adoc
ggallen: I think this is enough:
```
The dead letter sink stores the Cloud Events that could not
be delivered to the KubeArchive sink.
```

docs/modules/ROOT/pages/design/cloudevents.adoc
```diff
@@ -0,0 +1,125 @@
...
+|successfully received and processed the cloud event
+
+|422
+|successfully received a valid cloud event, but the `data` field could not be unmarshalled into a Kubernetes object. Do
+not send the cloud event again without modifying the `data` field first.
+
+|500
+|successfully received a valid cloud event with a kubernetes resource in it, but kubearchive encoutered an error while
+trying to process the resource. This can happen if there was issue writing to the database or requesting the Kubernetes
+API to delete an object. The cloud event can be sent again without modifying.
```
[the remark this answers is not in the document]
ggallen: Why do we even need the important section?  First, the user does not have to "react to each of the response codes given by KubeArchive".  Most will be 200, which don't require anything to be done.

Users can respond or not to other response codes.  I see no point to mark it as important.

ggallen:
Has this design been reviewed by the Knative team?

skoved:
> Has this design been reviewed by the Knative team?

The setup with the 2 brokers and a trigger to send the messages to the kubearchive sink came from suggestions from the knative team. They have not looked over our whole setup though.

I'd also like to mention that this was meant to capture the current state of cloud events in kubearchive and how things work. I still need to work with knative team to figure out how getting cloud events from the dead letter sink to the kubearchive sink should be setup

docs/modules/ROOT/pages/design/cloudevents.adoc
skoved: There's a section that explains the different HTTP error codes that the sink can respond with and how they should be treated. I don't really view the HTTP error codes as part of the high level goals, but more as required implementation details based on how Knative Eventing handles sending events over HTTP

docs/modules/ROOT/images/ce-pipeline.png
skoved: I didn't know you could do that! I've updated the diagram so it has a white background

docs/modules/ROOT/pages/design/cloudevents.adoc
skoved: I have removed the trigger for the dead letter sink from the diagram because the current configuration of KubeArchive does not create it. For I think it makes sense to talk about the one trigger KubeArchive deploys with by default and mention that there will be a second one eventually but the details have not been determined yet. When we finish the the work for integrating the dead letter sink, we can update this section accordingly

docs/modules/ROOT/pages/design/cloudevents.adoc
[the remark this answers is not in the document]
skoved: I agree with @ggallen. This is here so that we have an official list of the http response codes that the sink can generate and what they mean. This should not really matter to users since the handling of the response codes are done entirely by knative

docs/modules/ROOT/pages/design/cloudevents.adoc
skoved: done with a slight tweak

docs/modules/ROOT/pages/design/cloudevents.adoc
skoved: I changed the wording in this section according to the suggestions here and added a sentence to mention that KubeArchive can reject a cloud event that it receives and added a link to the section on http error codes

docs/modules/ROOT/pages/design/cloudevents.adoc
skoved: I made the change here and took the suggestion to not mention specific values but instead list the params that will affect if/how retries will be done

docs/modules/ROOT/pages/design/cloudevents.adoc
```diff
@@ -0,0 +1,125 @@
...
+Knative Eventing Broker. Brokers can receive, persist, and send messages. Using the broker allows messages to be
+sent somewhere that they can be stored long term and is separate from the KubeArchive sink and Knative Eventing
+ApiServerSource, so if either are restarted, the Cloud Events won't be lost. KubeArchive will use two brokers. One
+primary broker (called broker) and secondary broker (called dead letter sink). The default configuration of
+Knative Eventing will have these brokers be a created as In Memory Brokers, but changing the Knative Eventing
+configuration can allow this broker to be created, for example, as a Kafka Broker instead
+
+[IMPORTANT]
+====
+For Production workloads, Knative Eventing strongly recommends using Brokers types other than the In Memory Broker.
```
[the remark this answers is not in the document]
skoved: done

docs/modules/ROOT/pages/design/cloudevents.adoc
skoved: Made the change. I tweaked your suggestion slightly

docs/modules/ROOT/pages/design/cloudevents.adoc
skoved: I removed it from the list of goals

docs/modules/ROOT/pages/design/cloudevents.adoc
skoved: That's my bad. I think the solution should be to remove the second trigger from the diagram for now. 

KubeArchive does not currently include a second trigger for the dead letter sink and the configuration around the dead letter sink still needs to be figured out

maruiz93 (approved):

rh-hemartin:
@skoved at some point you mentioned you would change this, can you ping us when we need to review this again?

PR #844, opened 2025-02-20
Support empty KubeArchiveConfig.

### LINKED ISSUE #780: Support empty KubeArchiveConfig
One use case is to control KubeArchive for the whole cluster using a global KubeArchiveConfig (the one that is placed on the installation namespace is the global one) and having empty KubeArchiveConfigs where resources need to be archived. I tested it and the operator raises an error on its logs:

```
2025/01/24 09:21:06 ERROR Failed to create ApiServerSource controller=kubearchiveconfig controllerGroup=kubearchive.kubearchive.org controllerKind=KubeArchiveConfig KubeArchiveConfig.name=kubearchive KubeArchiveConfig.namespace=testing-kac namespace=testing-kac name=kubearchive reconcileID=163c34d1-8649-4dba-8754-80dbd5e718a9 err="ApiServerSource.sources.knative.dev \"kubearchive-a13e\" is invalid: spec.resources: Required value"
2025/01/24 09:21:06 ERROR Reconciler error controller=kubearchiveconfig controllerGroup=kubearchive.kubearchive.org controllerKind=KubeArchiveConfig KubeArchiveConfig.name=kubearchive KubeArchiveConfig.namespace=testing-kac namespace=testing-kac name=kubearchive reconcileID=163c34d1-8649-4dba-8754-80dbd5e718a9 err="ApiServerSource.sources.knative.dev \"kubearchive-a13e\" is invalid: spec.resources: Required value"
```

So without any filters, KubeArchive is not capable of creating an ApiServerSource. When a global KubeArchiveConfig is applied the ApiServerSource is created.

**Acceptance Criteria**

* If there is no ApiServerSource and the KubeArchiveConfig being reconciled is empty, don't try to create the ApiServerSource.
* OR don't create an ApiServerSource if its `resources` is empty. (I'm not sure about the implementation)

ggallen:

Resolves #780

## Release Note

```release-note
NONE
```

## Notes for Reviewers

ApiServerSource requires resources to run successfully.  These changes take that into account. If at any point reconcile finds there are no resources to give to ApiServerSource, then the existing ApiServerSource is deleted.  It will be recreated the next time reconcile runs and has resources.

### DIFF
--- a/test/integration/integration_test.go
+++ b/test/integration/integration_test.go
@@ -11,14 +11,15 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
-
 	"os/exec"
 	"testing"
+	"time"
 
 	"github.com/avast/retry-go/v4"
 	"github.com/kubearchive/kubearchive/test"
 	authenticationv1 "k8s.io/api/authentication/v1"
 	corev1 "k8s.io/api/core/v1"
+	errs "k8s.io/apimachinery/pkg/api/errors"
 	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
@@ -107,6 +108,18 @@ func TestNormalOperation(t *testing.T) {
 		errNamespace = clientset.CoreV1().Namespaces().Delete(context.Background(), namespaceName, metav1.DeleteOptions{})
 		if errNamespace != nil {
 			t.Fatal(errNamespace)
+		}
+
+		retryErr := retry.Do(func() error {
+			_, getErr := clientset.CoreV1().Namespaces().Get(context.Background(), namespaceName, metav1.GetOptions{})
+			if !errs.IsNotFound(getErr) {
+				return errors.New("Waiting for namespace "+namespaceName+" to be deleted")
+			}
+			return nil
+		}, retry.Attempts(10), retry.MaxDelay(3*time.Second))
+
+		if retryErr != nil {
+			t.Log(retryErr)
 		}
 	})
 
--- a/test/integration/operator_test.go
+++ b/test/integration/operator_test.go
@@ -1,0 +1,227 @@
+// Copyright KubeArchive Authors
+// SPDX-License-Identifier: Apache-2.0
+//go:build integration
+
+package main
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/avast/retry-go/v4"
+	kubearchivev1alpha1 "github.com/kubearchive/kubearchive/cmd/operator/api/v1alpha1"
+	"github.com/kubearchive/kubearchive/test"
+	corev1 "k8s.io/api/core/v1"
+	errs "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+type Test int
+
+const (
+	emptyKAC Test = iota
+	nonEmptyKAC
+)
+
+const (
+	kacName     = "kubearchive"
+	a13eName    = kacName + "-a13e"
+	sinkName    = kacName + "-sink"
+	filtersName = "sink-filters"
+)
+
+func TestEmptyKAC(t *testing.T) {
+	runTest(t, emptyKAC, map[string]any{})
+}
+
+func TestNonEmptyKAC(t *testing.T) {
+	resources := map[string]any{
+		"resources": []map[string]any{
+			{
+				"selector": map[string]string{
+					"apiVersion": "v1",
+					"kind":       "Pod",
+				},
+				"archiveWhen": "true",
+				"deleteWhen":  "status.phase == 'Succeeded'",
+			},
+		},
+	}
+	runTest(t, nonEmptyKAC, resources)
+}
+
+func runTest(t testing.TB, testName Test, resources map[string]any) {
+	t.Helper()
+
+	clientset, dynaclient, errClient := test.GetKubernetesClient()
+	if errClient != nil {
+		t.Fatal(errClient)
+	}
+
+	// Create test namespace
+	namespaceName := fmt.Sprintf("test-%s", test.RandomString())
+	t.Log("Running test in namespace "+namespaceName)
+	_, errNamespace := clientset.CoreV1().Namespaces().Create(context.Background(), &corev1.Namespace{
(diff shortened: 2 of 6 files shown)


rh-hemartin (approved):

ggallen:
Yeah, I'll add some tests.  I'm reworking this change now.

test/integration/operator_test.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+
+	err = checkResourcesAfterDelete(namespaceName, testName)
+	if err != nil {
+		t.Fatal(err)
+	}
+}
+
+func checkResourcesAfterApply(namespace string, testName Test) error {
+	clientset, dynaclient, _ := test.GetKubernetesClient()
```
skoved, line 120: Since `checkResourcesAfterApply` is essentially a helper function for `runTest`, I think it could be made simpler by doing the following:

```suggestion
func checkResourcesAfterApply(t testing.TB, namespace string, testName Test) {
	t.Helper()
	
	clientset, dynaclient, _ := test.GetKubernetesClient()
```

because `t.Helper()` is called, instead of having `checkResourcesAfterApply` return errors it encounters, it can call `t.Fatal` or `assert.<whatever>` itself instead of the `runTests` having to check a return value

test/integration/operator_test.go
```diff
@@ -0,0 +1,231 @@
...
+				"archiveWhen": "true",
+				"deleteWhen":  "status.phase == 'Succeeded'",
+			},
+		},
+	}
+	runTest(t, nonEmptyKAC, resources)
+}
+
+func runTest(t *testing.T, testName Test, resources map[string]any) {
+	clientset, dynaclient, errClient := test.GetKubernetesClient()
```
skoved, line 59: since runTest is a helper method I think the following change should be made

```suggestion
func runTest(t testing.TB, testName Test, resources map[string]any) {
	t.Helper()
	
	clientset, dynaclient, errClient := test.GetKubernetesClient()
```

`testing.TB` is an interface that is implemented by `*testing.T` and `*testing.B`. Doing this will make code reuse easier if we add benchmark tests in the future. Calling `t.Helper()` will make debugging unit tests easier, since the the test failure will be reported as the function that called `runTest`. This prevents potential confusion because both `TestEmptyKAC` and `TestNonEmptyKAC` call `runTest` so it may not be clear which one failed

test/integration/operator_test.go
```diff
@@ -0,0 +1,231 @@
...
+
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func checkResourcesAfterDelete(namespace string, testName Test) error {
+	clientset, dynaclient, _ := test.GetKubernetesClient()
```
skoved, line 169:
```suggestion
func checkResourcesAfterDelete(t testing.TB, namespace string, testName Test) error {
	t.Helper()
	
	clientset, dynaclient, _ := test.GetKubernetesClient()
```

for the same reason as `checkResourcesAfterApply` because you can have `checkResourcesAfterDelete` call `t.Fatal` or `assert.<whatever> directly

test/integration/integration_test.go
```diff
@@ -108,6 +109,18 @@ func TestNormalOperation(t *testing.T) {
...
+			_, getErr := clientset.CoreV1().Namespaces().Get(context.Background(), namespaceName, metav1.GetOptions{})
+			if !errs.IsNotFound(getErr) {
+				return errors.New("Waiting for namespace "+namespaceName+" to be deleted")
+			}
+			return nil
+		}, retry.Attempts(10), retry.MaxDelay(3*time.Second))
+
+		if retryErr != nil {
+			t.Log(retryErr)
+		}
```
maruiz93, line 123: If this `cleanup` function is going to be used in most of the tests, I believe we should move it to the `main.go` file of the `integration` directory.

test/integration/operator_test.go
```diff
@@ -0,0 +1,227 @@
...
+		}
+		return nil
+	}, retry.Attempts(10), retry.MaxDelay(3*time.Second))
+
+	if err != nil {
+		t.Fatal(err)
+	}
+}
+
+func createKAC(namespace string, resources map[string]any) *unstructured.Unstructured {
```
maruiz93, line 213: this name is a little missleading, at first I thought that the `createKAC` function was taking care of creating the resource in the cluster, but its only instantiating it. How about `newKAC`? 

```suggestion
func newKAC(namespace string, resources map[string]any) *unstructured.Unstructured {
```

test/integration/operator_test.go
```diff
@@ -0,0 +1,227 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+type Test int
+
+const (
+	emptyKAC Test = iota
+	nonEmptyKAC
```
maruiz93, line 28: These are good tests, but they test only the "local" creation of kac resources, without any global kac in place.

IIUC, when a global KAC is in place, the "empty" local KAC is allowed, and I think that would be great to test.

WDYT?

test/integration/integration_test.go
ggallen: I  will handle this as part of https://github.com/kubearchive/kubearchive/issues/863, which I just opened. All the tests can then use this cleanup method.

test/integration/operator_test.go
ggallen: Fixed.

test/integration/operator_test.go
ggallen: I will handle this as part of https://github.com/kubearchive/kubearchive/issues/863, which I just opened. I have a number of PRs currently working in this same area and want to get some resolved.

> IIUC, when a global KAC is in place, the "empty" local KAC is allowed, and I think that would be great to test.

Just to clarify, an empty `KubeArchiveConfig` is always possible.  A global `KubeArchiveConfig` is not required.

maruiz93: yes my bad, I'll rephrase it: an empty `KubeArchiveConfig` is it's not created with a global one, it won't trigger the creation of an `APIServerSource` but with a global `KubeArchiveConfig` that it's not empty, the `APIServerSource` will be created, right?

maruiz93 (approved):

test/integration/operator_test.go
ggallen: It's a little more general than that.  An `ApiServerSource` is only created when there are resources for it.  So the first `KubeArchiveConfig` that has resources will result in an `ApiServerSource` being created.  Conversely, when there are no `KubeArchiveConfigs` that have resources, the `ApiServerSource` will be removed.

Issue #870, opened 2025-02-28
Improve performance on the pagination integration test.


ggallen:
The pagination integration test takes a long time to run in Github. It's usually fast on my laptop, but now I see it slowing down there as well.  This was uncovered while looking at https://github.com/kubearchive/kubearchive/issues/863.
```
[gallen@gallen-thinkpadp1gen5 kubearchive] (issue-863)$ go test -count=1 -v ./test/integration/pagination_test.go -tags=integration -timeout 60m 
=== RUN   TestPagination
    pagination_test.go:33: Creating test namespace 'test-ztsjqlmx'
Forwarding from 127.0.0.1:8081 -> 8081
Forwarding from [::1]:8081 -> 8081

The HTTP status returned is OK, returned 1 items
The HTTP status returned is OK, returned 1 items
The HTTP status returned is OK, returned 1 items
The HTTP status returned is OK, returned 1 items
The HTTP status returned is OK, returned 1 items
The HTTP status returned is OK, returned 1 items
The HTTP status returned is OK, returned 1 items
The HTTP status returned is OK, returned 1 items
The HTTP status returned is OK, returned 1 items
The HTTP status returned is OK, returned 2 items
The HTTP status returned is OK, returned 2 items
The HTTP status returned is OK, returned 2 items
The HTTP status returned is OK, returned 2 items
The HTTP status returned is OK, returned 2 items
The HTTP status returned is OK, returned 3 items
The HTTP status returned is OK, returned 3 items
The HTTP status returned is OK, returned 3 items
The HTTP status returned is OK, returned 3 items
The HTTP status returned is OK, returned 3 items
The HTTP status returned is OK, returned 3 items
The HTTP status returned is OK, returned 4 items
The HTTP status returned is OK, returned 4 items
The HTTP status returned is OK, returned 4 items
The HTTP status returned is OK, returned 4 items
The HTTP status returned is OK, returned 4 items
The HTTP status returned is OK, returned 4 items
The HTTP status returned is OK, returned 5 items
The HTTP status returned is OK, returned 5 items
The HTTP status returned is OK, returned 5 items
The HTTP status returned is OK, returned 5 items
The HTTP status returned is OK, returned 5 items
The HTTP status returned is OK, returned 6 items
The HTTP status returned is OK, returned 6 items
The HTTP status returned is OK, returned 6 items
The HTTP status returned is OK, returned 6 items
The HTTP status returned is OK, returned 6 items
[...]

Issue #790, opened 2025-01-28
Implement an "archive everything" command


rh-hemartin:
We want to provide a feature that allow users to "archive everything" on our CLI. The idea of the command is that it sends cloudevents to the sink as if a change was made to the resources, triggering the archival and/or removal of that resource. Implementation-wise the tool would need to read KubeArchiveConfig's (or accept resource names/kinds), collect the resources, build cloudevents and sent them to the broker as if it was the ApiServerSource.

**Before**

* We need to agree if we read KubeArchiveConfig's or if we accept kind/resource (like `pod dummy-name`)

**Acceptance Criteria**

* The CLI should have a command whose name conveys "archive everything" that makes resources to be archived into KubeArchive.

maruiz93:
In the case of having already some data archived and some that isn't archived yet, do we want to first query the kubearchive API to know what are the resources that aren't archived to prevent having these events sent to the sink in case there are too much or do we prefer to be uncoupled from the kubearchive API, and rely to the user being aware of what they're asking to archive?

I prefer option 2 but I think it's important to explicitly decide that.

> We need to agree if we read KubeArchiveConfig's or if we accept kind/resource (like pod dummy-name)

I like the idea of only interacting with the k8s API, but I think that we should support the same cel expressions that we accept in the KubeArchiveConfig CR.

rh-hemartin:
> In the case of having already some data archived and some that isn't archived yet, do we want to first query the kubearchive API to know what are the resources that aren't archived to prevent having these events sent to the sink in case there are too much or do we prefer to be uncoupled from the kubearchive API, and rely to the user being aware of what they're asking to archive?

I wouldn't contact the KubeArchive API to see what is archived.  If we later find it is a problem then we can make the change.

> I like the idea of only interacting with the k8s API, but I think that we should support the same cel expressions that we accept in the KubeArchiveConfig CR.

Why would we need to support the cel expressions?

maruiz93:
>  Why would we need to support the cel expressions?

Your examples are based on the names of the resources, but I was thinking on generic way of archiving a bunch of resources based on some requirements, the same way we do with the configuration inside KubeArchiveConfig resources.

I don't think that doing the manual archival for one resource at a time makes too much sense and supporting cel expressions gives us some coherence regarding the other ways we have to interact with kubearchive to say what we want to archive.

Considering how the kubernetes API is structured, I'd also provide the name as optional in the case the user wants to only archive one resource, like it was mentioned above:

[...]

PR #874, opened 2025-03-04
Speed up integration test

### LINKED ISSUE #880: Sink does not respond to cloud events in the way that knative-eventing expects
Knative Eventing documents what it expects of Cloud Event recipients [here](https://github.com/knative/specs/blob/main/specs/eventing/data-plane.md#event-delivery). The sink needs to change its 200 Cloud Event received response to 202. It also needs to remove the body from the responses it sends because Knative Eventing expects either no response body, or a Cloud Event in the response body. By not doing this, the knative eventing broker will always retry sending events and this causes the performance issues seen in #870

skoved:

Resolves #880, Related to #870 

## Release Note

```release-note
Fixed a bug where KubeArchive rejected resource updates leading to a performance issue
```

## Notes for Reviewers

Follow knative-eventing spec by sending http 202 for successfully processed cloud events and remove response body from
the http response. 

https://github.com/knative/specs/blob/main/specs/eventing/data-plane.md#event-acknowledgement-and-delivery-retry

Add checks to TestPagination to make sure that TestPagination to make sure the sink-filters
ConfigMap and the ApiServerSource are ready before creating Pods.

Add check to the sink to make so that it can properly handle receiving a Cloud Event that matches a `DeleteWhen` CEL expression when the resource has already been deleted. The sink will see if the resource exists on the cluster and if it does not, it will log that the resource was already deleted, and return an http 202 response.

### DIFF
--- a/cmd/operator/internal/controller/kubearchiveconfig_controller.go
+++ b/cmd/operator/internal/controller/kubearchiveconfig_controller.go
@@ -254,7 +254,7 @@ func (r *KubeArchiveConfigReconciler) reconcileSinkRole(ctx context.Context, kac
 		resource := sourcesv1.APIVersionKindSelector{Kind: kar.Selector.Kind, APIVersion: kar.Selector.APIVersion}
 		resources = append(resources, resource)
 	}
-	return r.reconcileRole(ctx, kaconfig, k9eSinkName, r.getRules(ctx, resources, []string{"list", "delete"}))
+	return r.reconcileRole(ctx, kaconfig, k9eSinkName, r.getRules(ctx, resources, []string{"delete"}))
 }
 
 func (r *KubeArchiveConfigReconciler) reconcileRole(ctx context.Context, kaconfig *kubearchivev1alpha1.KubeArchiveConfig, roleName string, rules []rbacv1.PolicyRule) (*rbacv1.Role, error) {
--- a/cmd/operator/internal/controller/kubearchiveconfig_controller.go
+++ b/cmd/operator/internal/controller/kubearchiveconfig_controller.go
@@ -254,7 +254,7 @@ func (r *KubeArchiveConfigReconciler) reconcileSinkRole(ctx context.Context, kac
 		resource := sourcesv1.APIVersionKindSelector{Kind: kar.Selector.Kind, APIVersion: kar.Selector.APIVersion}
 		resources = append(resources, resource)
 	}
-	return r.reconcileRole(ctx, kaconfig, k9eSinkName, r.getRules(ctx, resources, []string{"delete"}))
+	return r.reconcileRole(ctx, kaconfig, k9eSinkName, r.getRules(ctx, resources, []string{"list", "delete"}))
 }
 
 func (r *KubeArchiveConfigReconciler) reconcileRole(ctx context.Context, kaconfig *kubearchivev1alpha1.KubeArchiveConfig, roleName string, rules []rbacv1.PolicyRule) (*rbacv1.Role, error) {
--- a/cmd/sink/routers/result.go
+++ b/cmd/sink/routers/result.go
@@ -8,12 +8,8 @@ import (
 	"github.com/cloudevents/sdk-go/v2/protocol"
 )
 
-func NewCEResult(statusCode int, err error) protocol.Result {
-	if statusCode >= 200 && statusCode < 400 {
-		// Cloud Event Processed successfully
-		return cloudevents.NewHTTPResult(statusCode, "cloud event processed successfully")
-	} else if err == nil {
-		return cloudevents.NewHTTPResult(statusCode, "an error occurred while processing the cloud event")
-	}
-	return cloudevents.NewHTTPResult(statusCode, "an error occurred while processing the cloud event", err.Error())
+// knative eventing expects that a response to a cloud event has an empty body, so NewHTTPResult should be called with
+// an empty string.
+func NewCEResult(statusCode int) protocol.Result {
+	return cloudevents.NewHTTPResult(statusCode, "")
 }
--- a/cmd/sink/routers/result.go
+++ b/cmd/sink/routers/result.go
@@ -8,7 +8,7 @@ import (
 	"github.com/cloudevents/sdk-go/v2/protocol"
 )
 
-// knative eventing expects that a response to a cloud event has an empty body, so NewHTTPResult should be called with
+// knative eventing expects that a response to a cloud event has an empty body, so NewHTTPResult must be called with
 // an empty string.
 func NewCEResult(statusCode int) protocol.Result {
 	return cloudevents.NewHTTPResult(statusCode, "")
--- a/test/integration/pagination_test.go
+++ b/test/integration/pagination_test.go
@@ -19,10 +19,11 @@ import (
 	"github.com/stretchr/testify/assert"
 	authenticationv1 "k8s.io/api/authentication/v1"
 	corev1 "k8s.io/api/core/v1"
-	errs "k8s.io/apimachinery/pkg/api/errors"
 	rbacv1 "k8s.io/api/rbac/v1"
+	errs "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	sourcesv1 "knative.dev/eventing/pkg/apis/sources/v1"
 )
 
 func TestPagination(t *testing.T) {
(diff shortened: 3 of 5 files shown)


cmd/operator/internal/controller/kubearchiveconfig_controller.go
```diff
@@ -254,7 +254,7 @@ func (r *KubeArchiveConfigReconciler) reconcileSinkRole(ctx context.Context, kac
 		resource := sourcesv1.APIVersionKindSelector{Kind: kar.Selector.Kind, APIVersion: kar.Selector.APIVersion}
 		resources = append(resources, resource)
 	}
-	return r.reconcileRole(ctx, kaconfig, k9eSinkName, r.getRules(ctx, resources, []string{"delete"}))
+	return r.reconcileRole(ctx, kaconfig, k9eSinkName, r.getRules(ctx, resources, []string{"list", "delete"}))
```
ggallen, line 257: It seems like `get` would be a better choice here than `list`.

cmd/sink/routers/result.go
```diff
@@ -8,12 +8,8 @@ import (
...
-		// Cloud Event Processed successfully
-		return cloudevents.NewHTTPResult(statusCode, "cloud event processed successfully")
-	} else if err == nil {
-		return cloudevents.NewHTTPResult(statusCode, "an error occurred while processing the cloud event")
-	}
-	return cloudevents.NewHTTPResult(statusCode, "an error occurred while processing the cloud event", err.Error())
+// knative eventing expects that a response to a cloud event has an empty body, so NewHTTPResult should be called with
+// an empty string.
+func NewCEResult(statusCode int) protocol.Result {
+	return cloudevents.NewHTTPResult(statusCode, "")
```
ggallen, line 14: This method seems like a waste now. Just call cloudevents.NewHTTPResult directly.

test/integration/pagination_test.go
```diff
@@ -50,7 +51,7 @@ func TestPagination(t *testing.T) {
 		retryErr := retry.Do(func() error {
 			_, getErr := clientset.CoreV1().Namespaces().Get(context.Background(), namespaceName, metav1.GetOptions{})
 			if !errs.IsNotFound(getErr) {
-				return errors.New("Waiting for namespace "+namespaceName+" to be deleted")
+				return errors.New("Waiting for namespace " + namespaceName + " to be deleted")
```
ggallen, line 54: This is a white space only change and unnecessary.

test/integration/pagination_test.go
```diff
@@ -19,10 +19,11 @@ import (
 	"github.com/stretchr/testify/assert"
 	authenticationv1 "k8s.io/api/authentication/v1"
 	corev1 "k8s.io/api/core/v1"
-	errs "k8s.io/apimachinery/pkg/api/errors"
 	rbacv1 "k8s.io/api/rbac/v1"
+	errs "k8s.io/apimachinery/pkg/api/errors"
```
ggallen, line 23: This change looks unnecessary.

test/integration/pagination_test.go
```diff
@@ -19,10 +19,11 @@ import (
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	sourcesv1 "knative.dev/eventing/pkg/apis/sources/v1"
```
ggallen, line 26: Shouldn't this be before unstructured?

test/integration/pagination_test.go
skoved: this is the result of running `goimports` on the file after saving the changes I made elsewhere

test/integration/pagination_test.go
skoved: this is the result of running `goimports` on the file after saving the changes I made elsewhere. We agreed at the kubearchive go code should be formatted with `goimports`

test/integration/pagination_test.go
skoved: `sources` comes alphabetically after `meta`

skoved: I can move it. I legitimately don't know if letters come before numbers

test/integration/pagination_test.go
ggallen: How did it ever get committed. Don't we always run `goimports`?

test/integration/pagination_test.go
ggallen: I assumed (which probably makes it wrong) it was alphabetical, using the last portion of the path ("unstructured", in this case) if there was no alias.

cmd/sink/routers/result.go
ggallen: Good point.  Then can you change the comment from "should be" to "must be"?

test/integration/pagination_test.go
skoved: so `goimports` handles ordering imports in a specific order and if I move `sourcesv1` above the `k8s.io` imports, `goimports` moves `sourcesv1` back to the bottom of the import list

ggallen: Works for me.  At least now I know.

test/integration/pagination_test.go
ggallen: Maybe we don't normally run `goimports` on test code?

cmd/operator/internal/controller/kubearchiveconfig_controller.go
skoved: I disagree. Using `list` seems counter intuitive but it lets the sink distinguish between a situation where there is a permissions/networking issue and the object already is deleted. The sink does a list using a field selector, so if the object is deleted, we get an empty list. With `get`, if the object was deleted, the sink gets an error, which is same thing that happens if there is a permissions/networking issue

ggallen: That's what this check does:
```
if !errs.IsNotFound(getErr) {
```

skoved: I'm confused what you're referring to. The code that I described is over [here](https://github.com/kubearchive/kubearchive/blob/651dcce1779e0f1c95e740a18e24471af91eecf9/cmd/sink/routers/routers.go#L213)

cmd/sink/routers/result.go
skoved: done

test/integration/pagination_test.go
skoved: it's possible. I know this isn't the first time I've edited a file and the formatter has made changes to lines that I didn't edit and you comment on it. I do know that the integration test are a little weird though. I have to disable the go plugin for vim when editing the the integration tests because it doesn't like that the package they are in is called `main`. It might break plugins in other editors too but I have no clue

rh-hemartin (approved):

rh-hemartin:
I'm always complaining about release notes, sorry. How about this:

```
- fixes bug where knative eventing assumed every cloud event sent to kubearchive was not received
+ Fixed a bug where KubeArchive rejected resource updates leading to a performance issue
```

maruiz93 (changes requested):

cmd/sink/routers/result.go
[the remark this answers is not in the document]
ggallen: > If that is the case I think it's useful to fill the body explaining the error in case we expect to receive events from different sources than the APIServerSource (e.g., the CLI)

But the body gets sent back to the broker, so how does that help us? Why wouldn't we just log the errors so we can see in our logs what is happening?

skoved:
> I'm always complaining about release notes, sorry. How about this:
> 
> ```
> - fixes bug where knative eventing assumed every cloud event sent to kubearchive was not received
> + Fixed a bug where KubeArchive rejected resource updates leading to a performance issue
> ```

done

cmd/sink/routers/result.go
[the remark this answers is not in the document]
maruiz93: I was thinking on the "archive everything" [feature](https://github.com/kubearchive/kubearchive/issues/790)
If the cli is going to send cloudevents for the sink in that case, do we expect to go through the broker?
I think that for that case, it makes more sense to go directly to the API and notify the cli user if in fails.

What do you think?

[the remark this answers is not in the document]
skoved: I'm going to leave a link to the note from the community meeting on this topic [here](https://docs.google.com/document/d/1cQ0DWoPLI1CAm29vvy4b9iUj_HpPYVG1RqbLhR7NMW0/edit?tab=t.0#bookmark=id.wfu3f5ow3mxe). My understanding is that we decided that at least for now, all responses bodies, including ones for errors, will be empty

test/integration/pagination_test.go
skoved: Thanks for the tip, that fixed the gopls errors I would get in the integration tests

cmd/operator/internal/controller/kubearchiveconfig_controller.go
ggallen: Aren't you using the same dynamic client that is used in `test/integration/operator_test.go`?

skoved: yes, I did not realize that `errs.IsNotFound(getErr)` was a utility function for check those kinds of errors. I will switch to using `get`

maruiz93 (approved):

cmd/operator/internal/controller/kubearchiveconfig_controller.go
skoved: actually, using `errs.IsNotFound(getErr)` I was able to make this work by checking the result of the already existing delete request. Please take a look

ggallen (approved):

PR #878, opened 2025-03-05
Add pprof support for KubeArchive

### LINKED ISSUE #828: add net/http/pprof endpoints to api/operator/sink
To make some performance issues easier to debug it would be useful to have endpoints on the api, operator, and sink that could generate results for cpu and memory profiling of a running kubearchive deployment. This should be something can be turned on/off. It seems that impact on performance of running pprof in production, seems to be relatively low.

**Acceptance Criteria:**

- api, operator, and sink expose pprof http endpoints
- pprof endpoints can be disabled through environment variable

[net/http/pprof](https://pkg.go.dev/net/http/pprof)

rh-hemartin:

Resolves #828 

## Release Note

```release-note
KubeArchive now exposes pprof endpoints by default, check the documentation on how to disable this feature.
```

### DIFF
--- a/cmd/api/main.go
+++ b/cmd/api/main.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"log/slog"
 	"net/http"
+	"net/http/pprof"
 	"os"
 	"os/signal"
 	"syscall"
@@ -34,6 +35,7 @@ const (
 	otelServiceName                   = "kubearchive.api"
 	cacheExpirationAuthorizedEnvVar   = "CACHE_EXPIRATION_AUTHORIZED"
 	cacheExpirationUnauthorizedEnvVar = "CACHE_EXPIRATION_UNAUTHORIZED"
+	enablePprofEnvVar                 = "KUBEARCHIVE_ENABLE_PPROF"
 )
 
 var (
@@ -86,6 +88,21 @@ func NewServer(k8sClient kubernetes.Interface, controller routers.Controller, ca
 
 	router.GET("/livez", controller.Livez)
 	router.GET("/readyz", controller.Readyz)
+
+	if os.Getenv(enablePprofEnvVar) == "true" {
+		router.GET("/debug/pprof/", gin.WrapF(pprof.Index))
+		router.GET("/debug/pprof/cmdline", gin.WrapF(pprof.Cmdline))
+		router.GET("/debug/pprof/profile", gin.WrapF(pprof.Profile))
+		router.POST("/debug/pprof/symbol", gin.WrapF(pprof.Symbol))
+		router.GET("/debug/pprof/symbol", gin.WrapF(pprof.Symbol))
+		router.GET("/debug/pprof/trace", gin.WrapF(pprof.Trace))
+		router.GET("/debug/pprof/allocs", gin.WrapH(pprof.Handler("allocs")))
+		router.GET("/debug/pprof/block", gin.WrapH(pprof.Handler("block")))
+		router.GET("/debug/pprof/goroutine", gin.WrapH(pprof.Handler("goroutine")))
+		router.GET("/debug/pprof/heap", gin.WrapH(pprof.Handler("heap")))
+		router.GET("/debug/pprof/mutex", gin.WrapH(pprof.Handler("mutex")))
+		router.GET("/debug/pprof/threadcreate", gin.WrapH(pprof.Handler("threadcreate")))
+	}
 
 	creds, credsErr := logging.GetKubeArchiveLoggingCredentials()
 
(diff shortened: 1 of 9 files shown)


maruiz93 (approved):

cmd/api/main.go
```diff
@@ -87,6 +89,21 @@ func NewServer(k8sClient kubernetes.Interface, controller routers.Controller, ca
...
+		router.GET("/debug/pprof/symbol", gin.WrapF(pprof.Symbol))
+		router.GET("/debug/pprof/trace", gin.WrapF(pprof.Trace))
+		router.GET("/debug/pprof/allocs", gin.WrapH(pprof.Handler("allocs")))
+		router.GET("/debug/pprof/block", gin.WrapH(pprof.Handler("block")))
+		router.GET("/debug/pprof/goroutine", gin.WrapH(pprof.Handler("goroutine")))
+		router.GET("/debug/pprof/heap", gin.WrapH(pprof.Handler("heap")))
+		router.GET("/debug/pprof/mutex", gin.WrapH(pprof.Handler("mutex")))
+		router.GET("/debug/pprof/threadcreate", gin.WrapH(pprof.Handler("threadcreate")))
+	}
+
```
skoved, line 106: since this is also done in the sink do you think it would make sense to put in this function that would accept router as a param and both the sink and the api could import it?

something like

```
func SetupPprof(router *gin.Engine) {
	...
}
```

cmd/sink/server/server.go
```diff
@@ -34,6 +37,21 @@ func NewServer(controller *routers.Controller) *Server {
...
+		router.POST("/debug/pprof/symbol", gin.WrapF(pprof.Symbol))
+		router.GET("/debug/pprof/symbol", gin.WrapF(pprof.Symbol))
+		router.GET("/debug/pprof/trace", gin.WrapF(pprof.Trace))
+		router.GET("/debug/pprof/allocs", gin.WrapH(pprof.Handler("allocs")))
+		router.GET("/debug/pprof/block", gin.WrapH(pprof.Handler("block")))
+		router.GET("/debug/pprof/goroutine", gin.WrapH(pprof.Handler("goroutine")))
+		router.GET("/debug/pprof/heap", gin.WrapH(pprof.Handler("heap")))
+		router.GET("/debug/pprof/mutex", gin.WrapH(pprof.Handler("mutex")))
+		router.GET("/debug/pprof/threadcreate", gin.WrapH(pprof.Handler("threadcreate")))
+	}
```
skoved, line 53: same as my last comment

cmd/api/main.go
```diff
@@ -34,6 +35,7 @@ const (
 	otelServiceName                   = "kubearchive.api"
 	cacheExpirationAuthorizedEnvVar   = "CACHE_EXPIRATION_AUTHORIZED"
 	cacheExpirationUnauthorizedEnvVar = "CACHE_EXPIRATION_UNAUTHORIZED"
+	enablePprofEnvVar                 = "KUBEARCHIVE_ENABLE_PPROF"
```
skoved, line 38: If you put the pprof setup function into a `pkg` package this `const` could be moved there and then the sink, operator and api could use the shared const

cmd/sink/server/server.go
```diff
@@ -18,6 +19,8 @@ import (
 	"github.com/kubearchive/kubearchive/pkg/middleware"
 )
 
+const enablePprofEnvVar = "KUBEARCHIVE_ENABLE_PPROF"
```
skoved, line 22: same as above

cmd/operator/main.go
```diff
@@ -39,7 +39,10 @@ import (
 	//+kubebuilder:scaffold:imports
 )
 
-const otelServiceName = "kubearchive.operator"
+const (
+	otelServiceName   = "kubearchive.operator"
+	enablePprofEnvVar = "KUBEARCHIVE_ENABLE_PPROF"
```
skoved, line 44: same as above

skoved (commented):
Thanks for setting this up!

cmd/api/main.go
rh-hemartin: I changed it to be a function and moved it under the `observability` package on `pkg`. Let me know how it looks.

cmd/operator/main.go
skoved, line 44: I think you can remove this const since we're now using `observability.EnablePprofEnvVar`

skoved (approved):
looks good

cmd/operator/main.go
rh-hemartin: Ops, my bad. Removed.

PR #882, opened 2025-03-05
switch to linear backoffPolicy

### LINKED ISSUE #881: Getting to the max retry if cloud event delivery fails is too long
as of right now the [broker](https://github.com/kubearchive/kubearchive/blob/main/config/templates/eventing/brokers.yaml) is configured to perform 4 retries, using an exponential backoff, with a 0.5s initial delay between retries.

The exponential backoff calculation is:

```backoffDelay*2^<numberOfRetries>```

This leads to means the last retry takes 8 seconds (if all retries are exhausted).

If we switch to a linear backoff, it will take 2s to get to the last retry (if all retries are exhausted) which is more reasonable.

### PR DESCRIPTION
**Pull request by skoved**

Resolves #881 

## Release Note

```release-note
NONE
```

## Notes for Reviewers

if all retries are exhausted, this reduces the time to reach the last retry from 8s to 2s

### DIFF
--- a/config/templates/eventing/brokers.yaml
+++ b/config/templates/eventing/brokers.yaml
@@ -15,7 +15,7 @@ metadata:
 spec:
   delivery:
     retry: 4
-    backoffPolicy: exponential
+    backoffPolicy: linear
     # ISO 8601 format
     backoffDelay: PT0.5S
     deadLetterSink:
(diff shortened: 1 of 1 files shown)


rh-hemartin (approved):

maruiz93 (approved):

skoved:
rebasing from main to allow this to merge

PR #892, opened 2025-03-06
Add docs for testing Datadog logging integration

### LINKED ISSUE #858: Test logging integration with Datadog logging backend system
We currently tested the logging integration of kubearchive with Splunk, but we should try other well-known logging backend systems.

### Acceptance Criteria

* Add documentation about how to deploy a test Datadog - (it can't be deployed automatically without an account)

maruiz93:

Resolves #858

## Release Note

```release-note
NONE
```

## Notes for Reviewers

To keep the `DEVELOPMENT.md` short I moved the specific instructions for the different backend logging systems in its folder and added a link from the `DEVELOPMENT.md` file.

These are instructions that will help on setting the datadog testing environment when implementing the changes needed to allow this integration in KubeArchive.

I will create and link the related issues to this PR.: https://github.com/kubearchive/kubearchive/issues/893


test/performance/run.sh
```diff
@@ -28,8 +28,8 @@ kubectl create clusterrole view-pods --verb=list,get --resource=pods,pods/status
 kubectl create clusterrolebinding view-pods --clusterrole view-pods --user system:serviceaccount:default:default --dry-run=client -o yaml | kubectl apply -f -
 export SA_TOKEN=$(kubectl create token default --namespace default)
 
-python3 -m venv venv
-./venv/bin/python -m pip install -r ${SCRIPT_DIR}/requirements.txt
+python3.11 -m venv venv
+./venv/bin/python3.11 -m pip install -r ${SCRIPT_DIR}/requirements.txt
```
rh-hemartin, line 32: I don't have this version installed, if `python3` does not work for you change it to `python`. These days `python` is an alias for `python3` or whatever version you have installed.

integrations/logging/datadog/README.md
```diff
@@ -0,0 +1,49 @@
+## Datadog
+
+1. Create a free 14-days Datadog account at [datadoghq.com](https://www.datadoghq.com/free-datadog-trial/#)
```
rh-hemartin, line 3: What happens after 14 days? We just lose access?

maruiz93: If you go with the free account yes

test/performance/run.sh
maruiz93: I'm sorry, I pushed this file by mistake...

maruiz93: Removed the file from the commit.

integrations/logging/datadog/README.md
rh-hemartin: Hm... I don't like that it does not have a free time-unlimited mode for testing :(

rh-hemartin (approved):
I guess its fine to introduce the datadog integration, but I don't like that it does not have an option as elasticsearch or splunk to avoid having an account and potentially paying with to integrate with it.

integrations/logging/datadog/README.md
maruiz93: If we want to go that route, we could request a [Datadog sandbox account](https://docs.datadoghq.com/developers/integrations/api_integration/#build-your-integration) for development, but if I understand correctly, that means being part of the [Datadog Partner Network](https://partners.datadoghq.com/s/login/?ec=302&startURL=%2Fs%2F)

Maybe it's something we want to investigate in the future, but I wouldn't do it right now.

maruiz93:
> I guess its fine to introduce the datadog integration, but I don't like that it does not have an option as elasticsearch or splunk to avoid having an account and potentially paying with to integrate with it.

We don't need to "pay" for it, we just have to make sure that we are compatible with their API so that if a KubeArchive user already uses datadog, it can make kubearchive retrieve the logs from there.

Deploying it locally it's just for making sure that everything works "manually" as we won't be able to fully automate a test by deploying a datadog instance ad hoc in the tests as we do with Splunk.

maruiz93:
> > > I guess its fine to introduce the datadog integration, but I don't like that it does not have an option as elasticsearch or splunk to avoid having an account and potentially paying with to integrate with it.
> > 
> > 
> > We don't need to "pay" for it, we just have to make sure that we are compatible with their API so that if a KubeArchive user already uses datadog, it can make kubearchive retrieve the logs from there.
> > Deploying it locally it's just for making sure that everything works "manually" as we won't be able to fully automate a test by deploying a datadog instance ad hoc in the tests as we do with Splunk.
> 
> Yes, my statement is not precise. I meant that after a while no maintainer will be able to deploy it, because all of our free trials will be expired, so in certain way we will be required to pay if we want to maintain the integration in case they change something in their API.

well you can always create another account, it's a matter of 2 mins (in fact I create two, the second one to replicate the process and make sure the instructions were accurate)

rh-hemartin:
> well you can always create another account, it's a matter of 2 mins (in fact I created two, the second one to replicate the process and make sure the instructions were accurate)

Ah, my bad then. I thought email reuse would be restricted. So you can register again for the 14 day free trial with the same email (or an + alias)?

ggallen:
> I guess its fine to introduce the datadog integration, but I don't like that it does not have an option as elasticsearch or splunk to avoid having an account and potentially paying with to integrate with it.

Agreed. But this is not our problem, it's Datadog's. I think it perfectly reasonable to tell potential users they have to create a free account to try it out.  It's really the only way.

And I think we definitely want this integration.

skoved (approved):

PR #904, opened 2025-03-12
remove deleteWhen so splunk correctly stores logs for pods with multiple containers

### LINKED ISSUE #905: TestDefaultContainer integration test fails when getting logs for the second container
There is discussion about the issue in [this thread in the kubearchive slack channel](https://kubernetes.slack.com/archives/C07MB5YBVCL/p1741713427166909?thread_ts=1741368177.687929&cid=C07MB5YBVCL).

The simple fix to unblock CI is to remove the `deleteWhen` clause from the KAC in `TestDefaultContainer`, but there maybe a more robust change that involves changes to the sink, like queuing delete operations for some period of time.

skoved:

Resolves #905

## Release Note

```release-note
NONE
```

## Notes for Reviewers

Splunk has issues with the logs for the container `second` because it seems like the pod gets deleted too quickly. For now the fix is removing the `deleteWhen` clause in the test, but a more robust change maybe implemented later

### DIFF
--- a/test/integration/logging_test.go
+++ b/test/integration/logging_test.go
@@ -20,8 +20,8 @@ import (
 	"github.com/kubearchive/kubearchive/test"
 	authenticationv1 "k8s.io/api/authentication/v1"
 	corev1 "k8s.io/api/core/v1"
-	errs "k8s.io/apimachinery/pkg/api/errors"
 	rbacv1 "k8s.io/api/rbac/v1"
+	errs "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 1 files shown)


skoved:
The change to the formatting around the `+` and the ordering of the imports were done automatically by `goimports` when I changed this file

rh-hemartin:
Just for the record, its fluentbit which has problems, but they are visible at Splunk that does not receive metadata from fluentbit because the pod gets deleted too quickly. Related ticket: https://github.com/fluent/fluent-bit/issues/2670

rh-hemartin (approved):

maruiz93 (approved):

PR #912, opened 2025-03-14
Enhance operator tests.

### LINKED ISSUE #863: Enhance operator tests.
The operator integration tests were added in https://github.com/kubearchive/kubearchive/issues/780. This was a start but not complete.

The following should be done to enhance/expand these tests.

- New tests should be added that add a global `KubeArchiveConfig`  and a local one.
- The tests should check to make sure the "sink-filters" `ConfigMap` contains the correct number of entries.
- Move the cleanup method of the integration tests to a shared method in `integration/main.go`.

ggallen:

Resolves #863

## Release Note

```release-note
NONE
```

## Notes for Reviewers

Also refactored the integration tests and added some speed enhancements.

The current test run had the integration tests taking 78 seconds to run.

### DIFF
--- a/test/integration/operator_test.go
+++ b/test/integration/operator_test.go
@@ -8,119 +8,63 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 	"testing"
 	"time"
 
 	"github.com/avast/retry-go/v4"
-	kubearchivev1alpha1 "github.com/kubearchive/kubearchive/cmd/operator/api/v1alpha1"
 	"github.com/kubearchive/kubearchive/test"
-	corev1 "k8s.io/api/core/v1"
 	errs "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
-type Test int
-
 const (
-	emptyKAC Test = iota
-	nonEmptyKAC
-)
-
-const (
-	kacName     = "kubearchive"
-	a13eName    = kacName + "-a13e"
-	sinkName    = kacName + "-sink"
+	a13eName    = test.KACName + "-a13e"
+	sinkName    = test.KACName + "-sink"
 	filtersName = "sink-filters"
 )
 
-func TestEmptyKAC(t *testing.T) {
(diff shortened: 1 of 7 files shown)


test/integration/operator_test.go
```diff
@@ -29,14 +32,13 @@ const (
...
-	a13eName    = kacName + "-a13e"
-	sinkName    = kacName + "-sink"
+	a13eName    = test.KACName + "-a13e"
+	sinkName    = test.KACName + "-sink"
 	filtersName = "sink-filters"
 )
 
 func TestEmptyKAC(t *testing.T) {
-	runTest(t, emptyKAC, map[string]any{})
+	runTest(t, "emptyKAC", map[string]any{})
```
maruiz93, line 41: Is there any reason to use this approach for running different versions of a test instead of the one that we use in the unit tests with [a map to store the different test cases](https://go.dev/wiki/TableDrivenTests#using-a-map-to-store-test-cases)?

I think it's the standard way to put different use cases of the same "feature" and it should be a small change from what you have right now.

test/integration/operator_test.go
```diff
@@ -210,18 +166,66 @@ func checkResourcesAfterDelete(t testing.TB, namespace string, testName Test) {
...
+			{
+				"selector": map[string]string{
+					"apiVersion": "v1",
+					"kind":       "Pod",
+				},
+				"archiveWhen":  "status.phase == 'Succeeded'",
+			},
+		},
+	}
+	localres := map[string]any{
```
rh-hemartin, line 186: My local tests indicate that if the localres is empty, the local namespace is not configured for archival because it is not included in the `sink-filters` so the sink does not know that it should archive anything there, as if was not enabled in the first place. So add a test to confirm this, so with an empty `localres`.

ggallen: Good catch. I think you're right. I'll check this as soon as I get back.

ggallen: This will be handled in https://github.com/kubearchive/kubearchive/issues/948.

test/integration/operator_test.go
ggallen: @maruiz93, the change is made. Can you review again?

rh-hemartin (approved):

maruiz93 (approved):

test/integration/operator_test.go
maruiz93: lgtm!

PR #913, opened 2025-03-14
Old versions of resources do not get written to db

### LINKED ISSUE #817: Sink should handle out of order events
When the sink writes a resource to the database, it should check that the version of the resource it is writing to the database is newer that the version that already exists in the database (if there already is an entry for the resource in the database). If the version of the resource that the sink has is older than the version in the database, the resource in the database should not be updated. This can be accomplished by adding a `WHERE` clause to the insert query for the resource table.

We need to make sure this is correct, but the resource version might be usable for this **IF** it always increases when the resource changes. Otherwise we can compare updated timestamps to see which one is the newer version.

**Acceptance Criteria:**

- If the sink attempts to write an older version of the resource into the database, the entry for the resource is not updated

skoved:

Resolves #817 

## Release Note

```release-note
sink does not update resources in the database with older version of that resource
```

## Notes for Reviewers

This change adds `cluster_updated_ts` which is the timestamp for the last update of the resource on the cluster. This can be calculated by iterating over the elements in the `metadata.managedFields` and finding the most recent `time` field

The Insert query for resources now includes a `WHERE` clause that checks if the current `cluster_updated_ts` in the database is older than the new `cluster_updated_ts`. If it is, the resource is updated. If it is not, the resource is left as is and the sink responds with a http 202 response to the cloud event

### DIFF
--- a/cmd/sink/k8s/objects.go
+++ b/cmd/sink/k8s/objects.go
@@ -1,0 +1,23 @@
+// Copyright KubeArchive Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package k8s
+
+import (
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+// GetLastUpdateTs returns the last time obj was updated on the cluster
+func GetLastUpdateTs(obj *unstructured.Unstructured) time.Time {
+	fields := obj.GetManagedFields()
+	var latest *metav1.Time
+	for _, elem := range fields {
+		if latest == nil || latest.Before(elem.Time) {
+			latest = elem.Time
+		}
+	}
+	return latest.Time
+}
--- a/cmd/sink/routers/routers.go
+++ b/cmd/sink/routers/routers.go
@@ -18,6 +18,7 @@ import (
 	"github.com/cloudevents/sdk-go/v2/protocol"
 	"github.com/gin-gonic/gin"
 	"github.com/kubearchive/kubearchive/cmd/sink/filters"
+	"github.com/kubearchive/kubearchive/cmd/sink/k8s"
 	"github.com/kubearchive/kubearchive/cmd/sink/logs"
 	"github.com/kubearchive/kubearchive/pkg/abort"
 	"github.com/kubearchive/kubearchive/pkg/database"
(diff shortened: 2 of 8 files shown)


rh-hemartin:
Hey, the release note again:

```
sink does not update resources in the database with older version of that resource
```

to

```
KubeArchive now prevents older changes to overwrite newer ones in case their are received out of order.
```

maruiz93:
> Hey, the release note again:
> 
> ```
> sink does not update resources in the database with older version of that resource
> ```
> 
> to
> 
> ```
> KubeArchive now prevents older changes to overwrite newer ones in case they are received out of order.
> ```

I'm wondering, is there any reason to prefix every change in `KubeArchive` with the prefix `KubeArchive now`?
I'm my personal opinion that is boilerplate that doesn't provide actual content to the sentence, so maybe we could go with somehing like:

- [verb] + details

- Discard events with a resource in an older state than the archived.

Something to think about is if we want to mention the component on the Release note or not, I've seen that we haven't been consistent with that. In this case for example, with the propose sentences we are not mentioning the `sink` but in the last Release we have sentences like: `KubeArchive Sink component now fail if it can't connect to the database`.

I'm mentioning this because I've seen it here but maybe we can create an issue to discuss it and create a "standard" so people now what the team expects with regards to the release notes format.

maruiz93 (approved):

skoved:
I think I agree that `KubeArchive now` is filler if it appears at the beginning of every release-note. I've still updated the release note to be more clear

cmd/sink/k8s/objects.go
```diff
@@ -0,0 +1,23 @@
...
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+// GetLastUpdateTs returns the last time obj was updated on the cluster
+func GetLastUpdateTs(obj *unstructured.Unstructured) time.Time {
+	fields := obj.GetManagedFields()
+	var latest *metav1.Time
```
rh-hemartin, line 16: This looks like it can be `nil` since its a pointer, if we are sure this won't be `nil` under any case its fine, if not we should do something with it. Maybe a default value.

rh-hemartin (approved):

cmd/sink/k8s/objects.go
skoved: I don't _think_ that we can get a resource where the `metadata.managedFields` list is empty, but to be safe I'm testing if I can set it to the zero value, which is Jan 1, year 1, 00:00:00.000000000 UTC. This way latest can't be nil

skoved: setting latest to the zero value for `metav1.Time` does not break the logic AND guarantees that `lastest` won't be `nil`

skoved:
rebasing to merge this PR

PR #916, opened 2025-03-17
Increase the work_mem parameter from 4MB (default) to 64MB

### LINKED ISSUE #901: Increase the work_mem for kubearchive schema in PostgreSQL
By default PostgreSQL [work_mem](https://www.postgresql.org/docs/current/runtime-config-resource.html#GUC-WORK-MEM) is 4MB, and if this parameter increases, then the queries that we do perform a lot better for large datasets.

As one of the main features of kubearchive is to query the archived resources, I think that we should increase the maximum memory consumed by the queries. 40 MB could be a good starting point, then each user can modify this parameter as needed.

### Acceptance Criteria:

* Increase `work_mem` to ~40MB~ 64MB

maruiz93:

Resolves #901 

## Release Note

```release-note
NONE
```

## Notes for Reviewers

There is a cosmetic change of a tab to spaces that was done automatically by `pg_dump`.
I'm not sure why there was a tab instead of a set of spaces previously.

To prevent further errors like this, I've added the change suggested by Sam to get the IDE configured to do spacing instead of tabbing in the SQL files.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -22,3 +22,7 @@ indent_size = 2
 
 [{*.go,go.mod,go.sum}]
 indent_style = tab
+
+[*.sql]
+indent_style = space
+indent_size = 4
--- a/integrations/database/postgresql/kubearchive.sql
+++ b/integrations/database/postgresql/kubearchive.sql
@@ -44,6 +44,7 @@ SET row_security = off;
 --
 
 ALTER DATABASE kubearchive SET "TimeZone" TO 'UTC';
+ALTER DATABASE kubearchive SET work_mem TO '64MB';
 
 
 \connect kubearchive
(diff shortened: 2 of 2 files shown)


skoved:
>  I'm not sure why there was a tab instead of a set of spaces previously.

I'm not 100% sure either, but to avoid this in the future it might be worth adding

```
[*.sql]
indent_style = space
indent_size = 4
```

to `.editorconfig`

maruiz93:
> > I'm not sure why there was a tab instead of a set of spaces previously.
> 
> I'm not 100% sure either. My guess is that this might happen in a senario where you already had a go file opened, so the editor was using the config for `*.go` files, since we don't have anything in `.editorconfig` for `*.sql` files. The following change to `.editorconfig` might prevent this in the future:
> 
> ```
> [*.sql]
> indent_style = space
> indent_size = 4
> ```

good point!! I'll add it

skoved (approved):

rh-hemartin (approved):

mafh314 (approved):

rh-hemartin:
@maruiz93 this is basically approved, there is one small change you need to do (introduce endline on a file).

PR #921, opened 2025-03-18
Refactor database to separate write and read functions

### LINKED ISSUE #823: Refactor DBInterface
Go encourages to use just the interface needed, with no extra functionality.
Currently, both api and sink "need" `DBInterface` but that's not true, because the api only need the "read" methods while the sink only need the "write" methods.

We should remove the `DBInterface` in favor of a `DBReader` and a `DBWriter` interface.

This shouldn't affect any implementation.

### Acceptance Criteria

- The api component relies on a DBReader interface with the read methods
- The sink component relies on a DBWriter interface with the write methods

maruiz93:

Resolves #823 

## Release Note

```release-note
NONE
```

## Notes for Reviewers

This change covers the following:
* Separation of DBInterface into DBWriter and DBReader to follow the "Interface Seggregation" Principle from SOLID recommended by Go, avoiding the so called "Interface Pollution".
* The equivalent of DBInterface is now a new interface called `Database` composed by `DBWriter` and `DBReader`
  * ~The common methods like `Ping` or `CloseDB` are duplicated now, but their default implementation is unified under `DBCommonImpl`.~
* ~New interface `sqlDatabase` composed of `Database` and getters of the different `DBSelector`, `DBFilter`... implementations -- This allows the test suite to test the actual implementation of the engine and not the default one. This is needed if we override any of the DatabaseImpl methods.
The methods are "private" (i.e, unexported) as they should only be used in the package by engine implementations and the DBReader/DBWriter test suite.~ --> https://github.com/kubearchive/kubearchive/pull/934
* Separate the reader and writer pieces in different files to improve readability of the code
* ~The actual database connection is setup when `NewReader` or `NewWriter` is called instead of on loading the different engine implementations. -- This makes more sense. Now in the unit tests we just setup the mock connection from the beginning instead of overriding it.~ --> https://github.com/kubearchive/kubearchive/pull/934
* Renamed `database_test` to `reader_test` as we are currently just doing unit tests for the reader interface, if we end up creating a test suite for the writer we could do it in a separate file `writer_test.go`
* ~The Database instance should be a single instance so I implemented the [Singleton Pattern](https://refactoring.guru/design-patterns/singleton/go/example#example-1) to run the `getDatabaseByEnv` only once~ --> https://github.com/kubearchive/kubearchive/pull/932
* Remove the `README.md` file as we have that information duplicated in our [docs](https://kubearchive.github.io/kubearchive/main/integrations/database.html#_adding_a_new_database)


rh-hemartin:
I created this diff from `main` that takes a different approach at the same idea: make the API and the Sink don't care about the underlying implementation and just depend on the methods they use. Take a look at it and let me know what you think:

https://gist.github.com/rh-hemartin/793aa5bd8daba59e883521dd6b659de6

maruiz93:
> I created this diff from `main` that takes a different approach at the same idea: make the API and the Sink don't care about the underlying implementation and just depend on the methods they use. Take a look at it and let me know what you think:
> 
> https://gist.github.com/rh-hemartin/793aa5bd8daba59e883521dd6b659de6

If I understand correctly you are declaring the `Reader` and `Writer` interfaces in the `sink` and the `api` respectively. 
I think that it makes more sense to declare them in the same package they are implemented, the reason is that you keep the interfaces duplicated (`DBInterface` need to keep up with the changes on `Reader` and `Writer` interfaces as it's not composed by them)

Having the `Database` former `DBInterface` composed by the `Reader` and `Writer` makes the need of a Database to implement `Reader` and `Writer` more explicit.

Also, keep in mind that with this PR I'm dong more changes in this refactor on purpose, because I'm suggesting also different changes in the implementation to get some benefits from them (detailed in the description). Are you suggesting not to do those changes? Or is this something just a proposal to have the `Reader` and `Writer` definitions moved out from the `pkg/database` and not doing it's composition on the `Database` interface?

rh-hemartin:
> I think that it makes more sense to declare them in the same package they are implemented

After a second look at it, I agree, it makes more sense to define them in the `database` package.

> Are you suggesting not to do those changes? 

Kind of. The changes segregate the implementation, as well as the interface, which leads to complications (duplication of functions that now require a suffix `R` or `W`). We can segregate the interface without segregating about the implementation, which is part of why interfaces are useful.

Other changes you implementing here are not related to segregation of interfaces as far as I understand, like the singleton or making changes so the tests can use the implementation.

I have another diff that defines the same interfaces I suggested but within the `database` package and also makes `Database` implement these two interfaces.

[diff.txt](https://github.com/user-attachments/files/19341123/diff.txt)

maruiz93:
> > I think that it makes more sense to declare them in the same package they are implemented
> 
> After a second look at it, I agree, it makes more sense to define them in the `database` package.
> 
> > Are you suggesting not to do those changes?
> 
> Kind of. The changes segregate the implementation, as well as the interface, which leads to complications (duplication of functions that now require a suffix `R` or `W`). We can segregate the interface without segregating the implementation, which is part of why interfaces are useful.
> 
> Other changes you are implementing here are not related to segregation of interfaces as far as I understand, like the singleton or making changes so the tests can use the implementation.
> 
> I have another diff that defines the same interfaces I suggested but within the `database` package and also makes `DBInterface` implement these two interfaces.
> 
> [diff.txt](https://github.com/user-attachments/files/19341123/diff.txt)

Fair point, having just one default implementation of Database instead of one per Reader and Writer makes sense:
* Prevents repeated methods like `PingR` and `PingW`
* We don't have an scenario were the implementation of the Reader or the Writer alone makes sense in the kubearchive environment.

I've made some changes to adjust to that while keeping the points that I made in the desc (I'm going to do a little adjustment for the mention of the implementation of common methods, but most of it still applies)

Take a look and tell me what you think.

pkg/database/database.go
```diff
@@ -5,408 +5,103 @@ package database
...
 
-type Database struct {
-	DB       *sqlx.DB
-	Flavor   sqlbuilder.Flavor
-	Selector facade.DBSelector
-	Filter   facade.DBFilter
-	Sorter   facade.DBSorter
-	Inserter facade.DBInserter
-	Deleter  facade.DBDeleter
+type DatabaseImpl struct {
```
skoved, line 32: does `DatabaseImpl` need to be exported from the database package? If the sink is using `DBWriter` and the api is using `DBReader`, it seems like we could leave `DatabaseImpl` type unexported

```suggestion
type databaseImpl struct {
```

pkg/database/mariadb.go
```diff
@@ -9,15 +9,30 @@ import (
...
 	"github.com/kubearchive/kubearchive/pkg/database/facade"
 )
 
 func init() {
-	RegisteredDatabases["mariadb"] = NewMariaDBDatabase
+	RegisteredDatabases["mariadb"] = NewMariaDBDatabase()
 	RegisteredDBCreators["mariadb"] = NewMariaDBCreator
 }
 
+type MariaDBDatabase struct {
```
skoved, line 20: same as previous comment

```suggestion
type mariaDBDatabase struct {
```

pkg/database/postgresql.go
```diff
@@ -11,17 +11,33 @@ import (
...
 )
 
 func init() {
-	RegisteredDatabases["postgresql"] = NewPostgreSQLDatabase
+
+	RegisteredDatabases["postgresql"] = NewPostgreSQLDatabase()
 	RegisteredDBCreators["postgresql"] = NewPostgreSQLCreator
 }
 
+type PostgreSQLDatabase struct {
```
skoved, line 25: same as previous comment

```suggestion
type postgreSQLDatabase struct {
```

pkg/database/postgresql.go
```diff
@@ -11,17 +11,33 @@ import (
...
+	*DatabaseImpl
+}
+
+func NewPostgreSQLDatabase() *PostgreSQLDatabase {
```
skoved, line 29: same as previous comment

```suggestion
func newPostgreSQLDatabase() *PostgreSQLDatabase {
```

skoved (commented):
I think separating the Read and Write interfaces is a good idea. With these changes, I think that there are some types and variables that don't need to be exported by the `database` package anymore

pkg/database/database.go
```diff
@@ -5,408 +5,103 @@ package database
...
-			} else {
-				ownersUuids = append(ownersUuids, row.Uuid)
+var once sync.Once
+var singleDatabase Database
+
+// getSingleDatabase implements singleton pattern to prevent executing getDatabaseByEnv more than once
+func getSingleDatabase() (Database, error) {
+	var err error
+	if singleDatabase == nil {
+		once.Do(func() {
```
rh-hemartin, line 69: I think these two lines are redundant. `sync.Once` already keeps track of when it was called, so there is no need to check if `singleDatabase` is `nil`. I would remove the variable `once` or remove the check against `nil`.

maruiz93:
Based on @skoved comment of unexporting some of the structs and functions (thanks for pointing that out) I've noticed that `FakeDatabase` was composed by `DatabaseImpl` to comply with the `Database` interface and that shouldn't be it, it's a database but it's *not a SQL database* so I created a "private" new interface called `sqlDatabase` and moved the getters of sql specific stuff there.

Then, I also realized that our test suite it's only for the sqlDatabase implementations, so I also renamed `reader_test.sql` to `sqlreader_test.sql`.

docs/modules/ROOT/pages/design/database-interface.adoc
```diff
@@ -15,17 +15,29 @@ so they can be plugged in and out of the different queries.
...
 
-To add a new DBMS the following interfaces must be implemented:
+To add a new database engine the following interfaces must be implemented:
 
-`DBInterface`::
-Include the methods that should be used by the rest of the KubeArchive application.
-Encapsulates the business logic related to the different read/write database operations:
-`QueryResources`, `WriteResources`, `WriteUrls`, `QueryLogUrls`, etc.
+`Database`::
+Composed by `DBWriter` and `DBReader` described below.
```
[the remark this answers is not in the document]
maruiz93: Fixed!

pkg/database/database.go
maruiz93: Fixed! Note https://github.com/kubearchive/kubearchive/pull/921#issuecomment-2741147390

pkg/database/mariadb.go
maruiz93: Applied!

pkg/database/postgresql.go
maruiz93: Applied!

pkg/database/postgresql.go
maruiz93: Applied!

rh-hemartin:
I think you are getting to conclusions because of how the interfaces are organized, while we are the ones that organize interfaces. Instead of differentiating SQL databases, we should move `setConn` to the `Database` interface because all databases we support are SQL. Also, differentiating between databases (and then opening the way to implement other databases types) was not in the scope of this PR.

I'm OK with a redesign of DB system but we need to discuss about it first.

maruiz93:
> Instead of differentiating SQL databases, we should move `setConn` to the `Database` interface because all databases we support are SQL

I don't agree. Note that the `setConn` receives a *sqlx.DB instance, so even if a non-sql database has a connection it's going to be using a different type of connection.

Also, note that you could potentially do the most "dummy" database based on a filesystem that doesn't need a connection.

A thing that could be done to make it more clear is renaming `setConn` to `setSQLConn` which reinforces the idea of setting a `sql` connection and not another one. I think that could be an improvement.

> I think you are getting to conclusions because of how the interfaces are organized, while we are the ones that organize interfaces.

I don't think so. I agree that we are the ones that organize the interfaces and that's why I'm proposing a reorganization that from my point of view, makes more sense.

> I'm OK with a redesign of DB system but we need to discuss about it first.

Sure, do you want to discuss it in a forum different from a PR? I think that the PR is the best place to see what someone is proposing and discuss it on top of the code.

To be honest, when I started this refactor I didn't think that it's going to come to this amount of changes, but one thing lead to another. 
I can try to split this PR into different ones, or at least different commits if you think that makes things more clear for the review.

rh-hemartin:
I am challenging the idea that we need to generalize our database system to non-sql databases here in this specific PR related with the specific issue you link. We started by just splitting an interface and we are now generalizing the system to support requirements that we don't even have.

rh-hemartin (approved):

maruiz93:
> I am challenging the idea that we need to generalize our database system to non-sql databases here in this specific PR related with the specific issue you link. We started by just splitting an interface and we are now generalizing the system to support requirements that we don't even have.

ok I'll do a separate PR for that part

maruiz93:
> > I am challenging the idea that we need to generalize our database system to non-sql databases here in this specific PR related with the specific issue you link. We started by just splitting an interface and we are now generalizing the system to support requirements that we don't even have.
> 
> ok I'll do a separate PR for that part

@rh-hemartin @skoved to prevent getting the PRs stuck due to covering too much, I've narrowed the scope of this one to the segregation of `Database` interface into `DBWriter` and `DBReader` and created the following PRs with their respective issues explaining the reasoning behind so we have a place to discuss the "intent" behind the PR as well as having smaller PRs easier to review:

* https://github.com/kubearchive/kubearchive/pull/932
* https://github.com/kubearchive/kubearchive/pull/934
* https://github.com/kubearchive/kubearchive/pull/935
* https://github.com/kubearchive/kubearchive/pull/937

PR #922, opened 2025-03-19
Fix GolangCI issues and schedule scan each month

### LINKED ISSUE #910: Fix golangci-lint issues
If you run `golangci-lint run`  locally you'll see the following warnings

```
pkg/cel/cel.go:34:3: SA1019: cel.Declarations is deprecated: use FunctionDecls and VariableDecls or FromConfig instead. (staticcheck)
		cel.Declarations(
		^
cmd/sink/filters/filters.go:214:23: SA1019: toolsWatch.NewRetryWatcher is deprecated: use NewRetryWatcherWithContext instead. (staticcheck)
	retryWatcher, err := toolsWatch.NewRetryWatcher("1", &cache.ListWatch{WatchFunc: watcher})
	                    ^
cmd/sink/filters/filters_test.go:148:25: SA1019: toolsWatch.NewRetryWatcher is deprecated: use NewRetryWatcherWithContext instead. (staticcheck)
			retryWatcher, err := toolsWatch.NewRetryWatcher("1", &cache.ListWatch{WatchFunc: watcher})
			                    ^
cmd/operator/internal/controller/kubearchiveconfig_controller.go:478:121: (*KubeArchiveConfigReconciler).reconcileA13e - result 0 (*knative.dev/eventing/pkg/apis/sources/v1.ApiServerSource) is never used (unparam)
func (r *KubeArchiveConfigReconciler) reconcileA13e(ctx context.Context, resources []sourcesv1.APIVersionKindSelector) (*sourcesv1.ApiServerSource, error) {
```

These warnings seem relatively straightforward to fix and we should especially move off of deprecated function calls

rh-hemartin:

Resolves #910 

## Release Note

```release-note
NONE
```

### DIFF
--- a/.github/workflows/golangci_lint.yml
+++ b/.github/workflows/golangci_lint.yml
@@ -4,6 +4,9 @@
 name: golangci-lint
 run-name: Linting Tests (golangci-lint)
 on:
+  schedule:
+    # 9:25 UTC every 1st of month
+    - cron: '25 09 1 * *'
   push:
     branches:
       - 'main'
@@ -19,7 +22,7 @@ jobs:
     uses: ./.github/workflows/log_run_details.yml
   golangci:
     # See https://github.com/golangci/golangci-lint-action
-    name: "golangci-lint security testing"
+    name: GolangCI Lint
     runs-on: ubuntu-latest
     steps:
       - name: Check out repository code
@@ -36,4 +39,22 @@ jobs:
         uses: golangci/golangci-lint-action@v6
         with:
           version: latest
-          only-new-issues: true
+          # When is scheduled, show all issues
+          only-new-issues: ${{ github.event_name != 'schedule' }}
+      - name: Notify golangci-lint failed
+        uses: jayqi/failed-build-issue-action@v1
+        if: failure() && github.event_name == 'schedule' && github.repository == 'kubearchive/kubearchive'
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          title-template: GolangCI Linting found errors
+          body-template: |
+            The scheduled workflow of GolangCI Lint found some problems. We run this scheduled
+            against the whole codebase, so when a PR is submitted authors do not need to deal with
+            problems they did not introduce. However we need to solve those problems from time to time.
+            The URL of the failed job is:
+
+            https://github.com/kubearchive/kubearchive/actions/runs/{{ runId }}
+
+            Run `golanci-lint` to see the errors locally, visit the worfklow run, make sure the problems
+            you found locally match. Investigate if they do not match. Solve the problems and open a
+            pull request to solve them.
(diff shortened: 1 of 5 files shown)


.github/workflows/golangci_lint.yml
```diff
@@ -36,4 +39,5 @@ jobs:
         uses: golangci/golangci-lint-action@v6
         with:
           version: latest
-          only-new-issues: true
+          # When is scheduled, show all issues
+          only-new-issues: ${{ github.event_name == 'schedule' && false || true }}
```
skoved, line 43: If it works it works, but I don't quite understand how the logic works to toggle between true and false

skoved (approved):

.github/workflows/golangci_lint.yml
rh-hemartin: Yes, GitHub decided that this confusing syntax will be their ternary operator... So basically it means

```
false if github.event_name == 'schedule' else true 
```

ggallen: > ${{ github.event_name == 'schedule' && false || true }}

It's actually the equivalent syntax of:
```
${{ github.event_name == 'schedule' ? false : true }}
```
But the re-use of `||` and `&&`  (and also `true` and `false` in this case) make it particularly ugly.

Can you do this?
```
          only-new-issues: ${{ github.event_name != 'schedule' }}
```
Or will that only give you a boolean value but won't turn it into the string?

rh-hemartin: Your suggestion works well @ggallen , so I'm adopting it.

.github/workflows/golangci_lint.yml
```diff
@@ -36,4 +39,5 @@ jobs:
         uses: golangci/golangci-lint-action@v6
         with:
           version: latest
-          only-new-issues: true
+          # When is scheduled, show all issues
+          only-new-issues: ${{ github.event_name != 'schedule' }}
```
skoved, line 43: I've been thinking about how we can be notified about when the scheduled version of the action runs. I found the [`Failed Build Issue` Action](https://github.com/kubearchive/kubearchive/pull/922/files) which can create an issue for us if the `golangci-lint` fails. The action needs `Read and write permissions`

```suggestion
          only-new-issues: ${{ github.event_name != 'schedule' }}
      - name: Notify golangci-lint failed
        uses: jayqi/failed-build-issue-action@v1
        if: failure() && github.event_name != 'schedule' && github.repository == 'kubearchive/kubearchive'
        with:
          github-token: ${{secrets.GITHUB_TOKEN}}
          title-template: "golangci-lint action failed"
          # by default the action will create a comment on the newest issue with this label so it should get its own label to use
          label-name: "golangci-lint"
```

rh-hemartin: I added something similar.

maruiz93 (approved):

PR #923, opened 2025-03-19
Inequality filter

### LINKED ISSUE #902: Modify the inequality label filter
The filter for doing the equivalent to `kubectl get pods -l ‘key1!=value1,key2!=value2’` isn't very performant as it doesn't use any of the index in place.

We should update the query, changing this:

```
WHERE NOT data->'metadata'->'labels'
 @> ANY(ARRAY['{"key1":"value1"}', '{"key2":"value2"}']::jsonb[])
```

to this:

```
WHERE uuid NOT IN (
SELECT uuid FROM resource
WHERE kind = '<kind>'
AND api_version = '<api_version>'
AND data->'metadata'->'labels' @> ANY(ARRAY['{"kay1":"value1"}','{"key2":"value2"}'])

```

### Acceptance Criteria:

* The label filter for inequality fits the the described filter

maruiz93:

Resolves #902

## Release Note

```release-note
NONE
```

## Notes for Reviewers

Still in draft to:
- [x] Add the extra clause as a parameter for all the "complex" filters (labels related)
- [x] Also I did it on top of #921 so I need to address that PR before~
- [x] Add docs to reflect the change on the DBFilter

Check the specific last commit to see the actual changes in the filter.

The idea is that for the "complex" filters we allow to pass a WhereClause with the kind, apiversion and namespace (if applies) filters of the main query so it can be attached to a subquery if it is done.

My intention with this was to keep the signature of the methods in DBFilter as unified as possible while keeping the DatabaseImplementation as agnostic as possible to the details of PostgreSQL implementation.

An alternative I considered was to override the whole `QueryResources` method for calling specific functions of PostgreSQLFilter but it seems cleaner this way.

Also, my first approach was to check the whereClause already introduced in the selector, and I created an [issue](https://github.com/huandu/go-sqlbuilder/issues/194) in the sqlbuilder project but it seems that's not recommended. Check it out for more context.

### DIFF
--- a/pkg/database/reader.go
+++ b/pkg/database/reader.go
@@ -40,13 +40,11 @@ func NewReader() (DBReader, error) {
 func (db *DatabaseImpl) QueryResources(ctx context.Context, kind, apiVersion, namespace, name,
 	continueId, continueDate string, labelFilters *LabelFilters, limit int) ([]string, int64, string, error) {
 	sb := db.selector.ResourceSelector()
-	sb.Where(
-		db.filter.KindApiVersionFilter(sb.Cond, kind, apiVersion),
-	)
-
+	sb.Where(db.filter.KindApiVersionFilter(sb.Cond, kind, apiVersion))
 	if namespace != "" {
 		sb.Where(db.filter.NamespaceFilter(sb.Cond, namespace))
 	}
+	mainWhereClause := sqlbuilder.CopyWhereClause(sb.WhereClause)
 	if name != "" {
 		sb.Where(db.filter.NameFilter(sb.Cond, name))
 	} else {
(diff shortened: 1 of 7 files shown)


pkg/database/reader.go
```diff
@@ -115,14 +115,8 @@ func (db *DatabaseImpl) QueryLogURL(ctx context.Context, kind, apiVersion, names
...
-	sb.Where(
-		db.filter.KindFilter(sb.Cond, kind),
-		db.filter.ApiVersionFilter(sb.Cond, apiVersion),
-	)
-
-	if namespace != "" {
-		sb.Where(db.filter.NamespaceFilter(sb.Cond, namespace))
-	}
+	mainWhereClause := db.mainWhereClause(kind, apiVersion, namespace)
+	sb.AddWhereClause(mainWhereClause)
```
rh-hemartin, line 119: I don't think you need this. After the two removed `sb.Where` you can pass `sb.WhereClause` to the `NotEqualsLabelFilter`, this way you avoid the extraction to a different function

rh-hemartin:
> The idea is that for the "complex" filters we allow to pass a WhereClause with the kind, apiversion and namespace (if applies) filters of the main query so it can be attached to a subquery if it is done.

If I understood correctly what a subquery is, I don't think its a good idea that the functions that are supposed to build filters do queries to the database themselves before returning the actual filter. If the filter requires some data it should expect it as a parameter and probably `QueryResources` should make the call, similar to what `QueryLog` does.

Remember that we started with this system in a scenario in which the queries were simple, now that grew in complexity we should change the system to accommodate the complexity. We all prefer simple things, but I think we can also agree that is best to see the complexity upfront rather than finding hidden complexity later.

If you want, ping me on Slack and we can explore your ideas on a call or something.

maruiz93:
> I don't think its a good idea that the functions that are supposed to build filters do queries to the database themselves before returning the actual filter.

There is no extra call being done, when I say "subquery" I mean "nested select", and for this particular case, the nested select reuse the same filters of `kind`, `api_version` and `namespace` but I wanted to avoid passing those values directly to the `NotEqualsLabelFilter` because that implies changing the signature of the method in the `DBFilter` interface implemented in the different "engines".

maruiz93:
Updated the description with the link to an issue related to this feature I opened on sqlbuilder project

pkg/database/reader.go
maruiz93: I've refactored this piece and removed the extra method. Please take a look and tell me if that's what you meant

rh-hemartin (approved):

skoved (approved):

PR #932, opened 2025-03-21
Apply singleton pattern to NewDatabase

### LINKED ISSUE #931: Apply singleton pattern to Database constructor
There should be only one instance of `Database` during runtime in KubeArchive components so we should implement its instantiation as a [Singleton](https://refactoring.guru/es/design-patterns/singleton)

### Acceptance Criteria

* No matter how many times the method to retrieve a Database is called, the same instance should be returned.

maruiz93:

Resolves #931

## Release Note

```release-note
NONE
```

### DIFF
--- a/pkg/database/database.go
+++ b/pkg/database/database.go
@@ -12,6 +12,7 @@ import (
 	"log/slog"
 	"slices"
 	"strings"
+	"sync"
 
 	"github.com/huandu/go-sqlbuilder"
 	"github.com/jmoiron/sqlx"
@@ -52,7 +53,18 @@ type Database struct {
 	Deleter  facade.DBDeleter
 }
 
+var db DBInterface
+var once sync.Once
+
 func NewDatabase() (DBInterface, error) {
+	var err error
+	once.Do(func() {
+		db, err = newDB()
+	})
+	return db, err
+}
+
+func newDB() (DBInterface, error) {
 	env, err := newDatabaseEnvironment()
 	if err != nil {
 		return nil, err


pkg/database/database.go
```diff
@@ -52,7 +53,18 @@ type Database struct {
 	Deleter  facade.DBDeleter
 }
 
+var db DBInterface
```
rh-hemartin, line 56: Should this be a pointer?

maruiz93: yes I think it's a good idea to also tackle this change in the PR

rh-hemartin: I'm not suggesting any radical change here. Since this is a singleton the thing getting passed around shouldn't be copied. I don't know what happens if you pass an interface around, does that automatically pass the pointer of the underlying object? Or we need to make this explictly a pointer. Also, I'm not sure if we can make a pointer out of a interface.

maruiz93: ah, right! I answered too quickly! The interfaces shouldn't be used as pointers (generally speaking) as they are already pointers to what's implemented so, using a pointer in here it's like returning a pointer to a pointer.

You can check how interfaces work [here](https://research.swtch.com/interfaces). Note specifically this sentence:

> Interface values are represented as a two-word pair giving a pointer to information about the type stored in the interface and a pointer to the associated data. Assigning b to an interface value of type Stringer sets both words of the interface value.

rh-hemartin: Then yes, it should be kept as it.

rh-hemartin (approved):

skoved (approved):

pkg/database/database.go
```diff
@@ -52,7 +53,18 @@ type Database struct {
...
+
 func NewDatabase() (DBInterface, error) {
+	var err error
+	once.Do(func() {
+		db, err = newDB()
+	})
+	return db, err
+}
+
+func newDB() (DBInterface, error) {
```
ggallen, line 67: Why not just embed the code for `newDB` in `newDatabase`?

maruiz93: There is no special reason, I changed it to embed this piece of code in `newDatabase`

ggallen (approved):

PR #935, opened 2025-03-21
Refactor DB package to segregate sql related code

### LINKED ISSUE #930: Refactor database package to make clear what is coupled to SQL database implementation and what not
We currently only support SQL database engines, but we have a nice generic `DBInterface` or `Database` after https://github.com/kubearchive/kubearchive/pull/921 to just say that whatever struct that implements its methods is suitable for being used by the API and sink of kubearchive.

This potentially opens the design to be able to implement whatever database in the strict sense, a fake database , a database that could be using file sotorage, SQL databases, Non-sql databases... with this design we are [Open to Extension bud Closed for Modification](https://en.wikipedia.org/wiki/SOLID#Open%E2%80%93closed_principle).

But then, we provide a "default" `DatabaseImpl`, that couldn't be used by anything but a SQL engine database implementation.

The idea of this issue is to *move* and *rename* things around so we separate what is coupled to SQL and what is not.

### Acceptance Criteria:

* Create a subpackage within `pkg/database` that stores anything related with sql:
    * the `facade` package
    * a `sqlDatabase` interface composed by `Database` that includes the `sql` related methods like `setSqlConn( conn *sqlx.DB)`
    * the unit tests based on `go-sqlmock`
    * the different implementations of `sqlDatabase` and the interfaces under `pkg/database/facade/sql`
* Keep the `pkg/database` package with interfaces, structs and vars agnostic to SQL databases

maruiz93:

Resolves #930 

## Release Note

```release-note
Modifies the interface that should be implemented for add a new database engine.
```

## Notes for Reviewers

While applying this change I've seen some problems with the dependencies among packages and the order of execution of `init()`.

To get this working apart from creating the `database/sql` package I did the following changes:
* Move the `ResourceNotFound` error to its own package so database.sql and cmd.api.routers can depend on it.
* Move the `LabelsFilter` to `models` pkg so database.sql and cmd.api.routers can depend on it.
* Statically fill the `RegisteredDatabases` instead of delegating on the go init function.
* Move the `Database` interface to its own package so every implementation and the constructor can depend on it without having a cyclic dependency.


skoved (approved):

pkg/database/db/database.go
```diff
@@ -0,0 +1,34 @@
+// Copyright KubeArchive Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package db
```
rh-hemartin, line 4: I don't think `db` is expressive enough, what about calling this `interfaces` or `types`?

maruiz93: I started calling the package `interface` but I didn't like it because even if it's doable, the word `interface` is reserved, so to import it you need always to get an alias. I guess that `interfaces` solves that problem and after reading about general naming conventions in go, the idea is to use singular names except to prevent name collisions (e.g. `strings`, `types`, `errors`...) so I'll go with `interfaces`.

maruiz93: Done!

maruiz93:
Updated the docs accordingly.

rh-hemartin (approved):

PR #937, opened 2025-03-21
Make resources not used outside its pkg unexported

### LINKED ISSUE #936: Don't expose unnecessary functions and structs
In the Database package we have a lot of resources related specifically to the implementation that are exposed but they shouldn't be used outside of its package (e.g. specific implementations of `DBInterface` and of the interfaces of `pkg/facade`).

### Acceptance criteria

* Unexport/Privatize the resources that aren't going to be used outside of its package.

maruiz93:

Resolves #936 

## Release Note

```release-note
NONE
```

### DIFF
--- a/pkg/database/fake/database_generated.go
+++ b/pkg/database/fake/database_generated.go
@@ -47,8 +47,8 @@ type LogUrlRow struct {
 	JsonPath      string
 }
 
-type Database struct {
-	database.Database
+type fakeDatabase struct {
+	database.DatabaseImpl
 	resources []*unstructured.Unstructured
 	logUrl    []LogUrlRow
 	jsonPath  string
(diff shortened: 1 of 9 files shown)


maruiz93:
> You have some side-effects from the find and replace, otherwise looks good to me

yes... I'll review them

maruiz93:
> > You have some side-effects from the find and replace, otherwise looks good to me
> 
> yes... I'll review them

reviewed and undo extra unintended changes on comments!

pkg/database/fake/database_generated.go
```diff
@@ -47,54 +47,54 @@ type LogUrlRow struct {
 	JsonPath      string
 }
 
-type Database struct {
+type fakeDatabase struct {
```
skoved, line 50: I don't know if the `fakeDatabase` type should be unexported. I know the sink uses this type for some it's unit tests

maruiz93: indeed, but I think there is no problem of having unexported if we have the "initializer" exported.

As I see it, all that needed to be exported related to database from API and sink POV are:
* The different constructors (`NewDatabase`, `NewFakeDatabase`)
* The interface `DBInterface` (or `DBReader` and `DBWriter`)

skoved (approved):

maruiz93:
After rebasing `DatabaseImpl` has to be exported as it is used outside of its package in `fake`.
If we end up going with https://github.com/kubearchive/kubearchive/pull/935 we can make it "private" again.

PR #938, opened 2025-03-21
Use sprintf + Var instead of sqlbuilder.Build

### LINKED ISSUE #926: Refactor the filter implementations to the recommended approach in sqlbuilder
sqlbuilder recommends doing `Sprintf` + `Var` instead of `Var` + `Build`.

E.g:

```go
query.Where(
		fmt.Sprintf(
			"NOT data->'metadata'->'labels' @> ANY($?::jsonb[])",
			innerQuery.Var(pq.Array(jsons),
		), 
	)
```

instead of

```go
query.Where(query.Var(
		sqlbuilder.Build(
			"NOT data->'metadata'->'labels' @> ANY($?::jsonb[])",
			pq.Array(jsons)
	))
```

### Acceptance criteria

* Revisit the filters implementations and change to the recommended approach whenever applies

maruiz93:

Resolves #926

## Release Note

```release-note
NONE
```

## Notes for Reviewers

With this change I believe it's more clear that we work with strings with regards to the filters, and that the sqlbuilder.Var, sqlbuilder.Cond are helpers to get those strings well formatted.

### DIFF
--- a/pkg/database/postgresql.go
+++ b/pkg/database/postgresql.go
@@ -66,39 +66,39 @@ type PostgreSQLFilter struct {
 }
 
 func (PostgreSQLFilter) CreationTSAndIDFilter(cond sqlbuilder.Cond, continueDate, continueId string) string {
-	return cond.Var(sqlbuilder.Build(
-		"(data->'metadata'->>'creationTimestamp', id) < ($?, $?)",
-		continueDate, continueId,
-	))
+	return fmt.Sprintf(
+		"(data->'metadata'->>'creationTimestamp', id) < (%s, %s)",
+		cond.Var(continueDate), cond.Var(continueId),
+	)
 }
 
 func (PostgreSQLFilter) OwnerFilter(cond sqlbuilder.Cond, owners []string) string {
-	return cond.Var(sqlbuilder.Build(
-		"jsonb_path_query_array(data->'metadata'->'ownerReferences', '$[*].uid') ?| $?",
-		pq.Array(owners),
-	))
+	return fmt.Sprintf(
+		"jsonb_path_query_array(data->'metadata'->'ownerReferences', '$[*].uid') ?| %s",
+		cond.Var(pq.Array(owners)),
+	)
 }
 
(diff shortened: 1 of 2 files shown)


rh-hemartin (approved):

ggallen (commented):
Doesn't this change leave us open to SQL injection attacks?

Probably not as KubeArchive is currently implemented, but it could in the future if we ever open things up.  But this is one danger with making this change.

Is this change worth it?

maruiz93:
> Doesn't this change leave us open to SQL injection attacks?
> 
> Probably not as KubeArchive is currently implemented, but it could in the future if we ever open things up. But this is one danger with making this change.
> 
> Is this change worth it?

No, it's not more open with SQL injection than it was before.
Note that we keep using the `sqlbuilder.Cond` and `sqlbuilder.Var` to create the different "string" pieces so we should be covered from SQL injection the same way we were before this PR (where we were still working with strings and the `sqlbuilder.Cond` + `sqlbuilder.Var` utilities, it's just a slight refactor to make things "clearer" as I see it and to adhere to the good practices stated by the `sqlbuilder` maintainer.

pkg/database/postgresql.go
```diff
@@ -66,50 +66,50 @@ type PostgreSQLFilter struct {
...
-func (PostgreSQLFilter) EqualsLabelFilter(cond sqlbuilder.Cond, labels map[string]string) string {
+func (PostgreSQLFilter) EqualsLabelFilter(_ sqlbuilder.Cond, labels map[string]string) string {
 	jsonLabels, _ := json.Marshal(labels)
-	return cond.Var(sqlbuilder.Build(
-		"data->'metadata'->'labels' @> $?",
+	return fmt.Sprintf(
+		"data->'metadata'->'labels' @> %s",
 		string(jsonLabels),
-	))
+	)
```
maruiz93, line 101: @ggallen this is the only change that is "open" to sql injection, I'm changing it so we use it for the jsonlabels var

maruiz93: fixed!

ggallen:
OK, I'm good.

ggallen (approved):

PR #939, opened 2025-03-21
Merge KindFilter and ApiVersionFilter

### LINKED ISSUE #925: Merge kind and apiversion filters
We currently have two separate methods for filtering `kind` and `api_version` in the `DBFilter` interface.

There shouldn't be an scenario where one is applied without the other, so we could shrink the interface merging both methods.

### Acceptance Criteria

* There is only one method in `DBFilter` for filtering by api_version and kind at the same time.

maruiz93:

Resolves #925 

## Release Note

```release-note
NONE
```

### DIFF
--- a/pkg/database/facade/filter.go
+++ b/pkg/database/facade/filter.go
@@ -8,8 +8,7 @@ import "github.com/huandu/go-sqlbuilder"
 // DBFilter encapsulates all the Filter functions that must be implemented by the drivers
 // All its functions share the same signature
 type DBFilter interface {
-	KindFilter(cond sqlbuilder.Cond, kind string) string
-	ApiVersionFilter(cond sqlbuilder.Cond, apiVersion string) string
+	KindApiVersionFilter(cond sqlbuilder.Cond, kind, apiVersion string) string
 	NamespaceFilter(cond sqlbuilder.Cond, ns string) string
 	NameFilter(cond sqlbuilder.Cond, name string) string
 	CreationTSAndIDFilter(cond sqlbuilder.Cond, continueDate, continueId string) string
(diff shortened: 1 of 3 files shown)


rh-hemartin (approved):

pkg/database/facade/filter.go
```diff
@@ -29,12 +28,8 @@ type DBFilter interface {
...
 type PartialDBFilterImpl struct{}
 
-func (PartialDBFilterImpl) KindFilter(cond sqlbuilder.Cond, kind string) string {
-	return cond.Equal("kind", kind)
-}
-
-func (PartialDBFilterImpl) ApiVersionFilter(cond sqlbuilder.Cond, apiVersion string) string {
-	return cond.Equal("api_version", apiVersion)
+func (PartialDBFilterImpl) KindApiVersionFilter(cond sqlbuilder.Cond, kind, apiVersion string) string {
+	return cond.And(cond.Equal("kind", kind), cond.Equal("apiVersion", apiVersion))
```
rh-hemartin, line 32: Just after checking the tests...

```suggestion
	return cond.And(cond.Equal("kind", kind), cond.Equal("api_version", apiVersion))
```

I'm wondering why this is not being raised on the unit tests... any idea?

maruiz93: Because the unit tests mock the database, so this is only raised when the actual table of the database don't have `apiversion` and have `api_version` instead as a column.

It's being raised in the integration test though.

BTW, now it's solved.

skoved (approved):

ggallen:
> There shouldn't be an scenario where one is applied without the other, so we could shrink the interface merging both methods.

What if I want all Tekton `PipelineRun`s of any API version?

maruiz93:
> > There shouldn't be an scenario where one is applied without the other, so we could shrink the interface merging both methods.
> 
> What if I want all Tekton `PipelineRun`s of any API version?

Well, that isn't allowed by the current API endpoints that we have, that are a replica of the ones used in Kubernetes API so I guess that we are not willing to offer that feature. Every retrieved resource by kubearchive API must be always filtered by `kind` + `apiVersion`

ggallen:
> > > There shouldn't be an scenario where one is applied without the other, so we could shrink the interface merging both methods.
> > 
> > 
> > What if I want all Tekton `PipelineRun`s of any API version?
> 
> Well, that isn't allowed by the current API endpoints that we have, that are a replica of the ones used in Kubernetes API so I guess that we are not willing to offer that feature. Every retrieved resource by kubearchive API must be always filtered by `kind` + `apiVersion`

OK, I was thinking of `kubectl` where I don't think you need to specify API version anywhere, you just `get` the kind.

ggallen (approved):

maruiz93:
> > > > There shouldn't be an scenario where one is applied without the other, so we could shrink the interface merging both methods.
> > > 
> > > 
> > > What if I want all Tekton `PipelineRun`s of any API version?
> > 
> > 
> > Well, that isn't allowed by the current API endpoints that we have, that are a replica of the ones used in Kubernetes API so I guess that we are not willing to offer that feature. Every retrieved resource by kubearchive API must be always filtered by `kind` + `apiVersion`
> 
> OK, I was thinking of `kubectl` where I don't think you need to specify API version anywhere, you just `get` the kind.

Right, but behind the scenes the cli does a call to the API with a specific version of the resource :)

PR #952, opened 2025-03-24
Only update fields in resources that the operator controls.

### LINKED ISSUE #940: kubearchive operator does not handle the kubearchive-a13e SA properly in Openshift
KubeArchive took down the konflux staging cluster. The reconcile loop for the product-kubearchive-a13e SA removes annotation applied by openshift secrets manager. Openshift secret manager creates a new secret (without removing the old one). There were > 250,000 secrets

[this line in the operator seems to be the culprit](https://github.com/kubearchive/kubearchive/blob/main/cmd/operator/internal/controller/kubearchiveconfig_controller.go#L218). I don’t really know what the intent is here, but we cannot just blindly wipe everything from the SA

**Additional Context:** 
`product-` is a prefix that konflux applies to the KubeArchive release manifest to get around the fact that Openshift does not let you create namespace that starts with the prefix `kube`

What I'm pretty sure happens is that the product-kubearchive-a13e SA is created, this service account is used by the ApiServerSource to generate a token that it can use to get, list, and watch resources listed in the `KubeArchiveConfig`. When the token is generated for the ServiceAccount, a controller for Openshift Secrets Manager creates a secret and applies an annotation to the service account. I'm not sure if this action directly causes the KubeArchive operator to reconcile the Service Account or if some other events happen eventually that cause the KubeArchive operator to reconcile this service account, but when that reconciliation happens, Openshift Secrets Manager has to generate a new secret for the token (because the annotation stating that a secret was created for it is gone) and the old secret is not removed because the service account no longer owns the old secret.

ggallen:

Resolves #940

## Release Note

```release-note
NONE
```


rh-hemartin (approved):

rh-hemartin:
Add the `kind/bug` label and add a release note that says something similar to:

```
Fixed a bug where the various resources were recreated removing any changes Kubernetes made, causing some problems.
```

cmd/operator/internal/controller/kubearchiveconfig_controller.go
```diff
@@ -479,29 +483,30 @@ func (r *KubeArchiveConfigReconciler) reconcileA13e(ctx context.Context, resourc
...
+			return nil, err
 		}
-	} else {
-		log.Error(err, "Failed to reconcile ApiServerSource")
-		return source, err
 	}
-
-	return source, nil
+	log.Error(err, "Failed to reconcile ApiServerSource")
+	return nil, err
```
maruiz93, line 509: Following the previous suggestions:

```
if errors.IsNotFound(err) {
	err = r.Create(ctx, desired)
	if err != nil {
		log.Error(err, "Failed to create ApiServerSource")
		return nil, err
	}
	return desired, nil
} else if err != nil {
	log.Error(err, "Failed to reconcile ApiServerSource")
	return nil, err
}

existing.Spec = desired.Spec
err = r.Update(ctx, existing)
if err != nil {
	log.Error(err, "Failed to update ApiServerSource")
	return nil, err
}
return existing, nil
```

cmd/operator/internal/controller/kubearchiveconfig_controller.go
```diff
@@ -591,37 +596,37 @@ func (r *KubeArchiveConfigReconciler) reconcileFilterConfigMap(ctx context.Conte
...
+			return nil, err
 		}
-	} else {
-		log.Error(err, "Failed to reconcile filter ConfigMap "+SinkFilterConfigMapName)
-		return cm, err
 	}
-
-	return cm, nil
+	log.Error(err, "Failed to reconcile filter ConfigMap "+SinkFilterConfigMapName)
+	return nil, err
```
maruiz93, line 629: Following the previous suggestions:

```
if errors.IsNotFound(err) {
	cm, err = r.desiredFilterConfigMap(ctx, kaconfig, nil)
	if err != nil {
		log.Error(err, "Unable to get desired filter ConfigMap "+SinkFilterConfigMapName)
		return nil, err
	}
	err = r.Create(ctx, cm)
	if err != nil {
		log.Error(err, "Failed to create filter ConfigMap "+SinkFilterConfigMapName)
		return nil, err
	}
	return cm, nil
} else if err != nil {
	log.Error(err, "Failed to reconcile filter ConfigMap "+SinkFilterConfigMapName)
	return nil, err
}

cm, err = r.desiredFilterConfigMap(ctx, kaconfig, cm)
if err != nil {
	log.Error(err, "Unable to get desired ConfigMap "+SinkFilterConfigMapName)
	return nil, err
}
err = r.Update(ctx, cm)
if err != nil {
	log.Error(err, "Failed to update filter ConfigMap "+SinkFilterConfigMapName)
	return nil, err
} 
return cm, nil
```

maruiz93 (commented):
A couple of comments to improve readibility

cmd/operator/internal/controller/kubearchiveconfig_controller.go
```diff
@@ -261,31 +255,34 @@ func (r *KubeArchiveConfigReconciler) reconcileRole(ctx context.Context, kaconfi
...
 		}
-	} else {
-		log.Error(err, "Failed to reconcile Role "+roleName)
-		return role, err
 	}
-
-	return role, nil
+	log.Error(err, "Failed to reconcile Role "+roleName)
+	return nil, err
 }
```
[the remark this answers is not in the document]

cmd/operator/internal/controller/kubearchiveconfig_controller.go
```diff
@@ -327,27 +324,30 @@ func (r *KubeArchiveConfigReconciler) reconcileClusterRole(ctx context.Context,
...
+			return nil, err
 		}
-	} else {
-		log.Error(err, "Failed to reconcile ClusterRole "+roleName)
-		return role, err
 	}
-
-	return role, nil
+	log.Error(err, "Failed to reconcile ClusterRole "+roleName)
+	return nil, err
```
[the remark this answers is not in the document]

cmd/operator/internal/controller/kubearchiveconfig_controller.go

cmd/operator/internal/controller/kubearchiveconfig_controller.go

maruiz93 (approved):

PR #974, opened 2025-03-28
renovate updates dependency versions in bash scripts

### LINKED ISSUE #973: renovate update
Renovate provides a [custom manager](https://docs.renovatebot.com/modules/manager/regex/) that use regex to match dependencies that need to be updated. If we follow the example, we could add the following to some of our bash scripts where we set the version of a dependency to a specific version so that renovate could bump the version when there are new releases

```
# renovate: datasource=<data source> depName=<name> packageName=<package name>
export DEP_NAME_VERSION=vX.Y.Z
```

skoved:

Resolves #973 

## Release Note

```release-note
NONE
```

## Notes for Reviewers

The goal is to get renovate to manage the versions of dependencies we install in our bash scripts. In doing this, I also realized that we document the version of Cert Manager and Knative Eventing in our documentation, so I extended this to also manage the version we list for those dependencies in our documentation as well

### DIFF
--- a/docs/modules/ROOT/pages/getting-started/installation.adoc
+++ b/docs/modules/ROOT/pages/getting-started/installation.adoc
@@ -12,7 +12,9 @@ This document explains how to install and configure KubeArchive in your Kubernet
 == Prerequisites
 
 * A PostgreSQL instance protected with TLS (can be self-signed, KubeArchive does not verify it)
+// renovate: datasource=github-releases depName=cert-manager packageName=cert-manager/cert-manager
 * CertManager is installed on the Kubernetes cluster (+v1.9.1)
+// renovate: datasource=github-releases depName=knative-eventing packageName=knative/eventing
 * Knative Eventing is installed on the Kubernetes cluster (+v1.15.0)
 
 == Prepare the Database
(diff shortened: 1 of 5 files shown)


maruiz93 (approved):

rh-hemartin (approved):

skoved:
On hold until discussion until discussion at the community meeting on thursday

mafh314 (approved):

maruiz93:
From [cert-manager docs](https://redhat-internal.slack.com/archives/CEXRYS5QC/p1743757874624099)

Related information on what they recommend regarding version updates:

> All cert-manager releases are supported at least until the release of a second subsequent version. That means there are always at least two supported versions of cert-manager at any given time, and possibly more if there's also a current Long Term Support version.
> We aim to do regular releases roughly every 4 months but release dates can vary when accounting for holidays, conferences (such as KubeCon), maintainer commitments and other world events.
> Only the last patch release of each branch is supported.

skoved:
> We need to make renovate open these PRs but they should be manually merged.

added some package rules. the first disables automerge for PR generated by the regex dependency manager. the second groups the knative-eventing dependencies in found by the regex matcher and groups them, so the documentation and bash script version stay in sync. the third does the same as the second but for cert-manager dependencies

rh-hemartin (approved):

PR #977, opened 2025-03-31
Add output format on the cli

### LINKED ISSUE #299: CLI allows for JSON or YAML return
It may be useful to have options to format the output of the kubearchive kubectl plugin as JSON or YAML for automated tool usage.

mafh314:

Resolves #299 

## Release Note

```release-note
NONE
```

### DIFF
--- a/pkg/cmd/get.go
+++ b/pkg/cmd/get.go
@@ -5,6 +5,7 @@ package cmd
 import (
 	"encoding/json"
 	"fmt"
+	"gopkg.in/yaml.v3"
 	"io"
 	"net/http"
 	"strings"
@@ -25,8 +26,10 @@ type GetOptions struct {
 
 	KubeArchiveHost string
 
-	RESTConfig *rest.Config
-	kubeFlags  *genericclioptions.ConfigFlags
+	RESTConfig    *rest.Config
+	kubeFlags     *genericclioptions.ConfigFlags
+	Output        string
+	IsValidOutput bool
 }
 
 func NewGetOptions() *GetOptions {
@@ -62,6 +65,7 @@ func NewGetCmd() *cobra.Command {
 	o.kubeFlags.AddFlags(cmd.Flags())
 	cmd.Flags().BoolVarP(&o.AllNamespaces, "all-namespaces", "A", o.AllNamespaces, "If present, list the requested object(s) across all namespaces. Namespace in current context is ignored even if specified with --namespace.")
 	cmd.Flags().StringVar(&o.KubeArchiveHost, "kubearchive-host", o.KubeArchiveHost, fmt.Sprintf("Host where the KubeArchive API Server is listening. Defaults to '%s'", o.KubeArchiveHost))
+	cmd.Flags().StringVarP(&o.Output, "output", "o", "", "Output format. One of: (json, yaml).")
 
 	return cmd
 }
@@ -90,6 +94,27 @@ func (o *GetOptions) Complete(args []string) error {
 
 	o.RESTConfig = config
 
+	if o.Output != "" {
+		err = o.Validate()
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (o *GetOptions) Validate() error {
+	validOutputFormats := []string{"json", "yaml"}
+	o.IsValidOutput = false
+	for _, validOutputFormat := range validOutputFormats {
+		if validOutputFormat == o.Output {
+			o.IsValidOutput = true
+		}
+	}
+	if !o.IsValidOutput {
+		return fmt.Errorf("unable to match a printer suitable for the output format %s, allowed formats are: json, yaml", o.Output)
+	}
 	return nil
 }
 
@@ -145,6 +170,38 @@ func (o *GetOptions) Run() error {
 	var objs []unstructured.Unstructured
 	objs = append(objs, clusterResources...)
 	objs = append(objs, kubearchiveResources...)
+
+	if o.Output != "" {
+		type List struct {
+			Items      []*unstructured.Unstructured `json:"items"`
+			ApiVersion string                       `json:"apiVersion" yaml:"apiVersion"`
+			Kind       string                       `json:"kind"`
+			Metadata   map[string]any               `json:"metadata"`
+		}
+		list := &List{
+			ApiVersion: "v1",
+			Kind:       "List",
+			Metadata:   map[string]any{},
+		}
+		for _, obj := range objs {
+			list.Items = append(list.Items, &obj)
+		}
+		if o.Output == "json" {
+			listData, err := json.MarshalIndent(list, "", "    ")
+			if err != nil {
+				return err
+			}
+			fmt.Println(string(listData))
+			return nil
+		} else if o.Output == "yaml" {
+			listYaml, err := yaml.Marshal(list)
+			if err != nil {
+				return err
+			}
+			fmt.Println(string(listYaml))
+			return nil
+		}
+	}
 
 	for ix := range objs {
 		fmt.Println(objs[ix].GetName())
--- a/pkg/cmd/get_test.go
+++ b/pkg/cmd/get_test.go
@@ -15,6 +15,7 @@ func TestCompleteAPI(t *testing.T) {
 		args            []string
 		expectedApiPath string
 		isCore          bool
+		output          string
 	}{
 		{
 			name:            "core",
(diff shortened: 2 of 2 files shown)


pkg/cmd/get.go
```diff
@@ -62,6 +65,7 @@ func NewGetCmd() *cobra.Command {
 	o.kubeFlags.AddFlags(cmd.Flags())
 	cmd.Flags().BoolVarP(&o.AllNamespaces, "all-namespaces", "A", o.AllNamespaces, "If present, list the requested object(s) across all namespaces. Namespace in current context is ignored even if specified with --namespace.")
 	cmd.Flags().StringVar(&o.KubeArchiveHost, "kubearchive-host", o.KubeArchiveHost, fmt.Sprintf("Host where the KubeArchive API Server is listening. Defaults to '%s'", o.KubeArchiveHost))
+	cmd.Flags().StringVarP(&o.Output, "output", "o", "", "Output format. One of: (json, yaml).")
```
rh-hemartin, line 68: I'm wondering if the default should have a name, feels weird to have "defaults to an empty string" or "do not pass it to print the resources as a table". What do you think about having three: table, json or yaml, being "table" the deafult.

pkg/cmd/get.go
```diff
@@ -90,6 +94,27 @@ func (o *GetOptions) Complete(args []string) error {
...
+func (o *GetOptions) Validate() error {
+	validOutputFormats := []string{"json", "yaml"}
+	o.IsValidOutput = false
+	for _, validOutputFormat := range validOutputFormats {
+		if validOutputFormat == o.Output {
+			o.IsValidOutput = true
+		}
+	}
+	if !o.IsValidOutput {
+		return fmt.Errorf("unable to match a printer suitable for the output format %s, allowed formats are: json, yaml", o.Output)
```
rh-hemartin, line 116:
```suggestion
		return fmt.Errorf("unable to match a printer suitable for the output format %s, allowed formats are: %s", o.Output, strings.Join(validOutputFormats, ", "))
```

pkg/cmd/get.go
```diff
@@ -146,6 +171,38 @@ func (o *GetOptions) Run() error {
...
+			Metadata   map[string]any               `json:"metadata"`
+		}
+		list := &List{
+			ApiVersion: "v1",
+			Kind:       "List",
+			Metadata:   map[string]any{},
+		}
+		for _, obj := range objs {
+			list.Items = append(list.Items, &obj)
+		}
```
rh-hemartin, line 188: Maybe this works the same way:

```suggestion
		list := &List{
			ApiVersion: "v1",
			Kind:       "List",
			Metadata:   map[string]any{},
			List:              objs,
		}
```

pkg/cmd/get.go
ggallen: You shouldn't be able to specify `-o` by itself.

> "do not pass it to print the resources as a table" 

This is the way `kubectl` works.  Don't pass `-o` and you get the table.

pkg/cmd/get.go
```diff
@@ -90,6 +94,27 @@ func (o *GetOptions) Complete(args []string) error {
...
+	validOutputFormats := []string{"json", "yaml"}
+	o.IsValidOutput = false
+	for _, validOutputFormat := range validOutputFormats {
+		if validOutputFormat == o.Output {
+			o.IsValidOutput = true
+		}
+	}
+	if !o.IsValidOutput {
+		return fmt.Errorf("unable to match a printer suitable for the output format %s, allowed formats are: json, yaml", o.Output)
+	}
```
skoved, line 117: I think would be a good to move the definition of `validOutputFormats` outside of the `Validate` function to make it easier to find them. That would also let define them as enums which would let us use a switch case here and later on in get resources, which i think helps make the logic easier to follow.

```suggestion

type outputFormat string

func (of outputFormat) String() string {
    return string(of)
}

func (of outputFormat) Is(str string) bool {
    return string(of) == str
}

const (
    outputJson outputFormat = "json"
    outputYaml outputFormat = "yaml"
    outputTable outputFormat = "" // since table is the default
)

func (o *GetOptions) Validate() error {
	switch o.Output {
	case outputJson.String(), outputYaml.String(), outputTable.String():
	    return nil
	default:
	    return fmt.Errorf("unable to match a printer suitable for the output format %s, allowed formats are: json, yaml", o.Output)
	}
```

pkg/cmd/get.go
```diff
@@ -146,6 +171,38 @@ func (o *GetOptions) Run() error {
...
+			return nil
+		} else if o.Output == "yaml" {
+			listYaml, err := yaml.Marshal(list)
+			if err != nil {
+				return err
+			}
+			fmt.Println(string(listYaml))
+			return nil
+		}
+	}
```
skoved, line 204: If you have enums for the output types, this whole if clause can be replaced with a switch statement. It would just require this to be defined first

```
type List struct {
    Items      []*unstructured.Unstructured `json:"items"`
    ApiVersion string `json:"apiVersion" yaml:"apiVersion"`
    Kind       string `json:"kind"`
    Metadata   map[string]any `json:"metadata"`
}

func NewList(objs []unstructured.Unstructured) *List {
    list := &List{
        ApiVersion: "v1",
        Kind: "List"
        Metadata: maps[string]any{},
    }
    for _, obj := range objs {
        list.Items = append(list.Items, &obj)
    }
    return list
}
```

```suggestion
		switch o.Output {
		case formatJson.String():
		    list = NewList(objs)
		    listData, err := json.MarshalIndent(list, "", "    ")
			if err != nil {
				return err
			}
			fmt.Println(string(listData))
			return nil
	    case formatYaml.String():
	        listYaml, err := yaml.Marshal(list)
			if err != nil {
				return err
			}
			fmt.Println(string(listYaml))
			return nil
		}
	}
```

skoved (changes requested):
Thanks for working on this. I think using enums for the output formats might make the logic easier to follow but otherwise this looks good!

pkg/cmd/get.go
maruiz93: For reference, this is the piece of code in the `kubectl get` module that includes the output option: https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/get/get_flags.go#L155

The option uses this `JSONYamlPrintFlags` defined [here](https://github.com/kubernetes/cli-runtime/blob/master/pkg/genericclioptions/json_yaml_flags.go) in which the options are just `json` and `yaml` (no table) so, I agree with Greg that we should stick with the same design.

Actually, I'm wondering if we could use exactly the same flags from the library `k8s.io/cli-runtime/pkg/genericclioptions` what do you think?

pkg/cmd/get.go
maruiz93: If we use the flags already defined in k8s.io/cli-runtime/pkg/genericclioptions the same way that `kubectl get` does ([here](https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/get/get_flags.go#L155)) we can just remove this part as it's already defined there.

pkg/cmd/get.go
maruiz93: I think that @rh-hemartin's suggestion is a good one, but let me propose an extra change.
If the struct is going to be used only inside the `if` (where it is defined) there is also the alternative to not declare the struct (which make sense if it isn't going to be referenced):

```go
	if o.Output != "" {
        list := struct {
            Items      []*unstructured.Unstructured `json:"items"`
			ApiVersion string                       `json:"apiVersion" yaml:"apiVersion"`
			Kind       string                       `json:"kind"`
			Metadata   map[string]any               `json:"metadata"`
        }{
            ApiVersion: "v1",
			Kind:       "List",
			Metadata:   map[string]any{},
			List:       objs,
        }
```

pkg/cmd/get.go
mafh314: that makes sense since I defined validOutputFormats :) Thanks!

mafh314:
> @mafh314 could you work on this?

yes, I just need to test the changes

pkg/cmd/get.go
mafh314: Changed

rh-hemartin (approved):

maruiz93 (changes requested):
Can you please take a look at https://github.com/kubearchive/kubearchive/pull/1178

It's going with another  approach, by using the printers used by kubectl instead of creating them manually.

This turned out to be really useful specially considering the table printer.

Issue #986, opened 2025-04-02
Investigate memory consumption with +500 KubeArchiveConfigs


rh-hemartin:
Currently we are seeing a large amount of memory being consumed by the sink when there are a lot of KubeArchiveConfigs. In the installation we have there are 531 entries in the `sink-filters` ConfigMap and there are global filters. We know for a fact that 530 of those 531 KubeArchive configs are empty, the only populated is the global one.

To reproduce a similar problem locally run:

```
kind create cluster
export KO_DOCKER_REPO="kind.local"
bash hack/quick-install.sh
bash integrations/observability/grafana/install.sh
```

Then apply a global config:

```
---
apiVersion: kubearchive.kubearchive.org/v1alpha1
kind: KubeArchiveConfig
metadata:
  name: kubearchive
spec:
  resources:
    - selector:
        apiVersion: batch/v1
        kind: CronJob
      archiveWhen: "true"
    - selector:
        apiVersion: batch/v1
        kind: Job
      archiveWhen: has(status.startTime)
      deleteWhen: has(status.completionTime)
    - selector:
        apiVersion: v1
        kind: Pod
      archiveOnDelete: "true"
```

And then:

```
for i in {1..500}; do kubectl create ns test-$i; kubectl apply -f kac.yaml -n test-$i; done
```

Where kac.yaml is:

```
---
apiVersion: kubearchive.kubearchive.org/v1alpha1
kind: KubeArchiveConfig
metadata:
  name: kubearchive
spec:
  resources: []
```

You can retrieve heap dumps to inspect resource usage as well:

```
kubectl port-forward -n kubearchive svc/kubearchive-sink 8080:80  &
go tool pprof -http :8081 -no_browser http://localhost:8080/debug/pprof/heap
# go to localhost:8081
```

skoved:
that makes sense with what I was seeing in pprof as well. The majority (~200MiB) are objects related to the compiled CEL expressions that the sink has to create


what i can see from pprof is that ~200MiB of memory being used is for objects created by the cel library when we compiled the cel expressions for all of the namespaces

that would mean ~386KiB per namespace

right now the kubearchive config only has 5 resources and each resource only has an archiveondelete expression. that means we're using ~77KiB per CEL expression

skoved:
something I'm thinking about regarding memory usage in the sink:
should we be compiling CEL expressions as ([local expr] || [global expr])?
Alternative:
[...]

PR #988, opened 2025-04-02
Create CEL environment one time.

ggallen:

Related to #986 

## Release Note

```release-note
NONE
```

## Notes for Reviewers

This change for filters has memory going from 133.79MB to 1024.33KB.

Overall memory from 142.23MB down to 7.65MB.

These values based on the test case provided in #986.

### DIFF
--- a/pkg/cel/cel.go
+++ b/pkg/cel/cel.go
@@ -17,17 +17,12 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
 
-// CompileOrCELExpression attempts to compile cel expression strings into a cel.Program. If exprs contains more than one cel
-// expression string, it will surround each cell expression in parentheses and or them together.
-func CompileOrCELExpression(exprs ...string) (*cel.Program, error) {
-	exprs = FormatCelExprs(exprs...)
-	expr := strings.Join(exprs, " || ")
-	return CompileCELExpr(expr)
-}
+var mapStrDyn = types.NewMapType(types.StringType, types.DynType)
+var env *cel.Env
 
-func CompileCELExpr(expr string) (*cel.Program, error) {
-	mapStrDyn := types.NewMapType(types.StringType, types.DynType)
-	env, err := cel.NewEnv(
+func init() {
+	var err error
+	env, err = cel.NewEnv(
 		celext.Strings(),
 		celext.Encoders(),
 		celext.Sets(),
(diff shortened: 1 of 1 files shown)


pkg/cel/cel.go
```diff
@@ -17,17 +17,12 @@ import (
...
-}
+var mapStrDyn = types.NewMapType(types.StringType, types.DynType)
+var env *cel.Env
 
-func CompileCELExpr(expr string) (*cel.Program, error) {
-	mapStrDyn := types.NewMapType(types.StringType, types.DynType)
-	env, err := cel.NewEnv(
+func init() {
+	var err error
+	env, err = cel.NewEnv(
```
skoved, line 25: I think having `mapStryDyn` and `env` as gobal vars is fine, but I don't think that we should initialize them in an `init` function because of the unrecoverable error.

instead I think you should use [sync.Once](https://pkg.go.dev/sync#Once).

you would create Once struct as a global var and then in `CompileCELExpr` you could have something like what is done in [newDatabase](https://github.com/kubearchive/kubearchive/blob/main/pkg/database/database.go#L48)

skoved (changes requested):

pkg/cel/cel.go
ggallen: > but I don't think that we should initialize them in an `init` function because of the unrecoverable error.

But it is an unrecoverable error, which is why there is a `panic`.  What is the point of returning an error and continuing?

ggallen: Yeah, as I look at documentation and comments for `init` vs. `sync.Once`, I disagree.

https://stackoverflow.com/questions/51953191/difference-between-init-and-sync-once-in-golang

skoved (approved):

rh-hemartin (approved):

rh-hemartin:
While this make sense I'm not able to reproduce the results. Where did you get those memory readings from? I'm getting the values from the Grafana dashboard, using the `go_memory_used_bytes`. I see a spike at 250Mi.

ggallen:
> While this make sense I'm not able to reproduce the results. Where did you get those memory readings from? I'm getting the values from the Grafana dashboard, using the `go_memory_used_bytes`. I see a spike at 250Mi.

@rh-hemartin, I was looking at the "in use space" graph.  I've attached a screenshot of the results after my change.

sink-4

rh-hemartin:
> > While this make sense I'm not able to reproduce the results. Where did you get those memory readings from? I'm getting the values from the Grafana dashboard, using the `go_memory_used_bytes`. I see a spike at 250Mi.
> 
> @rh-hemartin, I was looking at the "in use space" graph. I've attached a screenshot of the results after my change.
> 

I can't reproduce yet, I'll ping you.

maruiz93 (approved):

rh-hemartin:
We found that there are no improvements on memory consumption, however the change still makes sense.

rh-hemartin (approved):

PR #1003, opened 2025-04-09
Introduce custom metrics

### LINKED ISSUE #985: Introduce KubeArchive custom metrics
Currently we only offer metrics that are related with the instrumentation we added: HTTP, DB, process and runtime. We want to offer custom metrics that make sense for KubeArchive as well. The proposed metrics (in SQL) are:

```
=> select kind as "Kind", count(kind) as "Number Archived" from resource group by kind order by kind;
    Kind     | Number Archived 
-------------+-----------------
 PipelineRun |             137
 Pod         |            1073
 Release     |              32
 Snapshot    |             500
 TaskRun     |            1071
(5 rows)

=> select date(created_at) as "Date", count(created_at) as "Number Archived" from resource group by date(created_at) order by date(created_at);
    Date    | Number Archived 
------------+-----------------
 2025-03-28 |             159
 2025-03-29 |              35
 2025-03-30 |              19
 2025-03-31 |             112
 2025-04-01 |             806
 2025-04-02 |            1682
(6 rows)


=> select date(created_at) as "Date", kind as "Kind", count(created_at) as "Number Archived" from resource group by date(created_at), kind order by date(created_at);
    Date    |    Kind     | Number Archived 
------------+-------------+-----------------
 2025-03-28 | Snapshot    |             159
 2025-03-29 | Snapshot    |              35
 2025-03-30 | Snapshot    |              19
 2025-03-31 | Snapshot    |             112
 2025-04-01 | PipelineRun |              32
 2025-04-01 | Pod         |             303
 2025-04-01 | Release     |               8
 2025-04-01 | Snapshot    |             160
 2025-04-01 | TaskRun     |             303
 2025-04-02 | PipelineRun |             105
 2025-04-02 | Pod         |             770
 2025-04-02 | Release     |              24
 2025-04-02 | Snapshot    |              15
 2025-04-02 | TaskRun     |             768
(14 rows)
```

I'm not sure about the implementation, names for these metrics, etc. However one thing to take into account is that pulling these numbers from the database provides us with the real state of the database, while relying on metric counters provides us with the state of the metrics database. This means that if the metrics database misses points the numbers may will be slightly off.

So let's start the discussion

rh-hemartin:

Resolves #985  

## Release Note

```release-note
Added `kubearchive.cloudevents.total` OpenTelemetry metric. See the [observability documentation(https://kubearchive.github.io/kubearchive/main/configuration/observability.html) for more information.
```

## Notes for Reviewers

* IMPORTANT: with this implementation, the counter resets on sink restart. This does not impact to `increase` as it account for resets, however it impacts if we want to see the total amount of stuff.
* Added installation of Tekton components into `test/log-generator/pipelines/install.sh`
* Fixed URL present on `config-tracing` so APIServerSource can send traces to the observability stack.
* Moved common `writeResource` implementation to each separate implementation, as they are starting to differ.
* Removed examples of patches for observability enablement, to shorten the document.
* Introduced documentation for the metric.
* Separated tests for `writeResources` as the implementations now differ

To test:

```
bash hack/quick-install.sh
bash integrations/observability/grafana/install.sh
bash test/log-generators/cronjob/install.sh --num-jobs=5
bash test/log-generators/pipeline/install.sh --num-jobs=5

kubectl port-forward -n observability svc/grafana-lgtm 9090
```


pkg/database/sql/mariadb.go
```diff
@@ -132,6 +135,45 @@ type mariaDBDatabase struct {
...
+	}
+	execErr = tx.Commit()
+	if execErr != nil {
+		rollbackErr := tx.Rollback()
+		if rollbackErr != nil {
+			return false, fmt.Errorf("commit to database failed: %s and unable to roll back transaction: %s", execErr, rollbackErr)
+		}
+		return false, fmt.Errorf("commit to database failed and the transactions was rolled back: %s", execErr)
+	}
+	return true, nil
```
maruiz93, line 174: With this new approach I would add a //FIXME comment with an issue ID in a comment to address retrieving the real information of if it was updated or inserted.

maruiz93 (changes requested):

docs/modules/ROOT/pages/configuration/observability.adoc
```diff
@@ -0,0 +1,43 @@
+= Observability on KubeArchive
+
+KubeArchive is instrumented to emit observability data: metrics,
```
maruiz93, line 3: Passive tense:

```suggestion
OpenTelemetry instruments KubeArchive to emit observability data: metrics,
```

docs/modules/ROOT/pages/configuration/observability.adoc
```diff
@@ -0,0 +1,43 @@
...
+traces and logs. This can be enabled and configured with environment
+variables present on the KubeArchive's components.
```
maruiz93, line 5: Passive tense:

```suggestion
traces and logs. Enable and configure the instrumentation setting the environment
variables present on the KubeArchive's components.
```

maruiz93 (changes requested):

pkg/database/sql/postgresql.go
```diff
@@ -165,13 +169,57 @@ func (postgreSQLInserter) ResourceInserter(
...
+		models.OptionalTimestamp(k8sObj.GetDeletionTimestamp()),
+		data,
+	).BuildWithFlavor(db.flavor)
+
+	inserted := false
+	execErr := tx.QueryRowContext(
+		ctx,
+		query,
+		args...,
+	).Scan(&inserted)
```
maruiz93, line 202: I see no reason to do a query in a different way that we do in the reader implementations, using the `performSingleRowQuery`

I suggest going with that:

```suggestion
	tx, err := db.db.BeginTxx(ctx, nil)
	if err != nil {
		return false, fmt.Errorf("could not begin transaction for resource %s: %s", k8sObj.GetUID(), err)
	}
        inserter := db.inserter.ResourceInserter(
		string(k8sObj.GetUID()),
		k8sObj.GetAPIVersion(),
		k8sObj.GetKind(),
		k8sObj.GetName(),
		k8sObj.GetNamespace(),
		k8sObj.GetResourceVersion(),
		lastUpdated,
		models.OptionalTimestamp(k8sObj.GetDeletionTimestamp()),
		data,
	)
	boolQueryPerformer := newQueryPerformer[bool](tx, db.flavor)
	inserted, execErr := boolQueryPerformer.performSingleRowQuery(ctx, inserter)
```

As the performer doesn't allow transaction, some small adjustments need to be made along this change:

```
type queryPerformer[T any] struct {
	querier     QueryerContext
	flavor sqlbuilder.Flavor
}

func newQueryPerformer[T any](querier sqlx.QueryerContext, flavor sqlbuilder.Flavor) queryPerformer[T] {
	return queryPerformer[T]{db, flavor, tx}
}


func (q queryPerformer[T]) performSingleRowQuery(ctx context.Context, builder sqlbuilder.Builder) (T, error) {
	var t T
	query, args := builder.BuildWithFlavor(q.flavor)
	err := sqlx.GetContext(ctx, q.querier, &t, query, args...)
	return t, err
}

func (q queryPerformer[T]) performQuery(ctx context.Context, builder sqlbuilder.Builder) ([]T, error) {
	var res []T
	query, args := builder.BuildWithFlavor(q.flavor)
	err := sqlx.SelectContext(ctx, q.querier, &res, query, args...)
	return res, err
}

```

maruiz93 (changes requested):

pkg/database/sql/postgresql.go
maruiz93: Note that the only change in the `query.go` file is changing `db *sqlx.DB` for `querier QueryerContext` and then doing `sqlx.GetContext` and `sqlx.SelectContext` instead of the specific methods from the current `*sqlx.DB`

docs/modules/ROOT/pages/configuration/observability.adoc
rh-hemartin: Thanks for keeping an eye on this. Fixed.

cmd/sink/routers/routers.go
```diff
@@ -137,6 +140,13 @@ func (c *Controller) writeResource(ctx context.Context, obj *unstructured.Unstru
 		"namespace", obj.GetNamespace(),
 		"name", obj.GetName(),
 	)
+
+	if inserted {
+		observability.ArchivedResources.Add(
+			ctx, 1, metric.WithAttributes(attribute.String("kind", obj.GetKind())),
+		)
+	}
```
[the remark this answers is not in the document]
rh-hemartin: Hmmm how do you envision we use this information? Now we differentiate between update/nothing and insert. So if we currently add something here we are bloating the metric, because sometimes the query runs but it does nothing.

pkg/database/sql/mariadb.go
rh-hemartin: Right, added.

pkg/database/sql/postgresql.go
rh-hemartin: Got it, I implemented the changes.

pkg/database/sql/writer.go
```diff
@@ -7,52 +7,12 @@ import (
...
 	"errors"
 	"fmt"
-	"time"
 
 	"github.com/kubearchive/kubearchive/pkg/database/sql/facade"
 	"github.com/kubearchive/kubearchive/pkg/models"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
 
-func (db *sqlDatabaseImpl) WriteResource(ctx context.Context, k8sObj *unstructured.Unstructured, data []byte, lastUpdated time.Time) error {
```
maruiz93, line 17: As we are deleting this method, then the `sqlDatabaseImpl` partially implements `Database` interface, so we should update our [docs](https://kubearchive.github.io/kubearchive/main/integrations/database.html#_adding_a_new_database_engine) explicitly saying that `sqlDatabaseImpl` is a partial implementation of `Database` so the developer should make sure that the new engine implementation fully implements all the methods of the `Database` interface..

cmd/sink/routers/routers.go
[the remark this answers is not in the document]
maruiz93: I envision having metrics that may answer the question:

* How many times an update cloud event end up in an insert for a resource of a specific apiversion/kind? --> Already covered, but I would suggest adding the `apiversion` because although it's unusual a cluster could share a  `Kind` name within different groups.
* When are we inserting resources of a specific kind/apiversion? --> Already covered with the previous one
* How many times a resource of a specific kind and id has been updated?
* How many times an update cloud event end up in an update for a resource of a specific kind?
* How many times an update cloud event ends up being discarded? 

I think that the three latest ones could be covered with two more metrics within the sink, one for `UpdatedResources` and another for `DiscardedCloudEvents` which could be useful to know that the current KAC maybe its too broad (when UpdatedResources are being done too often) or maybe even to consider including filters in the APIServerSource to prevent some events to reach the sink (this is something we briefly talked about in the last community meeting).

BTW, something that I think we also should consider is creating traces with this information:
* A resource with [kind] [apiversion] [name] [uid] has been created
* The resource with [kind] [apiversion] [name] [uid] has been updated
* The resource with [kind] [apiversion] [name] [uid] received in the cloud event doesn't comply with the configured CEL expressions.

Note that the traces can also be used to create metrics when aggregating them.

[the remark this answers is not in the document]
ggallen: I believe this metric is currently useless as the SQL can actually have three different outcomes:

1. A record was inserted.
2. A record was updated.
3. Nothing was done.

However, I believe there are way, at least in Postgres, to find out what really happened.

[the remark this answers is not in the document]
rh-hemartin: I moved the discussion back to the issue because we are going back to defining the metrics.

pkg/database/sql/writer.go
rh-hemartin: Added

pkg/database/sql/mariadb.go
maruiz93: ah, right, I didn't remember that we had the "WHERE resource.cluster_updated_ts < $?" logic in place, so yes, we currently aren't differentiating between record updated and "record not updated because it came in an ordered different than the expected"

BUT, I think that regarding the metrics we could still consider the "false+noerror" as an "update" because it's an attempt to update but it's not due to a race condition, so, if we want to know how many updates we have per kind in a timeframe compared with the inserts, we could count it as an update. Also, even though the record wasn't changed, the database connection was done, so also still useful if we want to aggregate inserts+updates to know about the amount of database interactions with regards to the sink

maruiz93: ah, just read your answer [here](https://github.com/kubearchive/kubearchive/issues/985#issuecomment-2792011237)

so could we use a type different from bool to have 3 possible values and be aware of when the record is neither updated or inserted?

rh-hemartin: I introduced an enum

rh-hemartin:
@skoved @ggallen @maruiz93 I went with a different metric and introduced more changes, please review the whole thing again.

pkg/observability/metrics.go
```diff
@@ -0,0 +1,48 @@
...
+		return CEResultInsert
+	} else if result == interfaces.WriteResourceResultUpdated {
+		return CEResultUpdate
+	} else if result == interfaces.WriteResourceResultNone {
+		return CEResultNone
+	} else if result == interfaces.WriteResourceResultError {
+		return CEResultError
+	}
+
+	return CEResultError
```
maruiz93, line 35: I think a switch fits better here:

```suggestion
    switch result {
        case interfaces.WriteResourceResultInserted:
            return CEResultInsert
        case interfaces.WriteResourceResultUpdated:
            return CEResultUpdate
        case interfaces.WriteResourceResultNone:
            return CEResultNone
        case interfaces.WriteResourceResultError:
            return CEResultError
        default:
            return CEResultError
    }
```

maruiz93 (approved):
LGTM

pkg/observability/metrics.go
rh-hemartin: Changed to switch

skoved (approved):

PR #1007, opened 2025-04-10
Add cluster resource ClusterKubeArchiveConfig.

### LINKED ISSUE #998: Create cluster resource ClusterKubeArchiveConfig
There are situations in both the CLI (#790) and the vacuums (#991) where the global `KubeArchiveConfig`  kept in the `kubearchive` namespace is needed to be seen by others.  Since it's namespaced, there are all kinds of issues around giving the correct permissions.

And in general, users may want to see the global configuration when they are configuring local `KubeArchiveConfig` resources.

The logical solution is to create a `ClusterKubeArchiveConfig`.

## Acceptance Criteria

- There is only one ClusterKubeArchiveConfig resource allowed.
- The name of the `ClusterKubeArchiveConfig` will be `kubearchive`
- The operator should reconcile `ClusterKubeArchiveConfig` resources.
- The operator will add the `ClusterKubeArchiveConfig`  resource to the sink-filters `ConfigMap`.
- The `kubearchive` namespace will behave as any other namespace, meaning it may also have a local `KubeArchiveConfig` defined.

ggallen:

Resolves #998 

## Release Note

```release-note
Global filters are now defined in a ClusterKubeArchiveConfig named "kubearchive".
```


rh-hemartin:
I removed the `kind/breaking` label because it will cause us to release v1, which I don't think we want do now. https://kubearchive.github.io/kubearchive/main/contributors/release.html#_pull_request_labels

rh-hemartin (approved):

skoved:
I just merged #990 which made some changes to `cmd/sink/filters/filters.go` which conflict with this change

cmd/operator/api/v1alpha1/kubearchiveconfig_types.go
```diff
@@ -33,33 +35,52 @@ type KubeArchiveConfigStatus struct {
...
 	if err != nil {
 		return []KubeArchiveConfigResource{}, err
 	}
-	return LoadFromString(string(yamlBytes))
+	return LoadKubeArchiveConfigFromString(string(yamlBytes))
 }
 
-func LoadFromString(yamlString string) ([]KubeArchiveConfigResource, error) {
+func LoadKubeArchiveConfigFromString(yamlString string) ([]KubeArchiveConfigResource, error) {
+	// Go through the hoops od unmarshalling to JSON so that fields are not lost. Not
```
skoved, line 63:
```suggestion
	// Go through the hoops of unmarshalling to JSON so that fields are not lost. Not
```

cmd/sink/filters/filters.go
```diff
@@ -26,17 +26,17 @@ import (
...
 var (
-	ErrNoGlobal = errors.New("no global expressions exist")
-	globalKey   string // gets set in init() and should be treated as const
+	ErrNoGlobal          = errors.New("no global expressions exist")
+	kubearchiveNamespace string // gets set in init() and should be treated as const
 )
 
 func init() {
-	globalKey = os.Getenv(globalKeyEnvVar)
+	kubearchiveNamespace = os.Getenv("KUBEARCHIVE_NAMESPACE")
```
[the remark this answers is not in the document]
ggallen: What value does it serve as a `const`?  It's only use was in the call to `os.Getenv`.

maruiz93 (approved):

cmd/sink/filters/filters.go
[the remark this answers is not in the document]
ggallen: I'm putting the `const` back.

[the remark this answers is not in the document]
ggallen: Fixed.

cmd/operator/internal/controller/clusterkubearchiveconfig_controller.go
```diff
@@ -0,0 +1,128 @@
...
+		}
+	} else {
+		// The object is being deleted.
+		if controllerutil.ContainsFinalizer(ckaconfig, finalizerName) {
+			// Finalizer is present, clean up filters from ConfigMap and remove Namespace label.
+
+			log.Info("Deleting ClusterKubeArchiveConfig")
+
+			// Clean filters.
+			err := deleteNamespaceFromFilterConfigMap(ctx, r.Client, filters.GlobalNamespace)
```
skoved, line 71: Why does this code import filters from the sink only to use `GlobalNamespace` when this variable is already defined in the controller package in `cmd/operator/internal/controller/common.go`?

cmd/operator/internal/controller/clusterkubearchiveconfig_controller.go
```diff
@@ -0,0 +1,128 @@
...
+		return ctrl.Result{}, nil
+	}
+
+	yaml, err := convertToYamlString(ckaconfig.Spec.Resources)
+	if err != nil {
+		log.Error(err, "Failed to convert ClusterKubeArchiveConfig resources to YAML")
+		return ctrl.Result{}, err
+	}
+
+	cm, err := reconcileFilterConfigMap(ctx, r.Client, filters.GlobalNamespace, yaml)
```
skoved, line 93: same as above

cmd/sink/filters/filters.go
```diff
@@ -25,14 +25,17 @@ import (
 )
 
 const (
-	globalKeyEnvVar = "KUBEARCHIVE_NAMESPACE"
 	filtersCmName   = "sink-filters"
+	GlobalNamespace = "___global___"
```
skoved, line 29: I think it might be better to import this constant from the operator so that we don't end up in a situation where the values in the sink and the operator get out of sync

cmd/operator/internal/controller/common.go
```diff
@@ -0,0 +1,346 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	kubearchivev1alpha1 "github.com/kubearchive/kubearchive/cmd/operator/api/v1alpha1"
+)
+
+const (
+	SinkFilterConfigMapName   = "sink-filters"
+	ApiServerSourceLabelName  = "kubearchive.org/enabled"
+	ApiServerSourceLabelValue = "true"
+	GlobalNamespace           = "___global___"
```
skoved, line 38: If you have the sink import this variable maybe it doesn't make sense for it to be in `cmd/operator/internal` since it would become shared

mafh314 (approved):

skoved (changes requested):

cmd/operator/internal/controller/common.go
skoved: at the same time, the SinkFiltersConfigMapName could also be imported by the sink to remove duplication of constants as well

cmd/sink/filters/filters.go
skoved: at the same time, the SinkFiltersConfigMapName could also be imported from the operator to remove duplication of constants as well and would allow us to remove `filtersCmName` from this file

cmd/operator/internal/controller/clusterkubearchiveconfig_controller.go
ggallen: It shouldn't.  We should use the imported value. I will remove the local declaration.

cmd/operator/internal/controller/clusterkubearchiveconfig_controller.go
ggallen: It shouldn't.  We should use the imported value. I will remove the local declaration.

cmd/operator/internal/controller/clusterkubearchiveconfig_controller.go
ggallen: Fixed.

cmd/operator/internal/controller/clusterkubearchiveconfig_controller.go
ggallen: Fixed.

cmd/sink/filters/filters.go
ggallen: Done. I did the import the other way, so we keep the value in the filters code.

cmd/operator/internal/controller/common.go
ggallen: Done.

cmd/sink/filters/filters.go
```diff
@@ -25,14 +25,17 @@ import (
 )
 
 const (
-	globalKeyEnvVar = "KUBEARCHIVE_NAMESPACE"
-	filtersCmName   = "sink-filters"
+	FiltersCmName        = "sink-filters"
+	GlobalNamespace      = "___global___"
```
skoved, line 29: I don't think that the sink should be exporting these variables. The sink already has to import `github.com/kubearchive/kubearchive/cmd/operator/api/v1alpha1` to parse the KubeArchiveConfigResources from yaml and having the operator start importing from the sink makes it easier for us to create an infinite import loop. I think the either the operator should export them or we could put them in a shared package, like `pkg/cel`

ggallen: Fixed!

skoved (approved):

PR #1028, opened 2025-04-17
Fix empty log annotation use-case

### LINKED ISSUE #1027: API returns 404 if pod has empty log selector annotation
If a Pod has an empty annotation for log selection (`kubectl.kubernetes.io/default-container`) the API returns a 404 for logs, even if there are logs.

**Acceptance Criteria**

* The API should return the first container if the annotation is empty.

rh-hemartin:

Resolve #1027  

## Release Note

```release-note
Fixed bug where the API did not return logs if the annotation `kubectl.kubernetes.io/default-container` had an empty value. Now the API returns the first container of the pod instead.
```

## Notes for Reviewers

* Simplified the related integration tests.

### DIFF
--- a/test/integration/logging_test.go
+++ b/test/integration/logging_test.go
@@ -159,110 +159,89 @@ func TestDefaultContainer(t *testing.T) {
 			},
 		},
 	}
-
 	test.CreateKAC(t, namespaceName, resources)
 
-	// Create a pod
-	pod := &corev1.Pod{
-		ObjectMeta: metav1.ObjectMeta{
-			Name: "defaults-to-first",
+	tests := []struct {
+		podName     string
+		annotations map[string]string
+		expectedLog string
+	}{
+		{
+			podName:     "defaults-to-first",
+			annotations: map[string]string{},
+			expectedLog: "I'm the container called first.",
 		},
-		Spec: corev1.PodSpec{
-			RestartPolicy: corev1.RestartPolicyNever,
-			Containers: []corev1.Container{
-				{
-					Name:    "first",
-					Image:   "quay.io/fedora/fedora:latest",
-					Command: []string{"echo", "-n", "I'm the container called first."},
-				},
-				{
-					Name:    "second",
-					Image:   "quay.io/fedora/fedora:latest",
-					Command: []string{"echo", "-n", "I'm the container called second."},
-				},
-			},
-		},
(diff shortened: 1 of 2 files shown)


skoved (approved):

test/integration/logging_test.go
```diff
@@ -158,109 +158,78 @@ func TestDefaultContainer(t *testing.T) {
...
-					Command: []string{"echo", "-n", "I'm the container called first."},
-				},
-				{
-					Name:    "second",
-					Image:   "quay.io/fedora/fedora:latest",
-					Command: []string{"echo", "-n", "I'm the container called second."},
-				},
-			},
+		"empty-annotation": map[string]string{
+			"kubectl.kubernetes.io/default-container": "",
```
maruiz93, line 169: Instead of going with the separate structs for annotations and expected logs we could use [a map to store all the test cases](https://go.dev/wiki/TableDrivenTests#using-a-map-to-store-test-cases).
It's also more clear when you see the parameters and the expectations in the same place.

In the unit tests we use a list of structs with the name on it:

```go
	tests := []struct {
		name        string
		annotations map[string]string
		expected    string
	}{
		{
			name:        "defaults-to-first",
			annotations: map[string]string{},
			expected:    "I'm the container called first.",
		},
		{
			name:        "wants-second",
			annotations: map[string]string{"kubectl.kubernetes.io/default-container": "second"},
			expected:    "I'm the container called second.",
		},
		{
			name:        "empty-annotations",
			annotations: map[string]string{"kubectl.kubernetes.io/default-container": ""},
			expected:    "I'm the container called first.",
		},
	}
```

While in the docs I linked a map is used:

```go
	tests := map[string]struct {
		annotations map[string]string
		expected    string
	}{
		"default-to-first": {
			annotations: map[string]string{},
			expected:    "I'm the container called first.",
		},
		"wants-second": {
			annotations: map[string]string{"kubectl.kubernetes.io/default-container": "second"},
			expected:    "I'm the container called second.",
		},
		"empty-annotation": {
			annotations: map[string]string{"kubectl.kubernetes.io/default-container": ""},
			expected:    "I'm the container called first.",
		},
	}
```

I don't have any preferences but I believe we should go with one of those.

rh-hemartin: Done!

PR #1050, opened 2025-04-29
Add cluster and namespace vacuums.

### LINKED ISSUE #991: Create a vacuum process for missed archives and deletes
# KubeArchive Vacuum

## Problem Statement

  ```
  ```

- Missed archives/deletes

## The Vacuums

### Cluster Vacuum

### Namespace Vacuum

### Performance

### Dependent Tasks

ggallen:

Resolves #991 

## Release Note

```release-note
Add namespace and cluster vacuums.
```


hack/kubearchive-install.sh
```diff
@@ -19,7 +19,8 @@ echo ${NEXT_VERSION}
 export NEXT_VERSION=${NEXT_VERSION}
 
 bash cmd/operator/generate.sh
-kubectl kustomize config/ | envsubst | ko apply -f - --base-import-paths
+kubectl kustomize config/crds | envsubst | kubectl apply -f -
+kubectl kustomize config/templates | envsubst | ko apply -f - --base-import-paths
```
rh-hemartin, line 23: If these change I think the release process should change as well

cmd/vacuum/main.go
```diff
@@ -0,0 +1,21 @@
...
+func main() {
+	typePtr := flag.String("type", "namespace", "Type of vacuum, either 'namespace' or 'cluster'.")
+	configPtr := flag.String("config", "", "Name of configuration file.")
+
+	flag.Parse()
+
+	if *typePtr == "cluster" {
+		clusterVacuum(*configPtr)
+	} else {
+		namespaceVacuum(*configPtr)
```
rh-hemartin, line 19: I think they should return an error and we should `os.Exit(1) on error, or something similar.

config/kustomization.yaml
rh-hemartin, line 1: To keep `hack/release.sh` and `hack/kubearchive-install.sh` you should keep this file but make it reference `crds/` and `templates/`, so `config/kustomization.yaml` acts as an aggregator.

hack/kubearchive-install.sh
rh-hemartin: I firmly believe you are breaking the release process with the changes on `config/`. This is the line on the `release.sh` that uses that folder to generate the single YAML:

```
kubectl kustomize config/ | envsubst | ko resolve -f - --base-import-paths --tags=${NEXT_VERSION} > kubearchive.yaml
```

If I run that on your PR it says:

```
$ kubectl kustomize config/ | envsubst | ko resolve -f - --base-import-paths --tags=${NEXT_VERSION}
error: unable to find one of 'kustomization.yaml', 'kustomization.yml' or 'Kustomization' in directory '/home/hemartin/git/kubearchive/config'

---
```

ggallen: Yeah, but if I make the change you indicate two comments down (adding `config/kustomization.yaml` back as an aggregator, it works.

rh-hemartin: Solved then

ggallen: @rh-hemartin, so I'm leaving this split in `hack/kubearchive-install.sh` but unchanged in `hack/release.sh`.

rh-hemartin: Could you test to install the `kubearchive.yml` from the release command on a fresh `kind`? I think the same error may happen for users installing. I think that we may need to split the installation for release as well, I don't think you can install a CRD and a resource using it on the same file.

config/templates/operator/vacuum.yaml
```diff
@@ -0,0 +1,129 @@
...
+# SPDX-License-Identifier: Apache-2.0
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: "kubearchive-cluster-vacuum"
+  namespace: kubearchive
+  labels:
+    app.kubernetes.io/name: "kubearchive-operator"
+    app.kubernetes.io/component: operator
```
rh-hemartin, line 11: It does not matter that much, but here we could use `vacuum`.

rh-hemartin (approved):

config/templates/operator/vacuum.yaml
ggallen: No, that's a good point.  I've changed all the resources in that file.

skoved (approved):

PR #1052, opened 2025-04-29
Add cluster and namespace vacuum documentation.

### LINKED ISSUE #991: Create a vacuum process for missed archives and deletes

ggallen:

Resolves #991 

## Release Note

```release-note
Add vacuum documentation.
```

### DIFF
--- a/docs/modules/ROOT/pages/reference/vacuum.adoc
+++ b/docs/modules/ROOT/pages/reference/vacuum.adoc
@@ -1,0 +1,339 @@
+= KubeArchive Vacuums
+
+== Use Cases
+KubeArchive relies on updates to resources to work. An ApiServerSource
+instance is watching for changes to particular resources, and when those
+changes occur it sends cloud events to the KubeArchive sink. Those cloud
+events are then processed by the sink, and if the criteria for archival
+and deletion are met, then those operations are performed.
+
+This works fine in many instances, but there are some glaring gaps.
+
+- Archive/delete after some interval
++
+  For example, take this CEL expression specified for `deleteWhen`:
++
+[source,yaml]
+----
+  deleteWhen: timestamp(status.completionTime) < now() - duration("2h")
+----
++
+  This indicates that the operation should occur when the `completionTime`
+of the resource is more than two hours old. However, there will be no event
+to trigger KubeArchive at that time.
+
+- Missed archives/deletes
++
+  There may be cases, such as KubeArchive downtime, where events are missed.
+In this case when KubeArchive restarts it has no idea that it has missed
+events.
+
+- Pre-KubeArchive installation resources
++
+  Resources exist on the cluster before KubeArchive is installed and
+functional. If those resources are not updated after KubeArchive is installed,
+they do not cause cloud events to be emitted that would cause KubeArchive to
+archive or delete them.
+
+== The Vacuums
+
+The solution to these problems is to utilize KubeArchive vacuums.
+This can be done at the cluster level (by an KubeArchive admin)
+or at the namespace level (by a namespace user).
+
+Vacuums can be run on a schedule using a Kubernetes `CronJob` or
+as a single instance using a `Job`.
+
+=== Cluster Vacuum
+
+The cluster vacuum is used to vacuum resources cluster-wide. A cluster 
+vacuum job must be run from the "kubearchive" namespace. A cluster
+vacuum job allows the KubeArchive admin to control and vacuum all
+namespaces configured for KubeArchive.
+
+==== Cluster Vacuum Configuration
+
+The cluster vacuum can vacuum any or all namespaces that have
+been configured for KubeArchive. For each namespace, any or
+all resources defined in the `ClusterKubeArchiveConfig` and the
+namespace `KubeArchiveConfig` can be vacuumed. The `ClusterVacuumConfig`
+defines a list of namespaces, each of which may specify specific
+resources, to be vacuumed.
+
+An empty list of namespaces in the `ClusterVacuumConfig` indicates that
+all namespaces configured for KubeArchive should be vacuumed.
+
+A list of namespaces in the `ClusterVacuumConfig` indicates that only
+those namespaces should be vacuumed.
+
+A namespace with no resources specified indicates that all resources
+defined in the `ClusterKubeArchiveConfig` and the namespace `KubeArchiveConfig`
+should be archived.
+
+A namespace with specific resources defined indicates that only those
+resources in the namespace should be vacuumed.
+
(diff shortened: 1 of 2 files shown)


rh-hemartin (approved):

docs/modules/ROOT/pages/reference/vacuum.adoc
```diff
@@ -0,0 +1,325 @@
...
+all resources defined in the `ClusterKubeArchiveConfig` and the
+namespace `KubeArchiveConfig` can be vacuumed. The `ClusterVacuumConfig`
+defines a list of namespaces, each of which may specify specific
+resources, to be vacuumed.
+
+An empty list of namespaces in the `ClusterVacuumConfig` indicates that
+all namespaces configured for KubeArchive should be vacuumed.
+
+A list of namespaces in the `ClusterVacuumConfig` indicates that only
+those namespaces shoudle be vacuumed.
```
rh-hemartin, line 67:
```suggestion
those namespaces should be vacuumed.
```

docs/modules/ROOT/pages/reference/vacuum.adoc
```diff
@@ -0,0 +1,325 @@
...
+    namespace-1: {}
+    namespace-2:
+      resources:
+        - apiVersion: v1
+          kind: Pod
+----
+
+==== Sample Cluster Vacuum Cronjob
+Following is sample Cronjob that can be used to run a cluster vacuum.
+The following fields in the `Cronjob` may need to be modified.
```
rh-hemartin, line 168: I think here we should specify why the user would want to create a CronJob of its own, since there is already one configured by default.

docs/modules/ROOT/pages/reference/vacuum.adoc
```diff
@@ -0,0 +1,325 @@
...
+apiVersion: kubearchive.kubearchive.org/v1alpha1
+kind: NamespaceVacuumConfig
+metadata:
+  name: name
+  namespace: namespace
+spec:
+    resources: {}
+----
+
+This `NamespaceVacuumConfig` will vacuum only `Job` and `Pod` resources in
```
rh-hemartin, line 257:
```suggestion
This `NamespaceVacuumConfig` vacuums `Job` and `Pod` resources in
```

docs/modules/ROOT/pages/reference/vacuum.adoc
```diff
@@ -0,0 +1,325 @@
...
+resources in the namespace should be vacuumed.
+
+The special namespace "___all-namespaces___" indicates that all namespaces
+should be processed. As a namespace, "___all-namespaces___" may also 
+specify a list of resources. If "___all-namespaces___" is specified,
+then all namespaces *not* listed in the `ClusterVacuumConfig` will be 
+vacuumed according to the settings for "___all-namespaces___" and
+all listed namespaces will be vacuumed as configured.
+
+The following `ClusterVacuumConfig` will vacuum all resources
```
rh-hemartin, line 83:
```suggestion
The following `ClusterVacuumConfig` vacuums all resources
```

docs/modules/ROOT/pages/reference/vacuum.adoc
```diff
@@ -0,0 +1,325 @@
...
++
+  There may be cases, such as KubeArchive downtime, where events are missed.
+In this case when KubeArchive restarts it has no idea that it has missed
+events.
+
+- Pre-KubeArchive installation resources
++
+  Resources exist on the cluster before KubeArchive is installed and
+functional. If those resources are not updated after KubeArchive is installed,
+they will not cause cloud events to be emitted that would cause KubeArchive to
```
rh-hemartin, line 35:
```suggestion
they do not cause cloud events to be emitted that would cause KubeArchive to
```

docs/modules/ROOT/pages/reference/vacuum.adoc
ggallen: Good point, but note that the sample job configured is suspended so it won't run.  I'll add that too.

docs/modules/ROOT/pages/reference/vacuum.adoc
ggallen: Fixed.

docs/modules/ROOT/pages/reference/vacuum.adoc
```diff
@@ -0,0 +1,325 @@
...
+The namespace vacuum can process all resources in the namespace defined
+in the `ClusterKubeArchiveConfig` and the local `KubeArchiveConfig`. Which
+resources are actually vacuumed are determined by the `NamespaceVacuumConfig`
+custom resource. The `NamespaceVacuumConfig` lists the resources that 
+should be vacuumed by API version and kind. An empty list of resources
+in the`NamespaceVacuumConfig` indicates that all resources specified
+in both the `ClusterKubeArchiveConfig` and the local `KubeArchiveConfig`
+should be vacuumed.
+
+The following `NamespaceVacuumConfig` will vacuum all resources in the
```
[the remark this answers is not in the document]
ggallen: Fixed.

docs/modules/ROOT/pages/reference/vacuum.adoc
ggallen: Fixed.

docs/modules/ROOT/pages/reference/vacuum.adoc
```diff
@@ -0,0 +1,325 @@
...
+
+A namespace with specific resources defined indicates that only those
+resources in the namespace should be vacuumed.
+
+The special namespace "___all-namespaces___" indicates that all namespaces
+should be processed. As a namespace, "___all-namespaces___" may also 
+specify a list of resources. If "___all-namespaces___" is specified,
+then all namespaces *not* listed in the `ClusterVacuumConfig` will be 
+vacuumed according to the settings for "___all-namespaces___" and
+all listed namespaces will be vacuumed as configured.
```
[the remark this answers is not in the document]
ggallen: Fixed.

docs/modules/ROOT/pages/reference/vacuum.adoc
ggallen: Fixed.

docs/modules/ROOT/pages/reference/vacuum.adoc
ggallen: Fixed.

skoved (approved):

PR #1053, opened 2025-04-30
Fix vacuum documentation link.

### LINKED ISSUE #991: Create a vacuum process for missed archives and deletes

ggallen:

Resolves #991 

## Release Note

```release-note
NONE
```


skoved (approved):

fnasser:
LGTM

fnasser (approved):
LGTM

PR #1058, opened 2025-05-02
Bump APIGroup to v1

### LINKED ISSUE #803: Promote KubeArchiveConfig to v1 from v1alpha
In preparation to promote KubeArchive to v1.0.0 I think we need to promote our KubeArchiveConfig from v1alpha to v1. I'm not sure if we will use it, as keeping it versioned has its downs, but at least we can keep it on v1 for ever.

**Acceptance Criteria**

* Bump KubeArchiveConfig to v1 from v1alpha1, it could be considered as a breaking change, but since KubeArchive is on v0 I don't think its that relevant.

rh-hemartin:

Resolves #803 

## Release Note

```release-note
Changed KubeArchive custom resources from `v1alpha` to `v1`.
```

### DIFF
--- a/test/log-generators/pipelines/kaconf.yaml
+++ b/test/log-generators/pipelines/kaconf.yaml
@@ -1,7 +1,7 @@
 # Copyright KubeArchive Authors
 # SPDX-License-Identifier: Apache-2.0
 ---
-apiVersion: kubearchive.org/v1alpha1
+apiVersion: kubearchive.org/v1
 kind: KubeArchiveConfig
 metadata:
   name: kubearchive
--- a/cmd/operator/api/v1/clusterkubearchiveconfig_types.go
+++ b/cmd/operator/api/v1/clusterkubearchiveconfig_types.go
@@ -1,7 +1,7 @@
 // Copyright KubeArchive Authors
 // SPDX-License-Identifier: Apache-2.0
 
-package v1alpha1
+package v1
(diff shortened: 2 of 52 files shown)


ggallen:
We should discuss this before this change is merged.  I have added it to the agenda for our next meeting as @maruiz93 had questions about it.

rh-hemartin:
> We should discuss this before this change is merged. I have added it to the agenda for our next meeting as @maruiz93 had questions about it.

Yes, I'm playing around with it in case we want to see how it looks. I'm also discovering potential issues that can happen when upgrading.

ggallen (approved):

test/log-generators/pipelines/kaconf.yaml
```diff
@@ -14,8 +14,8 @@ spec:
     - selector:
         apiVersion: tekton.dev/v1
         kind: TaskRun
-      archiveOnDelete: "true"
```
ggallen, line 17: Why all the changes from double to single quotes?

Maybe we should enforce something so we are consistent?

https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.quoted_strings

rh-hemartin: I fixed the changes, now my editor works with double quote.

skoved (approved):

PR #1075, opened 2025-05-13
Add doc entry about delayed deletes

### LINKED ISSUE #906: Add a delay to delete resources
Currently we delete immediately a resource when we need to, this caused some problems related with fluentbit, that didn't have enough time to scan resources to fill log metadata causing logs to be sent without Kubernetes metadata. Other systems may want a delayed deletion too, so other systems that depend on the resource can react before the resource gets deleted.

**Acceptance Criteria**
* Introduce a configurable delay for deleting resources.

**Notes**
* Implementation-wise there are many options, some are: create a pool of resources to be deleted and have a timer that executes a function over that pool, wait on the same endpoint that needs to delete the resource, etc.

### LINKED ISSUE #951: Investigate performance impact of using different timers to create a delay before deleting resources
investigation related to #906 

We had some discussion about using timers vs periodically checking the database. I'm going to create a spike to investigate the performance of just using time.AfterFunc and other test for the performance of creating a worker thread that receives a time.Timer and a function to delete a resource on the cluster. It puts them in a queue, waits for the next timer in the queue to expire and then runs the function to delete the resource, and then moves to the next timer in the queue

To test the performance, I'm going to create a script that will use curl to send as many cloud events as it can in a minute. I'm gonna try to use parallel to send 3 cloud events at a time in parallel to try and create extra load. During the test, I will use the pprof endpoints that the sink provides to see what the load looks like in both instances

rh-hemartin:

Resolves #906, resolves #951 

## Release Note

```release-note
Added documentation page about Delayed Deletes in the configuration section.
```

### DIFF
--- a/docs/modules/ROOT/pages/configuration/delayed-deletes.adoc
+++ b/docs/modules/ROOT/pages/configuration/delayed-deletes.adoc
@@ -1,0 +1,140 @@
+= Delayed Deletes
+
+This document describes how to configure KubeArchive when resources need
+to be deleted after a certain time.
+
+KubeArchive features vacuums that rescan namespaces for stale or missed
+resources. Vacuums in combination with filters that include time expressions
+allow users to delay the deletion of resources. More information about Vacuums
+xref:/reference/vacuum.adoc[here].
+
+This feature is useful for resources that are monitored or modified by other
+systems without any flag on the resource itself that can be used by KubeArchive.
+For example metadata collection systems or build signature systems.
+
+== KubeArchiveConfig
+The following `KubeArchiveConfig` resource configures KubeArchive to delete
+`Pod` ten seconds after they started if their `status.phase` is `Succeeded`:
+
+[source,yaml]
+----
+apiVersion: kubearchive.org/v1
+kind: KubeArchiveConfig
+metadata:
+  name: kubearchive
+spec:
+  resources:
+    - selector:
+        apiVersion: v1
+        kind: Pod
+      deleteWhen: status.phase == "Succeeded" && timestamp(status.startTime) < now() - duration("10s")
+----
+
+Given a `Pod` that just prints `Hello`, when the latest update for it is received
+by KubeArchive it is likely that the condition for deletion does not yet evaluate
+to true.
+
+== Vacuum
+Create the following `NamespaceVacuumConfig` to configure the vacuum to scan all
+resources configured for the namespace:
+
+[source,yaml]
+----
+apiVersion: kubearchive.org/v1
+kind: NamespaceVacuumConfig
+metadata:
+  name: all
+spec:
+    resources: []
+----
+
+This won't trigger anything by itself, you need something that runs the actual vacuum
+process.
+
+== CronJob
+The following `CronJob` executes the vacuum process for the vacuum configuration
+created in the previous section:
+
+[source,yaml]
+----
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: all-vacuum
+spec:
+  schedule: "* * * * *"  # Execute each minute
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          serviceAccount: kubearchive-vacuum
+          containers:
+            - name: vacuum
+              image: quay.io/kubearchive/vacuum:<your KubeArchive version>
+              command: [ "/ko-app/vacuum" ]
+              args:
+                - --config
+                - all  # name of the NamespaceVacuumConfig
+              env:
+                - name: NAMESPACE
+                  valueFrom:
+                    fieldRef:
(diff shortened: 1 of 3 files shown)


docs/modules/ROOT/pages/configuration/delayed-deletes.adoc
```diff
@@ -0,0 +1,140 @@
+= Delayed Deletes
+
+This document describes how to configure KubeArchive when resources need
+to be deleted after a certain time.
+
+KubeArchive features vacuums that reescan namespaces for stale or missed
```
skoved, line 6: typo
```suggestion
KubeArchive features vacuums that rescan namespaces for stale or missed
```

docs/modules/ROOT/pages/configuration/delayed-deletes.adoc
```diff
@@ -0,0 +1,140 @@
...
+resources. Vacuums in combination with filters that include time expressions
+allow users to delay the delete of resources. More information about Vacuums
```
skoved, line 8: typo
```suggestion
allow users to delay the deletion of resources. More information about Vacuums
```

docs/modules/ROOT/pages/configuration/delayed-deletes.adoc
```diff
@@ -0,0 +1,140 @@
...
+xref:/reference/vacuum.adoc[here].
+
+This feature is useful for resources that are monitored or modified by other
+systems without any flag on the resource itself that can be used by KubeArchive.
+For example metadata collection systems, build signature systems...
```
skoved, line 13:
```suggestion
For example metadata collection systems, build signature systems, etc.
```

rh-hemartin: We don't like latin abbreviations: https://kubearchive.github.io/kubearchive/main/contributors/documentation.html#_latin_abbreviations

docs/modules/ROOT/pages/configuration/delayed-deletes.adoc
rh-hemartin: Fixed, thanks.

skoved (approved):

docs/modules/ROOT/pages/configuration/delayed-deletes.adoc
```diff
@@ -0,0 +1,140 @@
...
+metadata:
+  name: vacuum-all
+  namespace: kubearchive
+spec:
+  schedule: "* */1 * * *"  # Execute the first minute of each hour
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          serviceAccount: kubearchive-vacuum
```
ggallen, line 121: I think the service account is wrong here.  Check the vacuum documentation.  I think it's `kubearchive-cluster-vacuum`?

docs/modules/ROOT/pages/configuration/delayed-deletes.adoc
```diff
@@ -0,0 +1,140 @@
...
+            - name: vacuum
+              image: quay.io/kubearchive/vacuum:<your KubeArchive version>
+              command: [ "/ko-app/vacuum" ]
+              args:
+                - --type
+                - cluster
+                - --config
+                - all  # name of the ClusterVacuumConfig
+              env:
+                - name: NAMESPACE
```
ggallen, line 132: I think this needs to be `KUBEARCHIVE_NAMESPACE` for the cluster vacuum?

docs/modules/ROOT/pages/configuration/delayed-deletes.adoc
```diff
@@ -0,0 +1,140 @@
...
+          serviceAccount: kubearchive-vacuum
+          containers:
+            - name: vacuum
+              image: quay.io/kubearchive/vacuum:<your KubeArchive version>
+              command: [ "/ko-app/vacuum" ]
+              args:
+                - --type
+                - cluster
+                - --config
+                - all  # name of the ClusterVacuumConfig
```
ggallen, line 130: Should this be `vacuum-config-all` from the example config above?

rh-hemartin: Fixed, thanks!

docs/modules/ROOT/pages/configuration/delayed-deletes.adoc
rh-hemartin: Fixed, thanks!

docs/modules/ROOT/pages/configuration/delayed-deletes.adoc
rh-hemartin: Fixed, thanks!

ggallen (approved):

PR #1080, opened 2025-05-14
Sink no configured resources

### LINKED ISSUE #1029: sink should be more clear with resources that are ignored
When the sink receives a cloudevent that does not match any of the filters it logs the following message

> Resource was updated, no action is needed

https://github.com/kubearchive/kubearchive/blob/276a005987e6dc9fa7a6bed150ec50ba5012ca7f/cmd/sink/routers/routers.go#L178

There are two cases to handle here:

1. resource is configured for archiving, but this specific resource does not match the the filter criteria
2. resource is not configured for archiving at all

for case 1, this message is too ambiguous and a user may assume that kubearchive updated the resource in the database (the wrong behavior) instead of understanding that the resource was updated on the cluster, but does require any process by kubearchive because none of the filters apply (correct behavior). 

I don't have a suggestion for clearer wording **YET** so feel free to leave suggestions if you have any

for case 2, we should have a different message that says something like

> Resource update received. This resource kind is not configured for archiving

additionally, we could return a different response code in this case. If we do that, please consult [this table](https://github.com/knative/specs/blob/main/specs/eventing/data-plane.md#event-acknowledgement-and-delivery-retry) from the knative spec to understand how knative interprets different http response codes. We would need a response code that indicates an error and that retries should not be attempted

rh-hemartin:

Resolves #1029 

## Release Note

```release-note
Added a new label value for `kubearchive.cloudevents.total` to identify cloudevents from resources that are not configured. Read the observability documentation for more information.
```

### DIFF
--- a/cmd/sink/filters/fake/filters.go
+++ b/cmd/sink/filters/fake/filters.go
@@ -51,3 +51,7 @@ func (f *Filters) MustArchiveOnDelete(ctx context.Context, obj *unstructured.Uns
 	}
 	return false
 }
+
+func (f *Filters) IsConfigured(ctx context.Context, obj *unstructured.Unstructured) bool {
+	return f.MustArchiveOnDelete(ctx, obj) || f.MustArchive(ctx, obj) || f.MustDelete(ctx, obj)
+}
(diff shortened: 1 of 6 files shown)


skoved:
I agree that this PR is related to #1029 but I don't think that it should resolve the issue since this PR does not change any of the log messages that the sink generates

rh-hemartin:
> I agree that this PR is related to #1029 but I don't think that it should resolve the issue since this PR does not change any of the log messages that the sink generates

I totally forgot about that part, good catch.

skoved (approved):

maruiz93 (approved):

PR #1115, opened 2025-06-02
Inserting resources and logs are now a single transaction

### LINKED ISSUE #816: Sink should rollback database transaction for writing a resource if its logs could not be written
The sink has to generate the urls of logs for Kubernetes resources like Pods that have associated logs. Right now, it is possible for a Pod to be written to the resource table and then not have the urls for its logs written to the database because there was an error generating the urls OR the was error during the database transaction. Based on discussion we had, if either of these errors occur, the write of the resource to the database should be rolled back.

**Acceptance Criteria**:

- The write of a resource to the database should rollback IF it has logs AND those logs could not be written to the database

rh-hemartin:

Resolves #816 

## Release Note

```release-note
Changes resource archival to fail if KubeaArchive fails to write the logs.
```

### DIFF
--- a/cmd/sink/routers/routers.go
+++ b/cmd/sink/routers/routers.go
@@ -63,60 +63,42 @@ func NewController(
 	return controller, nil
 }
 
-func (c *Controller) writeLogs(ctx context.Context, obj *unstructured.Unstructured) error {
-	logCtx, cancel := context.WithTimeout(ctx, time.Second*5)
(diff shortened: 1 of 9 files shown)


rh-hemartin:
I'm having trouble because the logs are related to entries on the resources table, so I can't write the logs first and then the resource.

skoved (approved):

PR #1123, opened 2025-06-06
Merge API router tests

### LINKED ISSUE #713: Merge tests when possible
In some test files we use the approach of one function per test instead of having a set of parameters and iterate through them applying the same test over and over.

Making this change can be very useful when the tests are almost identical except for a couple of variables.

A good example of a file that can be improved is `cmd/api/routers/routers_test.go`

rh-hemartin:

Resolves #713 

## Release Note

```release-note
NONE
```

### DIFF
--- a/cmd/api/routers/routers_test.go
+++ b/cmd/api/routers/routers_test.go
@@ -139,57 +139,103 @@ func TestLabelSelectorQueryParameter(t *testing.T) {
 	}
 }
 
-func TestGetCoreResourcesLogURLs(t *testing.T) {
-	router := setupRouter(fake.NewFakeDatabase(testResources, testLogUrls, testLogJsonPath), true)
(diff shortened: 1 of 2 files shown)


maruiz93 (changes requested):
LGTM but could you please add the same change to pkg/logurls/logurls_test.go?

I think that the load*JSON functions can be unified too.

rh-hemartin:
> logurls_test

I merged the load*JSON functions using generics. Note that they are not testing functions but helper functions.

maruiz93:
> > I think that the load*JSON functions can be unified too.
> 
> I merged the load*JSON functions using generics. Note that they are not testing functions but helper functions.

right! BTW I think it looks better now with parametrics, less boilerplate and easier to read :)

skoved (approved):

PR #1139, opened 2025-06-16
Fix GolangCI Lint problems

rh-hemartin:

Resolves #1120 

## Release Note

```release-note
NONE
```


maruiz93:
What are you trying to solve here? the linked issue is not an issue but a PR. Are you having issues with the ci linter with the current code that get solved with this PR? which ones?

maruiz93:
> > What are you trying to solve here? the linked issue is not an issue but a PR. Are you having issues with the ci linter with the current code that get solved with this PR? which ones?
> 
> I'm trying to solve the GolangCI issues we have in our repository. The workflow should create an issue, I forgot that and wanted to link the previous issue, but I linked a PR instead. The workflow is failing to create an issue, I noticed it lacked issues permissions, I created this other PR so the workflow can create issues:
> 
> #1140

Ok so you should change the description of the PR from 

```
Resolves #1120
```

to 

```
Resolves #1119
``` 

right?
Or are you waiting for #1140 to be merged so the ci workflow can create the issue related to the fix on this PR?

Also, can you link the workflow with the specific failure?
I see this one https://github.com/kubearchive/kubearchive/actions/runs/15674782322 in the history of golangci-lint workflow executions but I don't see how this is related to the changes on this PR. I'm probably missing something here.

PR #1205, opened 2025-07-16
Add creationTimestamp filters in API

### LINKED ISSUE #1193: Add date-range filter based on creation timestamp for resource retrieval
### Context

As a tool whose purpose is to archive a potential large amount of data for a large date-range it makes sense to add a filter that is not available in Kubernetes API for filtering resources based on the creation timestamp

### Acceptance criteria

* The endpoints for querying both core and non-core resources accept filters for stating "created before than" and "created after than"
* There is a method in the database interface implemented in the PostgreSQL struct that applies this filter.

maruiz93:
Assisted-by: Cursor AI

Resolves #1193  

## Release Note

```release-note
Adds creationTimestampBefore and creationTimestampAfter filters on resource endpoints
```

## Notes for Reviewers

You can test the API call with this:

```
curl -k -H "Authorization: Bearer $(kubectl create token default -n test)" "https://localhost:8081/api/v1/pods?creationTimestampBefore=2025-07-16T15:19:00Z" | jq '.items|length'
```

You can try changing the date and compare it with the length without the filter after generating some activity with the cronjobs

### DIFF
--- a/cmd/api/routers/routers.go
+++ b/cmd/api/routers/routers.go
@@ -61,6 +61,26 @@ func (c *Controller) GetResources(context *gin.Context) {
 		return
 	}
 
+	// Parse timestamp filters
+	creationTimestampAfter, err := parseTimestampQuery(context, "creationTimestampAfter")
+	if err != nil {
+		abort.Abort(context, err, http.StatusBadRequest)
+		return
+	}
+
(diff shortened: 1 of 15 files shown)


rh-hemartin:
Do we want to prevent people from doing: before=Jan2025 after=March2025?

PR #1234, opened 2025-07-30
increase timeout for sink readyz enpoint and use Get instead of List for SinkFilters

### LINKED ISSUE #1233: Sink's readiness probe hits timeout
When KubeArchive is deployed in konflux we found that the kubernetes api server can experience periods where it takes a long time to respond to request ( > 1s) when under heavy load. Additionally, the list operation for resources always is slower than the get operation. Under load we have seen the list operations 99th percentile performance reach up to 60s vs ~1.5s for get. Since we always know the name of the sinkfilters object in the cluster, we should do a get for the sinkfilters instead of list. We should also increase the timeout for the readiness probe to 4s from the default of 1s

Acceptance criteria:
- readiness probe does get for SinkFilters
- readiness probe is configured with a timeout of 4s

skoved:

Resolves #1233  

## Release Note

```release-note
NONE
```

## Notes for Reviewers

Increase timeout for readyz because when the kubernetes api server is under pressure, it can take more than 1 second to respond. Switching from List to Get because Get is more performant

### DIFF
--- a/.github/workflows/copy-labels.yml
+++ b/.github/workflows/copy-labels.yml
@@ -29,6 +29,8 @@ jobs:
     name: Add assignee to the PR
     if: github.actor != 'kubearchive-renovate[bot]'
     steps:
+      - name: Checkout the repository
+        uses: actions/checkout@v4
       - name: Check if user can be assigned
         id: check-assignable
         run: |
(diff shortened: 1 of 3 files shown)


maruiz93 (approved):

skoved:
I've added a commit to fix the issue this PR hit with PR assignment action

rh-hemartin (approved):

mafh314 (approved):

PR #1250, opened 2025-08-06
instrument k8s dynamic client for the sink

### LINKED ISSUE #1249: Add otel instrumentation to the dynamic client used by the sink
We should add otel instrumentation to the dynamic client used by the sink so that we can generate traces and metrics for k8s api request made by the sink.

To see how to do this look at [this code in the api server](https://github.com/kubearchive/kubearchive/blob/main/cmd/api/main.go#L57)

Acceptance Criteria:
- when otel is enabled in the sink, the dynamic client will generate the expected traces and/or metrics

skoved:

Resolves #1249 

## Release Note

```release-note
Add spans for kubernetes api requests made by the sink when observability is enabled
```

## Notes for Reviewers

use `otelhttp.NewTransport` to add otel instrumentation on requests the sink makes to the k8s api server

### DIFF
--- a/cmd/sink/k8s/client.go
+++ b/cmd/sink/k8s/client.go
@@ -5,7 +5,9 @@ package k8s
 
 import (
 	"fmt"
+	"net/http"
 
+	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/dynamic"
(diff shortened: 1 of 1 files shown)


rh-hemartin (approved):

maruiz93:
FYI, I've been playing around with this metrics and the problem is that by default the `path` is instrumented but that's a high cardinality attribute so usually those are discarded in the collector.

I'm taking a look at having a route attribute in place which gives us information of the nature of the interaction being done with k8s API without having a high cardinality attribute.

maruiz93:
I think this PR may affect the end user, I would include some release notes

maruiz93 (changes requested):
PTAL at https://github.com/kubearchive/kubearchive/pull/1254 I addressed this and other changes so that the HTTP k8s client instrumentation gives us more insghtful data

PR #1271, opened 2025-08-12
Create spans on Sink functions

### LINKED ISSUE #1269: Add more detailed instrumentation in the sink
When looking at the sink traces we have a big gap of spans before the sql instrumentation:

Image

We should have at least the following spans:

- Cloud event handler
- Evaluate filters
- Write resource
- Evaluate delete filters
- Delete resource

rh-hemartin:

Resolves #1269 

## Release Note

```release-note
Adds OpenTelemetry spans to various functions executed in the KubeArchive Sink to improve observability.
```

## Notes for Reviewers

Traces will look like this now:

image

image

image

image

### DIFF
--- a/cmd/sink/filters/filters.go
+++ b/cmd/sink/filters/filters.go
@@ -14,6 +14,7 @@ import (
 	kubearchiveapi "github.com/kubearchive/kubearchive/cmd/operator/api/v1"
 	ocel "github.com/kubearchive/kubearchive/pkg/cel"
 	"github.com/kubearchive/kubearchive/pkg/constants"
+	"go.opentelemetry.io/otel"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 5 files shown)


skoved:
idk if this would be better as a metric or as an attribute on the span but we could calculate the time that `changeFilters` function has to wait for for the mutex as well as the amount of time that it locks out all of the readers (meaning that all of the request handlers have to wait).

We could calculate them like this
```
requestLock := time.Now()
f.Lock()
getLock := time.Now()
waitForLock := getLock.Sub(requestLock)

... // old filters removed and new filters added

lockedTime := time.Since(getLock)

// end of function
```

What do you think?

rh-hemartin:
> Could you pleas add a pic of a trace where it also handles deletion?

I added one, let me know what you think.

maruiz93:
> @skoved I added events, they look like this:
> 
> image
> The time on the events is relative to the span, so in this case these were emitted at the end of the span life.

Now that you added the events for the "write-lock" how about adding them for the read too? (calls to `RLock`) I don't know about the cost of adding them but maybe they are useful at some poing (even though I agree that the write ones are more critical)

maruiz93 (approved):

rh-hemartin:
> > @skoved I added events, they look like this:
> > image
> > The time on the events is relative to the span, so in this case these were emitted at the end of the span life.
> 
> Now that you added the events for the "write-lock" how about adding them for the read too? (calls to `RLock`) I don't know about the cost of adding them but maybe they are useful at some poing (even though I agree that the write ones are more critical)

I don't think that adding events to the read locks will be useful. Its a `RWMutex` which reads: `The lock can be held by an arbitrary number of readers or a single writer.` so it is far more interesting to have the lock on the single writer we have. If we start having problems that resemble a locking problem we will see that the writer locking takes time to complete. Also the reads happen much more often, so this way we avoid sending even more data.

skoved (approved):

PR #1294, opened 2025-08-25
sink only writes resources to db from namespaces configured by users

### LINKED ISSUE #1293: Sink should check if the namespace of a resource should be monitored if there is a match on global filters
We have encountered issues with ApiServersSource's watches on very large clusters. One solution is to have ApiServerSource watch specified resources cluster wide. This means that the sink will receive resources from namespaces that users do not want monitored. The sink should ignore these resources

Acceptance Criteria:

- the sink does not write resources to the database that match a global filter and are not in a namespace the user indicated kubearchive to monitor

skoved:

Resolves #1293 

## Release Note

```release-note
NONE
```

## Notes for Reviewers

Add a check in the sink when executing filters to make sure that resources that match the global filters only get written to the database if the resource is from a namespace that kubearchive is configured to watch.

### DIFF
--- a/cmd/sink/filters/filters.go
+++ b/cmd/sink/filters/filters.go
@@ -37,6 +37,7 @@ type Filters struct {
 	archive         map[NamespaceGroupVersionKind]cel.Program
 	delete          map[NamespaceGroupVersionKind]cel.Program
 	archiveOnDelete map[NamespaceGroupVersionKind]cel.Program
+	namespaces      map[string]struct{}
 }
 
 // NewFilters creates a Filters struct with empty archive, delete, and archiveOnDelete slices.
@@ -47,6 +48,7 @@ func NewFilters(dynclient dynamic.Interface) *Filters {
 		archive:         make(map[NamespaceGroupVersionKind]cel.Program),
 		delete:          make(map[NamespaceGroupVersionKind]cel.Program),
 		archiveOnDelete: make(map[NamespaceGroupVersionKind]cel.Program),
+		namespaces:      make(map[string]struct{}),
 	}
 }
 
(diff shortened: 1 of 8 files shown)


cmd/sink/filters/filters.go
```diff
@@ -60,7 +62,11 @@ func (f *Filters) changeFilters(namespaces map[string][]kubearchiveapi.KubeArchi
 	archiveMap := make(map[NamespaceGroupVersionKind]cel.Program)
 	deleteMap := make(map[NamespaceGroupVersionKind]cel.Program)
 	archiveOnDeleteMap := make(map[NamespaceGroupVersionKind]cel.Program)
+	namespacesMap := make(map[string]struct{})
 	for namespace, kacResources := range namespaces {
+		if namespace != constants.SinkFilterGlobalNamespace {
+			namespacesMap[namespace] = struct{}{}
+		}
```
ggallen, line 69: I'm not sure this works as namespaces with no resources do not get an entry in the `SinkFilter` object. For Konflux this will result in nothing getting archived as everything is controller in the `ClusterKubeArchiveConfig`.

We may have to change [this line](https://github.com/kubearchive/kubearchive/blob/main/cmd/operator/internal/controller/common.go#L313) as well.

If we do make this change, then you can probably check `sinkFilter.spec.Namespaces` directly, and there will be no need for the namespace set that is created in this PR. We also have to make there sure there are no other side-effects from having this empty entry.  For example, we may end up with extra (empty) entries in the maps in the sink.

ggallen: I think the check can probably use `slices.Contains` rather than a for loop.
```
if slices.Contains(sf.spec.Namespaces, namespace) {
   ...
}
```

ggallen: > but we end up using more memory than we need.

How are we using more memory?  We're using the same amount of memory that's currently in use. In that case it means your solution is adding more memory than we need, unless I am missing something.

> slice.contains runtime is linear, where f.namespaces[namespace] runtime is constant 

Then just use `sinkFilter.spec.Namespaces[namespace]`?  It's a map.

I'm sorry I suggested `slices.Contains`.

skoved:
rebased this PR from main to allow it to merge

maruiz93 (approved):

PR #1317, opened 2025-09-02
Introduce 'KLOG_LEVEL' env variable

### LINKED ISSUE #1282: Introduce klog verbosity level
As KubeArchive uses the k8s library, which works with `klog` it may be worth to introduce a way to specify the verbosity. `klog` attaches to a flagset where it adds the parameter `-v` to specify the level (for example `-v=9`). You can fake that using a flagset for `klog` and setting the flag from code:

```
	var fs flag.FlagSet
	klog.InitFlags(&fs)
	_ = fs.Set("v", "9")
```

This block would go in the `main` function of the different CMDs, or into the `setupLogging` function all of the CMDs share.

We may introduce a variable called `K8S_LOG_LEVEL`, `KLOG_LEVEL` or select a value given the current `LOG_LEVEL` we are exposing.

**Acceptance Criteria**

* Add a mechanism to set the klog log level so we have more complete logs when debugging.

rh-hemartin:

Resolves #1282 

## Release Note

```release-note
Introduce 'KLOG_LEVEL' env variable to control logging level of Kubernetes libraries.
```


pkg/logging/logging.go
```diff
@@ -6,18 +6,21 @@ import (
...
 	"github.com/gin-gonic/gin"
 	"go.opentelemetry.io/otel/trace"
+	"k8s.io/klog/v2"
 )
 
 type contextKey string
 
 const (
 	loggingLevelEnvVar            = "LOG_LEVEL"
+	klogLevelEnvVar               = "KLOG_LEVEL"
```
ggallen, line 23: Should this name be consistent with `loggingLevelEnvVar`, i.e. `kLoggingLevelEnvVar`?

ggallen (approved):

PR #1336, opened 2025-09-10
Add a test to assist with the log forwarding issues

### LINKED ISSUE #1335: Create an integration test to ensure that we don't miss logs and that the order is respected
When deployed in a cluster with a vector+loki logging integration we are experiencing some issues regarding missing logs as well as order problems.

To help troubleshoot the matter, introduce an integration test that generates a large amount of entries in a short time to ensure that the order is respected and that we don't miss logs.

This could also lead to manipulating the current vector+loki config in the integration.

### Acceptance Criteria

* There is an integration tests that bursts with a good amount of logs in a short period of time
* The configuration of the vector+loki integration is set up to allow this to properly happen

maruiz93:

Resolves #1335 

## Release Note

```release-note
NONE
```

## Notes for Reviewers

With this test I've seen that when using vector+loki along with a "reduce" manipulation of the kubernetes logs by vector to be sent in batches to loki we need to specify the batch size per entry counts and not per time.
This means that we need to use `max_events` property instead of `flush_period_ms`.
The reason behind this is that with timing and a large amount of logs we will likely send to loki entries that have larger size than it expects and the logs then may be rejected.

This needs to be handled along with the proper truncation of very long messages being aware that we are not truncating one log entry but a concatenated amount of log entries so we could potentially miss information if we don't do it correctly.

With the provided configuration the new test function logs are stored in loki in batches of 100 messages (the max limit) and kubearchive is able to retrieve them complete and ordered.

Extra stuff addressed in the issue:

- The logs forwarded as text and not as json. This improves performance and the logs shouldn't be necessary in a json format.
- Incremented the maximum line size received in loki, considering that each line can gather up to 100 entries now.
- Allow the entries per query to be larger
- Enable the `increment_duplicate_timestamp` to make sure that the logs aren't disordered. In case the timestamp of two entries is the same, loki should apply a slight increment to the most recent one to make sure that both batches don't share the same timestamp. BTW, I wasn't able to reproduce having the same timestamp at nanosecond precission but I think it's a good idea to have it.
- Disable cached components for simplicity in our development environment
- Set up hardcoded credentials for grafana in development to improve the easy of use while developing
- Make the install.sh script rerunnable with `helm upgrade --install` instead of `helm install`
- Use only one label to ensure one stream per pod+container. This helps keeping the logs ordered in a distributed loki.

### DIFF
--- a/integrations/logging/loki/README.md
+++ b/integrations/logging/loki/README.md
@@ -33,12 +33,8 @@ Run the log generators to create logs:
     ```bash
     kubectl port-forward -n grafana-loki service/grafana 3000:80
     ```
-1. Determine the `admin` password for Grafana:
-    ```bash
-    echo `kubectl -n grafana-loki get secret grafana -o jsonpath='{.data.admin-password}' | base64 --decode`
-    ```
-1. In your browser, navigate to `http://localhost:3000`, using the username `admin` and the password retrieved
-   in the previous step to login.
+
+1. In your browser, navigate to `http://localhost:3000`, using `admin`/`password` credentials.
 
 1. Go to `Explore` and select `Loki`. Then write a query (e.g. `{pod="generate-log-1-29110151-kc2n9", container="generate1"}`).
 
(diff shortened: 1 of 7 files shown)


maruiz93:
This is an example of how the logs batched looks like in grafana:

image

olegbet (approved):

maruiz93:
I've found out another config issue that *could potentially lead* to unordered logs from loki. The reason behind it is that we are creating multiple streams for each container and loki doesn't ensure having one stream in that case.

I'll refactor vector config to use just one label that concatenates the `pod_uid`+`container_name`

Here is some extra context: https://github.com/grafana/loki/issues/10229#issuecomment-1718409349

Issue #1006, opened 2025-04-09
Replace sink-filters ConfigMap with a CRD


ggallen:
The data section for a `ConfigMap` is limited to 1024 characters. While this isn't a problem now, it's likely to become one in future.

The `ConfigMap` should replaced with a `SinkFilters` CRD, This CRD should be limited to one instance named "sink-filters" in the KubeArchive  installation namespace. The `SinkFilters` spec should contain a "Resources" field which will be a map of namespace name to an array of `KubeArchiveConfigResource` items. 

## Acceptance Criteria

- A new CRD is created that contains the filtering information for each namespace KubeArchive is watching.
- The operator updates the `SinkFilters` resource as changes happen to `KubeArchiveConfig` resources (instead of updating the `ConfigMap`)
- The sink changes to get its information form the `SinkFilters` resource in the installation directory.
- The "sink-filters" `ConfigMap` is removed.

ggallen:
> But maybe I'm missing something and there is a way to track and load changes of a Custom Resource that I'm not aware of that makes things easier.

We should be able to put a watch on this resource in the same way that the `ConfigMap` is being watched now.  I have already discussed this with @skoved, and I think it will work. I think the sink had to change from actually mounting the `ConfigMap`.

> Should we consider making it a Cluster resource?

I don't think it makes sense as a Cluster resource as it's only used in  KubeArchive internally by the sink and the operator. It's essentially a communication mechanism between the two.

I also see no reason to expose all the namespaces participating in KubeArchive, and their filters, to potentially everyone on the cluster.

maruiz93:
> I don't think it makes sense as a Cluster resource as it's only used in KubeArchive internally by the sink and the operator. It's essentially a communication mechanism between the two.

The reason I'm proposing is exactly because of this. We want it to be a singleton that stores a global configuration only used internally within kubearchive namespace so it makes sense to use the setup that natively behaves like a singleton. It's also  easy to restrict the access for this resource by kubearchive namespace owners and admins.

> I also see no reason to expose all the namespaces participating in KubeArchive, and their filters, to potentially everyone on the cluster.

I'm not sure how this is related. I don't see a reason to expose all the namespaces participating in KubeArchive, but getting the CRD to be a cluster-scoped resource instead of a namespaced one doesn't change that. It could be only available from other namespaces if their users/service accounts have permissions for that, the same as for a namespaced resource.

[...]

PR #1356, opened 2025-09-24
Add SinkFilters controller and make knative optional

### LINKED ISSUE #1355: Remove dependency on knative-eventing from KubeArchive.
The overhead of including knative-eventing to use `ApiServerSource` is more than it is worth. We should remove the dependency on knative-eventing, which will simplify many things:

- No need for any `ApiServerSource` resources for KubeArchive.
- No need to install knative-eventing resource on the cluster.
- No need for a production grade broker (e.g. Kafka) on the system.
- Smaller footprint for the KubeArchive development environment.

The only thing `ApiServerSource` is used for is to watch for events on resources and send them to the KubeArchive sink. It would be more beneficial and steamline for KubeArchive to just do the watching itself.

ggallen:

Resolves #1355

## Release Note

```release-note
NONE
```


cmd/operator/internal/controller/sinkfilter_controller_example.md
```diff
@@ -0,0 +1,231 @@
...
+```yaml
+# SinkFilter resources
+- groups: ["kubearchive.org"]
+  resources: ["sinkfilters"]
+  verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
+
+# All resources for dynamic informers
+- groups: ["*"]
+  resources: ["*"]
+  verbs: ["get", "list", "watch"]
```
rh-hemartin, line 180: From this I understand that this controller needs permissions for everything, is that correct? We talked about not wanting more permissions than need when talking about Knative Eventing and the "monitor all namespaces" solution. Why the change?

config/templates/operator/configmap.yaml
```diff
@@ -0,0 +1,13 @@
...
+# SPDX-License-Identifier: Apache-2.0
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: kubearchive-operator-config
+  namespace: kubearchive
+  annotations:
+    # This annotation causes the operator deployment to restart when the ConfigMap changes
+    reloader.stakater.com/match: "true"
```
rh-hemartin, line 11: This needs a separate controller: https://github.com/stakater/Reloader

ggallen: No, it's not being used.  I'll remove the comment.

cmd/operator/internal/controller/sinkfilter_controller_example.md
ggallen: This file should not have been committed. Let me look more closely.

ggallen: Fixed.

cmd/operator/internal/controller/sinkfilter_controller_example.md
```diff
@@ -0,0 +1,231 @@
...
+
+# Knative brokers for cloud events
+- groups: ["eventing.knative.dev"]
+  resources: ["brokers"]
+  verbs: ["get", "list", "watch"]
+
+# KubeArchive configs for publisher
+- groups: ["kubearchive.org"]
+  resources: ["kubearchiveconfigs", "clusterkubearchiveconfigs"]
+  verbs: ["get", "list", "watch"]
```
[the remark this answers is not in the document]
ggallen: Removed this file.

maruiz93 (changes requested):
I miss having a design page with this and  maybe also describing a little bit the fields and the goal of the SinkFilter CRD (I think we should've had that since https://github.com/kubearchive/kubearchive/issues/1006)

Could you please add some docs for this as well as the new `kubearchive-operator-config` ConfigMap?

pkg/cloudevents/sink_publisher.go
```diff
@@ -41,22 +42,36 @@ type SinkCloudEventPublisher struct {
...
+const (
+	operatorConfigMapName = "kubearchive-operator-config"
+	useKnativeConfigKey   = "use-knative"
+)
+
+func NewSinkCloudEventPublisher(source string) (*SinkCloudEventPublisher, error) {
+	scep := &SinkCloudEventPublisher{source: source}
 
 	var err error
 	if scep.httpClient, err = otelObs.NewClientHTTP([]cehttp.Option{}, []client.Option{}); err != nil {
```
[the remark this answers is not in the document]
ggallen: @skoved, I think we agreed at the last meeting that the optional version of this change is not something we want to pursue.  I'm going to close this PR.

If your comments are relevant in #1360, please add them there.

PR #1357, opened 2025-09-24
Add name with wildcard filter in API

### LINKED ISSUE #1194: Add a filter for text matching on the name resource field
### Context

For large amount of resources one of the potential issues is to provide the user a good way to filter them for what they want.
One of the ways of doing this is providing a filter for returning the resources based on matching the name with wildcards.
This is a feature available in some Kubernetes clientes (like Openshift console). 

In this case loading all the data in-memory in the frontend to be able to filter it could be too much considering that we are an archival tool, so it makes sense to consider adding this filter in the backend, the KubeArchive API.

Nevertheless, this could be dangerous and impact on performance, so we should investigate which is the best way to approach this feature and if we want to provide it.

### Acceptance Criteria

* Explore the text match of the name resource field (case insensitive) filter.
* Consider applying this filter ONLY if the resources are less than a specific number and retrieve an error otherwise.

maruiz93:
Assisted-by: Cursor AI

Resolves #1194 

## Release Note

```release-note
Add name filter with wildcards for resource list
```

## Notes for Reviewers

We need to check if we should provide this feature along with a change in the schema for adding an index on the name field

### DIFF
--- a/integrations/database/postgresql/migrations/02_name_idx.down.sql
+++ b/integrations/database/postgresql/migrations/02_name_idx.down.sql
@@ -1,0 +1,1 @@
+DROP INDEX name_idx;
--- a/pkg/database/sql/reader.go
+++ b/pkg/database/sql/reader.go
@@ -30,9 +30,19 @@ func (db *sqlDatabaseImpl) QueryResources(ctx context.Context, kind, apiVersion,
 		sb.Where(db.filter.NamespaceFilter(sb.Cond, namespace))
 	}
 	mainWhereClause := sqlbuilder.CopyWhereClause(sb.WhereClause)
+
+	isWildcardQuery := false
 	if name != "" {
-		sb.Where(db.filter.NameFilter(sb.Cond, name))
-	} else {
+		if strings.Contains(name, "*") {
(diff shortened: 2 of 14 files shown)


pkg/database/sql/reader.go
```diff
@@ -20,6 +20,23 @@ import (
...
+	}
+	// Convert wildcard pattern to SQL LIKE pattern and make it case-insensitive
+	sqlPattern := strings.ReplaceAll(pattern, "*", "%")
+	return strings.ToLower(sqlPattern)
+}
+
+// hasWildcards checks if a name pattern contains wildcard characters
+func hasWildcards(pattern string) bool {
+	return strings.Contains(pattern, "*")
+}
```
rh-hemartin, line 38: Probably `strings.Contains(pattern, "*")` was not that hard to read to justify a function.

integrations/database/postgresql/migrations/02_name_idx.down.sql
```diff
@@ -0,0 +1 @@
+DROP INDEX name_idx;
```
rh-hemartin, line 1: Should we remove the extension as well?

PR #1371, opened 2025-10-02
Do not always populate the 'continue' token

### LINKED ISSUE #1213: The API is retrieving a `continueID` for pagination even if there are no more pages
We should follow the same schema as Kubernetes API and not return a continue token if there are no more pages.

rh-hemartin:

Resolves #1213 

## Release Note

```release-note
Fix an error where the `continue` token was always populated.
```

### DIFF
--- a/cmd/api/routers/routers.go
+++ b/cmd/api/routers/routers.go
@@ -104,9 +104,12 @@ func (c *Controller) GetResources(context *gin.Context) {
 
 	// We send namespace even if it's an empty string (non-namespaced resources) the Database
 	// knows what to do
-	resources, lastId, lastDate, err := c.Database.QueryResources(
+	// We send limit+1 because we want to know if there are more resources than requested
+	// later we just returned what we were asked to return
+	newLimit := limit + 1
+	resources, err := c.Database.QueryResources(
 		context.Request.Context(), kind, apiVersion, namespace, name, id, date, labelFilters,
-		creationTimestampAfter, creationTimestampBefore, limit)
+		creationTimestampAfter, creationTimestampBefore, newLimit)
 
 	if err != nil {
 		abort.Abort(context, err, http.StatusInternalServerError)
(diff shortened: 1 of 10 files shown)


cmd/api/routers/routers.go
```diff
@@ -109,7 +109,15 @@ func (c *Controller) GetResources(context *gin.Context) {
 		return
 	}
 
-	continueToken := pagination.CreateToken(lastId, lastDate)
+	// We can't know if there are more resources in the DB than the limit, because the number
+	// of resources that we know about is limited by the limit value in the first place.
+	// So we do know that if the number of resources is equal to the limit, it is possible that
+	// there are more resources stored. With some usage this would be true most of the time
+	continueToken := ""
+	if len(resources) == limit {
```
maruiz93, line 117: I would say that it's better to do a `limit+1` and drop the last result instead of giving the user the idea that there are more resources when they aren't. This is likely to happen not too many times but I think it's better to be accurate in here, specially if anyone want to rely on the `continueToken` field for an automation.

The limit+1 approach is far more efficient than doing a separate `count()`.

What do you think?

maruiz93 (changes requested):
I'd like to change how we do things so we actually know if there are more elements in the database or not

cmd/api/routers/routers.go
rh-hemartin: I understand that you are proposing to run the DB query with an increased limit value than the provided by the user, then to drop the resources the user didn't ask for so we can check if there are more resources than the user asked for.

If that is right, I think it is worth a try.

maruiz93: yes, that's what I mean :)

rh-hemartin: Marta, give it a look and see what do you think. I still need to do tests, so give me feedback on the implementation.

PR #1418, opened 2025-10-20
Check for valid duration strings in CEL expressions.

### LINKED ISSUE #1416: Sink should check for runtime errors on CEL expressions
The documentation for [Program.Eval](https://pkg.go.dev/github.com/google/cel-go/cel#Program) (we use Program.ContextEval but returns the same values) in the cel library explains possible outcomes of evaluating a CEL expression

```
// *  `val`, `details`, `nil` - Successful evaluation of a non-error result.
// *  `val`, `details`, `err` - Successful evaluation to an error result.
// *  `nil`, `details`, `err` - Unsuccessful evaluation.
```

`ExecuteBooleanCEL` should check if the error returned is not nil and should log the error. This makes it easier to catch runtime errors from CEL expression like providing an invalid duration string to the duration macro. (ex: `duration("5d")`)

A non-nil error returned from `Program.ContextEval` indicated that their is either an issue with the syntax of the CEL expression that cannot be checked until execution or there is an edge case that the author of the CEL expression did not consider

ggallen:
Check for invalid duration strings when validating KubeArchive config files.
Sample `KubeArchiveConfig`:
```
---
apiVersion: kubearchive.org/v1
kind: KubeArchiveConfig
metadata:
  name: kubearchive
spec:
  resources:
    - archiveWhen: status.state != 'Completed'
      deleteWhen: timestamp(status.completionTime) < now() - duration("5xxx")
      selector:
        apiVersion: batch/v1
        kind: Job
```
Webhook validation output:
```
[gallen@gallen-thinkpadp1gen5 kubearchive] (issue-1404)$ kc apply -f kaconfig-1.yml
Error from server (Forbidden): error when creating "kaconfig-1.yml": admission webhook "vkubearchiveconfig.kb.io" denied the request: invalid duration string 'duration("5xxx")'
[gallen@gallen-thinkpadp1gen5 kubearchive] (issue-1404)$
```

Resolves #1416 

## Release Note

```release-note
NONE
```

### DIFF
--- a/cmd/operator/api/v1/kubearchiveconfig_webhook.go
+++ b/cmd/operator/api/v1/kubearchiveconfig_webhook.go
@@ -7,8 +7,10 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"regexp"
 
 	"github.com/kubearchive/kubearchive/pkg/cel"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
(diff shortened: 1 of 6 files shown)


cmd/operator/api/v1/kubearchiveconfig_webhook.go
```diff
@@ -90,25 +92,57 @@ func (kaccv *KubeArchiveConfigCustomValidator) validateKAC(kac *KubeArchiveConfi
...
+
+func validateDurationString(expr string) []error {
+	emptyObj := unstructured.Unstructured{
+		Object: map[string]interface{}{},
+	}
+	var BadConversion = errors.New("type conversion error from 'string' to 'google.protobuf.Duration'")
+
+	re := regexp.MustCompile(`(duration *\([^)]+\))`)
+	matches := re.FindAllString(expr, -1)
+	kaclog.Info("GREG", "expr", expr, "matches", matches)
```
skoved: I think you forgot to clean up this log line

ggallen:
> Could you add a test in which we reject the creation invalid strings?

Done.

rh-hemartin (approved):

PR #1419, opened 2025-10-21
Add discovery client to cli

### LINKED ISSUE #1002: The apiversion parameters in the "kubectl archive get" command should be the same as the "kubectl get" command
## Context

`kubectl get` accepts version and group parameters in the following format:

```
kubectl get [resource in singular or plural].[version].[group]
```

And the version and group are optional, taking the deployed apiversion by default.

The get command of the `kubectl-archive` CLI should be working the same way.

```bash
kubectl get deployments.v1.apps #valid
kubectl get deployment.apps #valid
kubectl get deployments.v1 #invalid
kubectl get deploy #valid
```

NOTE: The core resources don't admit the version or group

## Acceptance Criteria

* kubectl archive get validates and get the apiversion information the same way as the `kubectl get` command does.
* Use the libraries used in https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/get/get.go for this matter.

maruiz93:
This allows following the
same syntax of kubectl

Assisted-by: Cursor AI

Resolves #1002 

## Release Note

```release-note
Adopted a new syntax for specifying resource types and names in the CLI to be consistent with kubectl's convention (e.g., `type/name`).
```

### DIFF
--- a/pkg/cmd/config/config.go
+++ b/pkg/cmd/config/config.go
@@ -8,8 +8,11 @@ import (
 	"net/http"
 	"os"
 	"strconv"
+	"strings"
 
 	"github.com/spf13/pflag"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/client-go/rest"
 )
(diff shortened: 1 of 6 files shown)


rh-hemartin (approved):
There are some docs you may need to change as we have some CLI calls in there

skoved (approved):

PR #1420, opened 2025-10-21
Added gzip compression in API

### LINKED ISSUE #229: KubeArchive API Server allows compression
Kubernetes API allows the retrieval of the results for GET [compressed](https://kubernetes.io/docs/reference/using-api/api-concepts/#response-compression) in a gzip for reducing network bandwith.

KubeArchive API Server should also implement a similar thing as the KubeArchive database will likely keep large amount of data.

maruiz93:
Assisted-by: Cursor AI

Resolves #229 

## Release Note

```release-note
Supported gzip compression in KubeArchive API
```

## Notes for Reviewers

When adding the new dependency I upgraded gin version too to [1.11](https://gin-gonic.com/es/blog/news/gin-1-11-0-release-announcement/). Among other changes they included a fix that prevents me checking the full middleware
chain like we were doing before.
That's the reason of changing the tests.


maruiz93:
Instead of unit tests, I finally thought that e2e tests made more sense, because in the routers we should only check the logic of the different routers functions. I've added there, ready for review while checking linting issues

rh-hemartin (approved):

ggallen (approved):

test/integration/integration_test.go
```diff
@@ -64,13 +66,16 @@ func TestNormalOperation(t *testing.T) {
...
+
+	// Retrieve the objects from the DB using the API.
 	retryErr := retry.Do(func() error {
 		list, getUrlErr := test.GetUrl(t, token.Status.Token, url, map[string][]string{})
 		if getUrlErr != nil {
 			return getUrlErr
 		}
 
 		if len(list.Items) == 1 {
+			t.Log("✓ Successfully retrieved Job from the API")
```
mafh314: silly question: the "check symbol" can not create a format problem, right?

maruiz93: it's ascii so it shouldn't, but I can remove it, cursor started with it and I found them nice

you can check them here: https://github.com/kubearchive/kubearchive/actions/runs/18713305693

mafh314: ok! good for me them :) just checking hehe

mafh314 (approved):

PR #1428, opened 2025-10-22
reduce default page size from 100 to 30

### LINKED ISSUE #1427: KubeArchive API should use a smaller default page size
If a client makes a request to the API without specifying a page size, the page size will be 200. If there are 200 or more items, then 200 will be returned which can be really slow (500ms or longer). KubeArchive should use a smaller page size to improve latency

skoved:

Resolves #1427

## Release Note

```release-note
reduce default api page size to 30
```

## Notes for Reviewers

This should improve perfomance when a client does not specify a page size and there are 100 or more large resources to return.

### DIFF
--- a/cmd/api/pagination/pagination.go
+++ b/cmd/api/pagination/pagination.go
@@ -22,7 +22,7 @@ const (
 	continueKey     = "continue"
 	continueDateKey = "continueDate"
 	continueIdKey   = "continueId"
-	defaultLimit    = "100"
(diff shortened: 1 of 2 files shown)


ggallen:
Let's *not* push this directly to production. I committed [PR 1415](https://github.com/kubearchive/kubearchive/pull/1415) and would prefer that *not* go along with this fix, but rather be added on a branch and pushed to production that way.

PR #1447, opened 2025-10-29
Change watch timeout to 240 seconds.

### LINKED ISSUE #1446: Decrease watch timeout value
The watch timeout it currently set to be between 5-10 minutes. This may be causing issues with missed events when the /etc cache is cleared every 5 minutes.  Lowering it to 240 seconds (4 minutes) to see if that helps.

ggallen:
Decrease the watch timeout to see if helps with missing events.

Resolves #1446 

## Release Note

```release-note
NONE
```


rh-hemartin:
I think we can close this one.

PR #1460, opened 2025-11-04
Add more performant index

### LINKED ISSUE #1459: Add more performant index
When the database increase in size and there are too many resources of a type under a namespace, filter by name starts not to be performant, so we need to add the `name` param to the btree index.

maruiz93:

Resolves #1459  

## Release Note

```release-note
Increased performance for resource search by name
```

## Notes for Reviewers

The index can be created CONCURRENTLY to avoid locking the table but the migration process inherently creates a transaction for each version so it can be rolledback.

Info about CONCURRENTLY: https://www.bytebase.com/blog/postgres-create-index-concurrently/

### DIFF
--- a/integrations/database/postgresql/migrations/04_btree_name_idx.down.sql
+++ b/integrations/database/postgresql/migrations/04_btree_name_idx.down.sql
@@ -1,0 +1,8 @@
+CREATE INDEX IF NOT EXISTS resource_kind_namespace_idx ON public.resource
+    USING btree (kind, api_version, namespace);
+
+DROP INDEX IF EXISTS resource_kind_namespace_name_idx;
+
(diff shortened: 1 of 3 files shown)


rh-hemartin:
You need to change the version number hardcoded somwhere, this is why the integration tests should be failing.

maruiz93:
> You need to change the version number hardcoded somwhere, this is why the integration tests should be failing.

done

maruiz93:
The PR works, but the new index won't be used if there is a GIST name_idx.

We need to make sure of how would we proceed regarding to this:

1. Remove the name_idx GIST?
2. Find a way to define the name_idx so the combined btree for kind+api_version+namespace+name has more priority?

Issue #1456, opened 2025-11-03
Sink has increased 500 http response rate with cloud event changes in the operator


skoved:
After changes to make the kubearchive operator send cloud events instead of ApiServerSource, we are seeing an increase of http 500 responses and missing urls for pod logs

PR #1461, opened 2025-11-04
add index for uuid in the log_url table

skoved:

Related to #1456  

## Release Note

```release-note
NONE
```

## Notes for Reviewers

uuid is the value that the sink uses for looking up log_urls to delete. adding this index should speed up these delete operations.

### DIFF
--- a/integrations/database/postgresql/migrations/03_uuid_idx_log_url.down.sql
+++ b/integrations/database/postgresql/migrations/03_uuid_idx_log_url.down.sql
@@ -1,0 +1,2 @@
+DROP INDEX IF EXISTS log_url_uuid_idx;
+CREATE INDEX IF NOT EXISTS log_url_uuid_idx ON resource USING btree (uuid);
--- a/integrations/database/postgresql/migrations/03_uuid_idx_log_url.up.sql
+++ b/integrations/database/postgresql/migrations/03_uuid_idx_log_url.up.sql
@@ -1,0 +1,2 @@
+DROP INDEX IF EXISTS log_url_uuid_idx;
+CREATE INDEX IF NOT EXISTS log_url_uuid_idx ON log_url USING btree (uuid);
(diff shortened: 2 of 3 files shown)


integrations/database/postgresql/migrations/03_uuid_idx_log_url.down.sql
```diff
@@ -0,0 +1,2 @@
+DROP INDEX log_url_uuid_idx;
+CREATE INDEX log_url_uuid_idx ON resource USING btree (uuid);
```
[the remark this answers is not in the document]
ggallen: Never mind. Leave it.

ggallen (approved):

integrations/database/postgresql/migrations/03_uuid_idx_log_url.down.sql
maruiz93: what about 
```suggestion
CREATE INDEX IF NOT EXISTS log_url_uuid_idx ON resource USING btree (uuid);
```

That way this allows us or other to have this index in place before the migration and don't make it crash.

We can do the same with the rest:

```
DROP INDEX IF EXISTS log_url_uuid_idx;
```

integrations/database/postgresql/migrations/03_uuid_idx_log_url.up.sql
```diff
@@ -0,0 +1,2 @@
+DROP INDEX log_url_uuid_idx;
+CREATE INDEX log_url_uuid_idx ON log_url USING btree (uuid);
```
maruiz93:
```suggestion
DROP INDEX IF EXISTS log_url_uuid_idx;
CREATE INDEX IF NOT EXISTS log_url_uuid_idx ON log_url USING btree (uuid);
```

integrations/database/postgresql/migrations/03_uuid_idx_log_url.down.sql
skoved: done

PR #1509, opened 2025-11-24
Add config command in CLI

### LINKED ISSUE #1445: Improve CLI configuration
The CLI configuration can be improved. 
It's currently configured through environment variables and the host by default is usually for no use, the user needs to explicit it and it's not easy to know how.

We could include a config file, something similar to `kubeconfig`, it could be by default under `~/.config/kubectlkaconfig` and the path could be overriden through an env var.

When running a `kubectl ka` command, the file should be checked and if it's not there, a prompt should ask the user if they want to proceed with some configuration.

If yes:

1. Create the file
2. Check if the user is currently connected to a kubernetes cluster
2.1 If not, ask the user to connect to a cluster
2.2 If yes, ask the user if they want to proceed configuring kubearchive against <cluster-name>
3. Look for probable namespaces where kubearchive could be deployed: `kubearchive`, or a namespace that contains that string and look for the api-server service
4. See if we can infer the url (this needs a little more investigation on my part), openshift has routes, but plain kubernetes don't and we should see if we could guess the host, or check if the api is being port-forwarded is useful too.
5. If after trying several approaches this doesn't work show something like "The host couldn't be inferred. Please contact the admin and configure it manually with `kubectl ka config --set-host <host>`
6. If a host is found, reach the /livez endpoint to see if the connection is successful. If it's not, show the user the proper error, it could be the previous one or one of the following:
  6.1. The KubeArchive endpoint needs a certificate, do you have a cert to be used? y/n
          If yes, ask for the path, if no, go to 6.2
  6.2. Do you want to allow insecure connections to KubeArchive API? y/n
          If yes, set the insecure flag
7. If everything was ok, check if there is a token configured for the current context, if not, ask the user if it wants to configure a bearer token for the KubeArchive API requests.

This process is inspired in other cli's config like "oh-my-zsh" that guide you step by step.

The final config file would look like:

```yaml
clusters:
  prod-cluster:                    
    cluster_name: "prod-cluster"
    server_url: "https://api.prod-cluster.example.com:6443"
    host: "https://kubearchive-api.prod-cluster.example.com"
    tls_insecure: false
    cert_path: "/path/to/prod-ca.crt"
    token: "eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9..."
    
  staging:
    cluster_name: "staging"
    server_url: "https://api-staging-abc123.region.provider.com:6443"
    host: "https://kubearchive-api-staging.apps.staging.example.com"
```

This will also allow having multiple kubeArchive configs for different clusters.

This will have to be added with a CLI command `config` that allows setting different config parameters without the need of editing the file.

### Acceptance Criteria

* There is a new command `config` that allows setting a persistent config for a kubernetes cluster (setting host, insecure flag, cert path, token)
* There is a new config parameter for the config file. By default it will be under `~/.config/` directory
* KubeArchive will look for the config file on each `get`/`logs` command. If it's not there, start the setup process to guide the user, if it is, look for the current kubernetes cluster and see if it's configured in the file. If true, go with that config, if not, start the setup process.
* KubeArchive will have a new command `setup` to start the config process to guide the user even if the file and the cluster has already been configured.
* The different errors that can appear in the normal working mode, need to be parsed to check if it's a config problem and ask the user if they want to start the setup process.
* The `config` command allows showing the complete config or the config for a specific cluster (or the current cluster)
* The `config` command allows deleting a config for a cluster (or the current cluster)
* Env variables will be kept so if they are configured they will be used instead of the file config
* KubeArchive config will be persisted for different clusters preventing the user the need to do repeated config

maruiz93:
Assisted-by: Cursor AI

Resolves #1445 

## Release Note

```release-note
Add `config` subcommand to KubeArchive CLI
```

## Notes for Reviewers

Pending some unit tests and docs:
- [x] User documentation
- [x] Unit tests for the subcommands


ggallen (commented):
I haven't looked too closely but I have some questions and comments that maybe we want to talk about.  Sorry this is so late in the process.

I think we should piggyback on `kubectl config`.  You should never set a host, that is all set in `kubectl config`. The context names in the KA config file should match context names in the kubeconfig file.

I don't think you should be able to change to a KubeArchive config context, you should only do that by using kubectl to change the context.

maruiz93:
> I haven't looked too closely but I have some questions and comments that maybe we want to talk about. Sorry this is so late in the process.
> 
> I think we should piggyback on `kubectl config`. You should never set a host, that is all set in `kubectl config`. The context names in the KA config file should match context names in the kubeconfig file.
> 
> I don't think you should be able to change to a KubeArchive config context, you should only do that by using kubectl to change the context.


Note that we currently (I mean before this PR) need to set the host where kubearchive API is accessible, localhost if it's port-forwarded, an openshift route, or whatever host is created by any ingress controller used.

> You should never set a host, that is all set in `kubectl config`

`kubectl config` has a host for kubernetes cluster (and we use that) but it doesn't have a way to store the kubearchive endpoint.

> I don't think you should be able to change to a KubeArchive config context

You can't use kubearchive to change the context, the context is only managed by `kubectl`, think of this `kubectl-ka` config file as an extension of the information shown in the `kubeconfig`, it doesn't store any context at all.


So, the logic is as follows:
1. Check kubernetes builtin config to see what cluster am I connected to
2. See if there is any kubearchive config (the host for the kubearchive API calls, the certs, if it should be done securely or insecure, and the token for the cases that your user is authenticated with certs instead with bearer tokens). 

What I'm providing here is just some tools to make this more automatic, but this is already there with env vars and flags, see: https://kubearchive.github.io/kubearchive/main/cli/configuration.html#_global_configuration_parameters

ggallen:
> > I haven't looked too closely but I have some questions and comments that maybe we want to talk about. Sorry this is so late in the process.
> > I think we should piggyback on `kubectl config`. You should never set a host, that is all set in `kubectl config`. The context names in the KA config file should match context names in the kubeconfig file.
> > I don't think you should be able to change to a KubeArchive config context, you should only do that by using kubectl to change the context.
> 
> Note that we currently (I mean before this PR) need to set the host where kubearchive API is accessible, localhost if it's port-forwarded, an openshift route, or whatever host is created by any ingress controller used.
> 
> > You should never set a host, that is all set in `kubectl config`
> 
> `kubectl config` has a host for kubernetes cluster (and we use that) but it doesn't have a way to store the kubearchive endpoint.
> 
> > I don't think you should be able to change to a KubeArchive config context
> 
> You can't use kubearchive to change the context, the context is only managed by `kubectl`, think of this `kubectl-ka` config file as an extension of the information shown in the `kubeconfig`, it doesn't store any context at all.
> 
> So, the logic is as follows:
> 
> 1. Check kubernetes builtin config to see what cluster am I connected to
> 2. See if there is any kubearchive config (the host for the kubearchive API calls, the certs, if it should be done securely or insecure, and the token for the cases that your user is authenticated with certs instead with bearer tokens).
> 
> What I'm providing here is just some tools to make this more automatic, but this is already there with env vars and flags, see: https://kubearchive.github.io/kubearchive/main/cli/configuration.html#_global_configuration_parameters

@maruiz93, this all makes sense. Thanks for the explanation.

PR #1513, opened 2025-11-25
Fix --archived and --in-cluster CLI flags

### LINKED ISSUE #1512: --archived and --in-cluster flags don't work as expected
The flags currently only work if the values are explicitly provided:

`kubectl ka get po -n product-kubearchive --archived=true --in-cluster=false`

But that should be equivalent to run:

`kubectl ka get po -n product-kubearchive --archived`

but, instead of that, the flag is ignored. The same happens with `--in-cluster` flag.

### Acceptance Criteria

* When one of the flags is provided without an explicit value, only that kind of resources should be shown

maruiz93:

Resolves #1512  

## Release Note

```release-note
Fix `--in-cluster` and `--archived` flags used without explicit value
```

### DIFF
--- a/pkg/cmd/get.go
+++ b/pkg/cmd/get.go
@@ -15,6 +15,7 @@ import (
 	"time"
 
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/util/duration"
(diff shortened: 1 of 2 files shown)


mafh314:
I was writing about some tests in pkg/cmd/get_test.go that need a change, but that exact issue is raised by the PR checks

maruiz93:
> I was writing about some tests in pkg/cmd/get_test.go that need a change, but that exact issue is raised by the PR checks

yes, It's fixed now :)

rh-hemartin (approved):

PR #1517, opened 2025-11-26
Add --after --before and --limit flags to CLI

### LINKED ISSUE #1448: Use limit and creationTimestampAfter and creationTimestampBefore in the CLI
The CLI currently returns one page with the default size.

The CLI command `get` should accept a `--limit` option and use something like "100" by default.

This should also come with the support of `--after`  and `--before` to use the `creationTimestampAfter` and `creationTimestampBefore` filters.

maruiz93:
Assisted by: Cursor AI

Resolves #1448  

## Release Note

```release-note
Add `--limit`, `--before` and `--after` flags to the CLI
```

## Notes for Reviewers

To improve the performance of the code I'm sorting + applying limits + applying timestamp filters at the same time.
The sorting uses the binary tree lookup for performance.
The early apply of filters improve memory usage.

There is a corner case that isn't covered with this feature related to having two resources with the exact timestamp in different pages. That could have unexpected outcome.

### DIFF
--- a/pkg/cmd/get.go
+++ b/pkg/cmd/get.go
@@ -9,7 +9,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
-	"sort"
+	"slices"
 	"strings"
 	"text/tabwriter"
 	"time"
@@ -25,16 +25,20 @@ import (
 type GetOptions struct {
 	genericiooptions.IOStreams
 	KARetrieverCommand
-	AllNamespaces      bool
-	APIPath            string
-	ResourceInfo       *ResourceInfo
-	Name               string
-	LabelSelector      string
-	OutputFormat       *string
-	JSONYamlPrintFlags *genericclioptions.JSONYamlPrintFlags
-	IsValidOutput      bool
-	InCluster          bool
-	Archived           bool
+	AllNamespaces          bool
+	APIPath                string
+	ResourceInfo           *ResourceInfo
+	Name                   string
+	LabelSelector          string
+	OutputFormat           string
+	JSONYamlPrintFlags     *genericclioptions.JSONYamlPrintFlags
+	IsValidOutput          bool
+	InCluster              bool
+	Archived               bool
+	Limit                  int
+	After                  time.Time
+	Before                 time.Time
+	kubearchiveQueryParams url.Values
 }
 
 // ResourceWithAvailability tracks a resource and its availability in different APIs
@@ -44,10 +48,31 @@ type ResourceWithAvailability struct {
 	Archived  bool
 }
 
+// KubeArchiveResponse represents the response structure from KubeArchive API
+type KubeArchiveResponse struct {
+	Kind       string                      `json:"kind"`
+	APIVersion string                      `json:"apiVersion"`
+	Metadata   map[string]interface{}      `json:"metadata"`
+	Items      []unstructured.Unstructured `json:"items"`
+}
+
+// getContinueToken extracts the continue token from KubeArchive API response metadata
+func getContinueToken(bodyBytes []byte) string {
+	var response KubeArchiveResponse
+	if err := json.Unmarshal(bodyBytes, &response); err != nil {
+		return ""
+	}
+	if response.Metadata != nil {
+		if continueToken, ok := response.Metadata["continue"].(string); ok {
+			return continueToken
+		}
+	}
+	return ""
+}
+
 func NewGetOptions() *GetOptions {
-	outputFormat := ""
 	return &GetOptions{
-		OutputFormat:       &outputFormat,
+		OutputFormat:       "",
 		JSONYamlPrintFlags: genericclioptions.NewJSONYamlPrintFlags(),
 		IOStreams: genericiooptions.IOStreams{
 			In:     os.Stdin,
@@ -55,6 +80,7 @@ func NewGetOptions() *GetOptions {
 			ErrOut: os.Stderr,
 		},
 		KARetrieverCommand: NewKARetrieverOptions(),
+		Limit:              100, // Default limit as per API
 	}
 }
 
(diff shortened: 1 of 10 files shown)


pkg/cmd/get.go
```diff
@@ -142,9 +235,144 @@ func (o *GetOptions) Complete(args []string) error {
...
+		if processedIdx > o.Limit || insertIdx > o.Limit {
+			break
+		}
+	}
+
+	return result, k8sTrimmed, kubeArchiveTrimmed
+}
+
+// findInsertionIndex finds the correct index to insert a resource to maintain sorted order (newest first)
+func (o *GetOptions) findInsertionIndex(resources []*ResourceWithAvailability, newResource *unstructured.Unstructured) int {
```
ggallen: I think you have to handle nil here and make sure that nil items are always sorted to the end of the list.

Something similar is done for `keepLastWhen` [here](https://github.com/kubearchive/kubearchive/blob/main/cmd/vacuum/publisher.go#L289).

pkg/cmd/get.go
```diff
@@ -142,9 +235,144 @@ func (o *GetOptions) Complete(args []string) error {
...
+			result[idx] = resourceWithAvailability
+		} else {
+			result = append(result, resourceWithAvailability)
+		}
+
+		// Trim to limit to save memory
+		if len(result) > o.Limit {
+			result = result[:o.Limit]
+			k8sTrimmed = true
+		}
```
ggallen: This seems inefficient to keep appending a single item to the list and then trimming it. I think it makes more sense to just declare the slice as `min(len(resources), o.Limit)+1` and just trim before returning.

An example of this was implemented for `keepLastWhen` in [this method](https://github.com/kubearchive/kubearchive/blob/main/cmd/vacuum/publisher.go#L108).

pkg/cmd/get.go
```diff
@@ -142,9 +235,144 @@ func (o *GetOptions) Complete(args []string) error {
...
+
+		// Trim to limit to save memory and stop processing if we have enough
+		if len(result) > o.Limit {
+			result = result[:o.Limit]
+			if insertIdx < len(result) {
+				k8sTrimmed = true
+			} else {
+				kubeArchiveTrimmed = true
+			}
+		}
```
ggallen: This code pretty much duplicates what is done in `processK8sResourcesOptimized`.  Can they be combined into a method?

pkg/cmd/get.go
```diff
@@ -142,9 +235,144 @@ func (o *GetOptions) Complete(args []string) error {
...
+	creationTime := resource.GetCreationTimestamp().Time
+	if !o.After.IsZero() && creationTime.Before(o.After) {
+		return false
+	}
+
+	if !o.Before.IsZero() && creationTime.After(o.Before) {
+		return false
+	}
+
+	return true
```
ggallen: I think this can all be replaced by these two lines:
```
creationTime := resource.GetCreationTimestamp().Time
return creationTime.After(o.After) && creationTime.Before(o.before)
```
I believe this logic is correct (even with zero values), and I think the existing code is wrong in that it will return true for values that equal `o.Before` or `o.After`.

pkg/cmd/get.go
```diff
@@ -78,13 +106,43 @@ func NewGetCmd() *cobra.Command {
...
+	}
+
+	// Add KubeArchive-specific parameters
+	params.Set("limit", fmt.Sprintf("%d", o.Limit))
+	if !o.After.IsZero() {
+		params.Set("creationTimestampAfter", o.After.Format(time.RFC3339))
+	}
+	if !o.Before.IsZero() {
+		params.Set("creationTimestampBefore", o.Before.Format(time.RFC3339))
+	}
```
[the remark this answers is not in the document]
maruiz93: I'll check this more thoroughly, but my idea here was to:
- Allow the user specify only before
- Allow the user specify only after
- Allow the user to specify both before and after

> I don't see where a user is forced to specify both --after and --before, and I think you want to allow only one.

why not? that's specifying a timerange, what if I want to see all my resources created last month? or all my resources created during a specific day?

> Why isn't o.After allowed to be zero?

I will revisit these lines, but it is allowed to not be set, in fact, there is a pagination in place that you can use by using `--limit` and `--before` without `--after`

> Allowing o.Before to be zero doesn't make sense.

I think it makes sense, what if I want to see all the resources in the last two days? that way I would use a `--after` with the value of now-2days and `--before` would not be set

pkg/cmd/get.go
maruiz93: but this is a private function only called in one place and nil isn't ever going to be a possible value. Either way I will handle it

pkg/cmd/get.go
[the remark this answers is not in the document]
ggallen: > That of course means your query always has a time interval.

Actually, you could just check if `o.After` is zero or `o.Before` is the largest date and omit them. But you'd still get rid of all the other checks.

pkg/cmd/get.go
maruiz93: you are right, I've combined them in my last refactor

pkg/cmd/get.go
```diff
@@ -149,9 +215,110 @@ func (o *GetOptions) Complete(flags *pflag.FlagSet, args []string) error {
...
+
+	var k8sTrimmed, k9eTrimmed bool
+
+	for _, resource := range resources {
+
+		creationTime := resource.GetCreationTimestamp().Time
+		if creationTime.Before(o.After) || creationTime.After(o.Before) {
+			continue
+		}
+
```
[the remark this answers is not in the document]
maruiz93: uh! and this is a really nice suggestion, I wasn't taking full advantage of `BinarySearchFunc`

Change done :)

PR #1527, opened 2025-12-03
Add prunedFromEtcd filter

### LINKED ISSUE #230: Add a filter in the API Server endpoints for retrieving the archived data that has already been pruned from etcd
To provide a smooth integration with the CLI API Server should allow a parameter that allows the retrieval of only the resources that have been pruned from etcd.

olegbet:
Signed-off-by: obetsun <obetsun@redhat.com>
Assisted-by: Claude

rh-pre-commit.version: 2.3.2
rh-pre-commit.check-secrets: ENABLED

Resolves: #230

## Release Note

```release-note
This change introduces the API parameter prunedFromEtcd which filters out the resources which were deleted from etcd already and as a result has cluster_deleted_ts populated.
```


maruiz93 (changes requested):
This PR is querying the resources in kubernetes API. This is fundamentally not ok because that implies A LOT of overhead and waiting. The filter is for making things faster, not otherwise. Consider please adding just a filter in the postgresql piece that filter the resources that has a `cluster_deleted_ts` on it. This will be simpler, and the complexity is a lot less

maruiz93:
Also, this affects the API exposed to the user so you should include release notes in the PR description. Note that this PR description is meant to be used by the current release process for increasing the version. If all the PRs have no release notes the release process will fail

PR #1589, opened 2026-01-15
Change the error returned when no logs are found

### LINKED ISSUE #1450: Modify the not found message in the logging API integration
### Acceptance Criteria

Change the message in [logging.go](https://github.com/kubearchive/kubearchive/blob/main/cmd/api/logging/logging.go#L144-L144) when no logs are found for:

```
No logs found for container "x" in pod "y"
```

For that the context should have the Pod and the Container name and not only the URL.

rh-hemartin:

Resolves #1450 

## Release Note

```release-note
Change the error string when logs are not found to a more generic and friendly version.
```

### DIFF
--- a/cmd/api/logging/logging.go
+++ b/cmd/api/logging/logging.go
@@ -6,6 +6,7 @@ package logging
 import (
 	"bufio"
 	"crypto/tls"
+	"errors"
 	"fmt"
 	"io"
 	"log/slog"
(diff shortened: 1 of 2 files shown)


maruiz93 (commented):
> I'm not using the proposed string on the issue because the security recommendations says not to reflect input back to the user, so we can't print back the pod and container name they requested. So the error is very generic.

Can you point me to that security recommendation? If you reach this is because you have enough permissions to see which containers are inside a pod, so I don't understand this.

Also, the current url that we print gives this information, so we are going backwards with regards of information provided to the user

rh-hemartin:
> > I'm not using the proposed string on the issue because the security recommendations says not to reflect input back to the user, so we can't print back the pod and container name they requested. So the error is very generic.
> 
> Can you point me to that security recommendation? If you reach this is because you have enough permissions to see which containers are inside a pod, so I don't understand this.
> 
> Also, the current url that we print gives this information, so we are going backwards with regards of information provided to the user

I will send you the recommendation.

rh-hemartin:
> If you reach this is because you have enough permissions to see which containers are inside a pod, so I don't understand this.

That is true... Then I will add the names

PR #1742, opened 2026-02-16
downgrade go version to 1.23.9

### LINKED ISSUE #1500: Use go 1.23.9
renovate updated the go version in the github actions to `1.25.4` which has allowed it to update us to dependencies that are not compatible with `1.23.9`. Set the go version back to `1.23.9` and downgrade dependencies in `go.mod` accordingly`

skoved:

Resolves #1500 

## Release Note

```release-note
NONE
```

## Notes for Reviewers

renovate updated the version again... This happened because renovate upgraded the version of go used in CI which allowed a different PR to upgrade the go version. I add a new rule that I think might stop this from happening in the future but we will see.


rh-hemartin (approved):

beerparty (approved):
LGTM

PR #1777, opened 2026-03-02
Refactor the logging integration to support tailing

### LINKED ISSUE #1489: The log endpoint should return information about the owner of the logs
The log endpoint currently return the logs and we should provide more information about the owner of the logs: the container, the pod name...

### Acceptance Criteria

* If the logs are only from one container, output container and pod information in the response headers
* If there are logs from several containres (e.g. using labels), output the pod+container before the logs to distinguish one from another.
* The CLI should use this information to properly render the logs

### LINKED ISSUE #1560: Support tailing logs in logging integration
Currently the URL for fetching logs with all the query params is fixed to the container at archival time by the sink.
This makes the logging integration rigid as it doesn't support other ways of consuming the logging backend API for doing different things for getting the logs like for example, getting the last n lines.

Kubernetes API supports tailing logs with the `tailLines` query param: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.28/#read-log-pod-v1-core

### Tasks

* Create a design document for changing the architecture of kubearchive to be more flexible regarding how the URL and other parameters are stored in the database
* Create configmaps to provide the configuration of how to contact the backend logging system for different purpose (full logs / tailed logs)
* Modify the sink to insert the log URL data according to the new design
* Create a new migration for the database to migrate the log_url data acording to the changes defined in the design document
* Modify the logging middleware in the API service to adapt to the queries done to the backend based on the new information stored in the database combined with the information in the new configmaps

maruiz93:
Provide flexibility on the calls at read time.
Allow different type of calls like tailing.

Assisted by: Claude CLI

Resolves #1560 , resolves #1489, related to #1450

## Release Note

```release-note
Refactored the way logging integration is configured.
Added support to tailing logs.
```

## Notes for Reviewers

WIP:
- [x] In review of some late changes I did when merging the code with the streaming change
- [x] Need to add docs
- [x] Locally test the migrate_logs_urls.sh

I've changed the integration config of systems different than loki (elasticsearch + splunk) but I haven't tested those. This PR only is tested against loki (the integration tested in the CI)

This PR should be merged only after https://github.com/kubearchive/kubearchive/pull/1783
When deploying a version with this changes we should do the following:

1. Apply migration to schema version 5
2. Deploy the new kubearchive version
3. Apply the migration to schema version 6
4. Run the migrate_logs_urls.sh file

(Steps 3 and 4 can be interchanged)


integrations/database/postgresql/migrations/05_log_url_refactor.up.sql
```diff
@@ -0,0 +1,18 @@
...
+UPDATE public.log_url
+SET
+    query = substring(url from 'query=([^&]+)'),
+    "start" = substring(url from 'start=([^&]+)'),
+    "end" = substring(url from 'end=([^&]+)')
+WHERE url ILIKE '%loki%';
+
+-- Strip URL down to just protocol + host + port
+UPDATE public.log_url
+SET url = regexp_replace(url, '(https?://[^/]+).*', '\1');
```
rh-hemartin: If the system is not Loki we don't popuate `query`, `start` and `end` but we change `url` so we remove the possibility for the user to apply a manual migration as their old URLs don't exist anymore. That is a problem.

I'm not sure how to deal with it.

maruiz93: there is a good point in here, and it's the order of applying stuff.

Ideally we want:

1 - Apply the changes in the configmaps + sink + api (and make the code retrocompatible so it still works if the url doesn't fit what's configured
2 - Gradually migrate the old urls


I am thinking of moving out of this query and instead, provide a helper script that migrate the URL in batches to prevent locking writes the minimum possible.

PR #1783, opened 2026-03-04
Refactor schema migration process

### LINKED ISSUE #1782: Make the migration process more flexible
Now that we are including new features that imply substantial changes in the schema, we want to apply them with the minimum timedown when upgrading. This implies:

- The api and sink should keep working while applying the changes, not locking or locking the minimum possible of the tables
- Some schema changes need to be applied after the sink and the api have been updated (e.g., droping indexes, droping columns, droping tables...)

To address this we need to:

- Allow both sink and api to work with a range of databse schema versions (not only one)
- The version should be related with the databse engine, the migration could be different for postgresql and for mariadb
- The schema-migration job should accept the target schema version as a parameter, so we can run it twice, once with an "intermediate version" and the other with the final version.

This is going to be useful for addressing #1560 and #1678

maruiz93:
Update schema migration job to accept a target version.
Provide a range of schema versions compatible with the codebase.

Assisted by: Claude CLI

Resolves #1782 

## Release Note

```release-note
The schema migration job accepts a target version to do gradual upgrades.
```

### DIFF
--- a/pkg/database/database_test.go
+++ b/pkg/database/database_test.go
@@ -4,7 +4,7 @@
 package database
 
 import (
-	"fmt"
+	"sync"
 	"testing"
 
 	"github.com/kubearchive/kubearchive/pkg/database/fake"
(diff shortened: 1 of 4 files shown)


pkg/database/database_test.go
```diff
@@ -16,36 +16,44 @@ func TestNewDatabase(t *testing.T) {
...
 		{
 			name:          "current schema version",
-			schemaVersion: CurrentDatabaseSchemaVersion,
-			err:           nil,
+			schemaVersion: "4",
+		},
+		{
+			name:          "version above max",
+			schemaVersion: "5",
+			err:           "database schema version 5 is outside accepted range [4, 4]",
```
beerparty: Add a test with a non-numeric version for the schema (something like "v9"), so you can verify that the code handles it elegantly.

PR #1828, opened 2026-03-25
Add server-side resource count endpoint

FaisalAl-Rayes:
Add a `?count=true` query parameter to the API server and a `--count` flag to the kubectl-ka CLI. When used, the API executes SELECT COUNT(*) instead of fetching full resource data, returning {"count": N}.

This avoids transferring and deserializing all resource data when only the count is needed. All existing filters (labelSelector, timestamp range, namespace, kind) apply unchanged.

## Issue
https://github.com/kubearchive/kubearchive/issues/1829

## Release Note

```release-note
Adds a `count` query parameter to the API and a `--count` flag to the CLI that return the number of matching resources without fetching full resource data. All existing filters (label selectors, timestamp ranges, namespace) apply to count queries.
```

## Notes for Reviewers
This adds server-side counting to avoid the cost of fetching and deserializing full resource data just to get a count. The implementation reuses the
  existing filter pipeline (label selectors, timestamp ranges, namespace/kind scoping) but executes SELECT COUNT(*) instead of SELECT data.

  Key files to review:
  - `pkg/database/sql/reader.go` — `buildResourceCountQuery` and `CountResources` (core logic)
  - `cmd/api/routers/routers.go` — `?count=true` handling in `GetResources`
  - `pkg/cmd/get.go` — `--count` CLI flag and response parsing

  And the PR should be labeled kind/feature.

### DIFF
--- a/cmd/api/routers/routers.go
+++ b/cmd/api/routers/routers.go
@@ -4,7 +4,6 @@
 package routers
 
 import (
-	stdctx "context"
 	"errors"
 	"fmt"
 	"io"
@@ -113,10 +112,8 @@ func (c *Controller) GetResources(context *gin.Context) {
 			abort.Abort(context, errors.New("cannot use 'continue' with 'count'"), http.StatusBadRequest)
 			return
 		}
-		countCtx, cancel := stdctx.WithTimeout(context.Request.Context(), 30*time.Second)
-		defer cancel()
 		count, err := c.Database.CountResources(
-			countCtx, kind, apiVersion, namespace, name, labelFilters,
+			context.Request.Context(), kind, apiVersion, namespace, name, labelFilters,
 			creationTimestampAfter, creationTimestampBefore)
 		if err != nil {
 			abort.Abort(context, err, http.StatusInternalServerError)
--- a/cmd/api/routers/routers.go
+++ b/cmd/api/routers/routers.go
@@ -112,11 +112,11 @@ func (c *Controller) GetResources(context *gin.Context) {
 			abort.Abort(context, errors.New("cannot use 'continue' with 'count'"), http.StatusBadRequest)
 			return
 		}
-		count, countErr := c.Database.CountResources(
+		count, err := c.Database.CountResources(
 			context.Request.Context(), kind, apiVersion, namespace, name, labelFilters,
 			creationTimestampAfter, creationTimestampBefore)
-		if countErr != nil {
-			abort.Abort(context, countErr, http.StatusInternalServerError)
+		if err != nil {
+			abort.Abort(context, err, http.StatusInternalServerError)
 			return
 		}
 		context.JSON(http.StatusOK, gin.H{"count": count})
--- a/cmd/api/routers/routers.go
+++ b/cmd/api/routers/routers.go
@@ -4,6 +4,7 @@
 package routers
 
 import (
+	stdctx "context"
 	"errors"
 	"fmt"
 	"io"
@@ -112,8 +113,10 @@ func (c *Controller) GetResources(context *gin.Context) {
 			abort.Abort(context, errors.New("cannot use 'continue' with 'count'"), http.StatusBadRequest)
 			return
 		}
+		countCtx, cancel := stdctx.WithTimeout(context.Request.Context(), 30*time.Second)
+		defer cancel()
 		count, err := c.Database.CountResources(
-			context.Request.Context(), kind, apiVersion, namespace, name, labelFilters,
+			countCtx, kind, apiVersion, namespace, name, labelFilters,
 			creationTimestampAfter, creationTimestampBefore)
 		if err != nil {
 			abort.Abort(context, err, http.StatusInternalServerError)
--- a/cmd/api/routers/routers.go
+++ b/cmd/api/routers/routers.go
@@ -102,6 +102,27 @@ func (c *Controller) GetResources(context *gin.Context) {
 		apiVersion = fmt.Sprintf("%s/%s", group, version)
 	}
 
+	// Count-only request
+	if context.Query("count") == "true" {
+		if context.Query("limit") != "" {
+			abort.Abort(context, errors.New("cannot use 'limit' with 'count'"), http.StatusBadRequest)
+			return
+		}
+		if context.Query("continue") != "" {
+			abort.Abort(context, errors.New("cannot use 'continue' with 'count'"), http.StatusBadRequest)
+			return
+		}
+		count, countErr := c.Database.CountResources(
+			context.Request.Context(), kind, apiVersion, namespace, name, labelFilters,
+			creationTimestampAfter, creationTimestampBefore)
+		if countErr != nil {
+			abort.Abort(context, countErr, http.StatusInternalServerError)
+			return
+		}
+		context.JSON(http.StatusOK, gin.H{"count": count})
+		return
+	}
+
 	// Single resource by exact name - no streaming needed
 	if name != "" && !strings.Contains(name, "*") {
 		var resources []labelFilter.Resource
--- a/cmd/api/routers/routers.go
+++ b/cmd/api/routers/routers.go
@@ -112,11 +112,11 @@ func (c *Controller) GetResources(context *gin.Context) {
 			abort.Abort(context, errors.New("cannot use 'continue' with 'count'"), http.StatusBadRequest)
 			return
 		}
-		count, err := c.Database.CountResources(
+		count, countErr := c.Database.CountResources(
 			context.Request.Context(), kind, apiVersion, namespace, name, labelFilters,
 			creationTimestampAfter, creationTimestampBefore)
-		if err != nil {
-			abort.Abort(context, err, http.StatusInternalServerError)
+		if countErr != nil {
+			abort.Abort(context, countErr, http.StatusInternalServerError)
 			return
 		}
 		context.JSON(http.StatusOK, gin.H{"count": count})
--- a/pkg/cmd/get.go
+++ b/pkg/cmd/get.go
@@ -148,6 +148,20 @@ func (o *GetOptions) buildKubeArchiveQueryParams() url.Values {
 	return params
 }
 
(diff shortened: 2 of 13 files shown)


maruiz93 (commented):
Thank you for your contribution! 

I added some comments on the PR, and please be mindful to add the following:

- The proper docs under the docs directory for both API reference and CLI usage
- Some integration tests under the test/inegration directory

Also, please add a github issue describing the work being done here and link it to the PR

pkg/database/sql/reader.go
```diff
@@ -86,6 +86,47 @@ func (db *sqlDatabaseImpl) buildResourceListQuery(ctx context.Context, kind, api
...
+	if creationTimestampBefore != nil {
+		sb.Where(db.filter.CreationTimestampBeforeFilter(sb.Cond, *creationTimestampBefore))
+	}
+	if !labelFilters.IsEmpty() {
+		if err := db.filter.ApplyLabelFilters(ctx, db.db, sb, labelFilters); err != nil {
+			return nil, err
+		}
+	}
+	return sb, nil
+}
```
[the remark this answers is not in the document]
olegbet: Hi, @FaisalAl-Rayes can you please address this comment?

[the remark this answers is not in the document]
FaisalAl-Rayes: @olegbet I think i did address them in later commits.

@maruiz93 can you confirm or if there is something else please let me know?

pkg/cmd/get.go
```diff
@@ -370,6 +390,28 @@ func (o *GetOptions) Run() error {
 				strings.Contains(apiErr.Message, "authentication failed") {
 				return fmt.Errorf("KubeArchive authentication required: %s", apiErr.Message)
 			}
+		}
+
+		// Handle count requests as a distinct branch
+		if o.Count {
```
maruiz93, line 409: The count block is inserted between the `apiErr` authentication check (line 388) and the existing `apiErr != nil` fallback (line 414), splitting a single error-handling flow into two disjoint blocks. A reader must track that `apiErr` was already partially checked above, and future editors risk introducing bugs in the gap.

The 404-to-zero mapping also silently swallows errors that could indicate a real problem (e.g. namespace doesn't exist vs. zero resources).

Consider handling the count path as a fully self-contained branch earlier — right after the HTTP call returns, before falling into the list-resource flow:
```go
if o.Count {
    return o.handleCountResponse(bodyBytes, apiErr)
}
```

pkg/database/sql/reader.go
```diff
@@ -43,32 +43,30 @@ func (db *sqlDatabaseImpl) QueryResourceByUID(ctx context.Context, kind, apiVers
...
 
-// buildResourceListQuery builds the SQL select query for listing resources.
-// This is shared between QueryResources and StreamResources.
-func (db *sqlDatabaseImpl) buildResourceListQuery(ctx context.Context, kind, apiVersion, namespace, name,
-	continueId, continueDate string, labelFilters *models.LabelFilters,
-	creationTimestampAfter, creationTimestampBefore *time.Time, limit int) (*sqlbuilder.SelectBuilder, error) {
-	sb := db.selector.ResourceSelector()
+// applyResourceFilters applies common filters (kind, namespace, name, timestamps, labels)
+// to a select builder. Returns true if the query is a list/wildcard query (not an exact name lookup).
+func (db *sqlDatabaseImpl) applyResourceFilters(ctx context.Context, sb *sqlbuilder.SelectBuilder,
```
maruiz93, line 49: This function returns `isListQuery` to signal whether pagination/sorting should be applied. In `buildResourceCountQuery` (line 112), the return is discarded (`_, err := ...`).

When `isListQuery` is false (exact name, no wildcard), timestamp and label filters are **skipped** — so a `CountResources` call with an exact name and timestamp filters would silently ignore those filters. The router currently prevents this combination, but the database interface accepts it.

Consider either (a) always applying all filters regardless of the name match pattern, or (b) moving the list/exact determination to callers who actually need it. Option (a) is simpler and more defensively correct — there's no reason a COUNT query should skip filters the caller provided.

pkg/database/sql/facade/selector.go
```diff
@@ -19,6 +20,11 @@ type DBSelector interface {
 // with the default selectors with non-specific DBMS functions
 type PartialDBSelectorImpl struct{}
 
+func (PartialDBSelectorImpl) ResourceCountSelector() *sqlbuilder.SelectBuilder {
+	sb := sqlbuilder.NewSelectBuilder()
+	return sb.Select("COUNT(*)").From("resource")
```
maruiz93, line 25: nit: PostgreSQL `COUNT(*)` cannot use index-only scans due to MVCC — broad filters on large tables (all pods in a big namespace) must visit every matching row. This is still far better than fetching full JSON blobs, but could take seconds on very large tables.

Consider adding `context.WithTimeout` around the `CountResources` call (e.g. 30s) as a safeguard against slow counts tying up a connection indefinitely. Not a blocker.

pkg/database/fake/database_generated.go
```diff
@@ -238,6 +238,17 @@ func (f *fakeDatabase) queryFilteredResources(ctx context.Context, kind, version
 	return resources
 }
 
+func (f *fakeDatabase) CountResources(ctx context.Context, kind, version, namespace, name string,
+	_ *models.LabelFilters,
```
maruiz93, line 242: Label filters are discarded here (`_`), so unit tests using the fake can't verify that label-filtered counts work correctly. Neither the router tests nor the integration tests cover count with label selectors or timestamp filters, despite docs showing label-filtered counts as a primary use case.

Consider adding at least one router test combining `count=true` with a `labelSelector` or timestamp filter.

cmd/api/routers/routers.go
```diff
@@ -102,6 +102,27 @@ func (c *Controller) GetResources(context *gin.Context) {
...
+	if context.Query("count") == "true" {
+		if context.Query("limit") != "" {
+			abort.Abort(context, errors.New("cannot use 'limit' with 'count'"), http.StatusBadRequest)
+			return
+		}
+		if context.Query("continue") != "" {
+			abort.Abort(context, errors.New("cannot use 'continue' with 'count'"), http.StatusBadRequest)
+			return
+		}
+		count, countErr := c.Database.CountResources(
```
maruiz93, line 115: nit: `countErr` is non-standard — the rest of this file uses `err` consistently. Since this is in its own scoped block, there's no shadowing concern. Rename to `err` for consistency.

pkg/database/sql/reader.go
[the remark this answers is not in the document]
maruiz93: yep, fine by me

pkg/cmd/get.go
FaisalAl-Rayes: Done. Extracted into `handleCountResponse(bodyBytes, apiErr)` and dropped the 404-to-zero mapping — the API returns `200 {"count": 0}` for zero matches, so a 404 is a real error.

pkg/database/sql/reader.go
FaisalAl-Rayes: Done. Timestamp and label filters are now applied unconditionally — moved outside the `isListQuery` guard. The return value is still used by `buildResourceListQuery` for pagination/sorting/limit only.

pkg/database/fake/database_generated.go
FaisalAl-Rayes: Added a router test for count with label selector. The fake discards labels consistently across `CountResources`, `QueryResources`, and `StreamResources`, so the test verifies parsing and plumbing. Label filtering correctness is covered by integration tests.

cmd/api/routers/routers.go
FaisalAl-Rayes: Done. Renamed to `err`.

pkg/database/sql/facade/selector.go
FaisalAl-Rayes: Implemented — added a 30s `context.WithTimeout` around the `CountResources` call.

maruiz93: The hardcoded 30-second timeout may be too short for count queries with complex filters on large databases, and it's inconsistent with the rest of the router — `QueryResources` and `StreamResources` use `context.Request.Context()` directly with no handler-level timeout. Consider removing this timeout — configurable query timeouts will be addressed separately in #1928.

pkg/database/sql/reader.go
```diff
@@ -43,49 +43,90 @@ func (db *sqlDatabaseImpl) QueryResourceByUID(ctx context.Context, kind, apiVers
...
+	sb := db.selector.ResourceCountSelector()
+	_, err := db.applyResourceFilters(ctx, sb, kind, apiVersion, namespace, name,
+		labelFilters, creationTimestampAfter, creationTimestampBefore)
+	if err != nil {
+		return nil, err
+	}
+	return sb, nil
+}
+
+func (db *sqlDatabaseImpl) CountResources(ctx context.Context, kind, apiVersion, namespace, name string,
```
maruiz93, line 119: `reader_test.go` thoroughly tests `QueryResources`, `StreamResources`, etc. using sqlmock, but the new `CountResources` and `buildResourceCountQuery` have no corresponding sqlmock tests. The actual SQL generated (including `COUNT(*)` and WHERE clauses) is not verified for either PostgreSQL or MariaDB flavors. Consider adding sqlmock-based unit tests following the existing pattern.

pkg/database/sql/facade/selector.go
FaisalAl-Rayes: Reverted — timeout removed. Will be handled globally in #1928.

pkg/database/sql/reader.go
FaisalAl-Rayes: Added sqlmock tests for `CountResources`: basic count with/without namespace, timestamp filters (after, before, both), and zero-result cases. All run against both PostgreSQL and MariaDB.

maruiz93:
@FaisalAl-Rayes Can you address the small linter issues?

maruiz93 (approved):

PR #1877, opened 2026-05-04
Make 07_08 migration script resumable

### LINKED ISSUE #1875: Make 07_08_populate_label_tables.sh resumable
## Problem

The `07_08_populate_label_tables.sh` migration script processes the entire resource ID range from scratch every time it runs. On a production database with ~225M resource IDs, this takes 48+ hours at a batch size of 10,000.

If the migration job is killed mid-way (e.g. by ArgoCD resync, pod eviction, node drain, OOM), the script restarts from ID 1 and all progress is lost. This makes it effectively impossible to complete the migration on large databases without guaranteeing uninterrupted execution for multiple days.

## Proposal

Make the script resumable by tracking progress. When the script starts, it should check where the previous run left off and continue from there instead of starting from the beginning.

Possible approaches:

1. **Track progress in the database**: Store the last successfully processed resource ID in a metadata/progress table. On restart, query this table and resume from that point.
2. **Rely on idempotency + skip empty batches faster**: The script already uses `ON CONFLICT DO NOTHING`, so re-processing is safe but slow. A faster skip mechanism for already-processed ranges would help.
3. **Use the resource_label table itself**: Query `MAX(resource_id)` from `resource_label` to determine where to resume. This works because the script processes IDs sequentially.

Option 3 is the simplest and requires no schema changes — the data itself tracks progress.

## Context

- Production incident: migration job on stone-prod-p02 was killed after running for 2 days (~48% through 225M rows), losing all progress
- The script uses batch inserts with `ON CONFLICT DO NOTHING` so re-processing is safe, just slow

maruiz93:

Resolves https://github.com/kubearchive/kubearchive/issues/1875

## Release Note

```release-note
The 07_08 label population migration script is now resumable. If interrupted, it resumes from where it left off instead of restarting from the beginning.
```

## Notes for Reviewers

The batch loop in `07_08_populate_label_tables.sh` now checks `MAX(resource_id)` in `resource_label` before starting. If a previous run made partial progress, it resumes from that point instead of restarting from the beginning. `ON CONFLICT DO NOTHING` handles any overlap from re-scanning the tail end of the last batch.

### DIFF
--- a/integrations/database/postgresql/migrations/07_08_populate_label_tables.sh
+++ b/integrations/database/postgresql/migrations/07_08_populate_label_tables.sh
@@ -45,6 +45,16 @@ fi
 
 log "Resource ID range: ${MIN_ID} to ${MAX_ID}"
 
+# Check if a previous run made partial progress by looking at the highest
+# resource_id already recorded in resource_label.
+# shellcheck disable=SC2086
+RESUME_ID=$(psql ${PSQL_OPTS} -c "SELECT COALESCE(MAX(resource_id), 0) FROM resource_label;")
+
+if [ "${RESUME_ID}" -gt "${MIN_ID}" ]; then
+    log "Resuming from resource ID ${RESUME_ID} (previous progress detected)"
+    MIN_ID="${RESUME_ID}"
+fi
+
 # ============================================================================
 # Single-pass batched population of all label tables.
 #


rh-hemartin (approved):

olegbet (approved):
/lgtm

PR #1881, opened 2026-05-05
Make database schema version range configurable via env vars

### LINKED ISSUE #1874: Make database schema version compatibility range configurable via environment variables
## Problem

The database schema version compatibility range is hardcoded in `pkg/database/database.go`:

```go
var DatabaseSchemaVersions = map[string]SchemaVersionRange{
    "postgresql": {Min: 10, Max: 13},
    "mariadb":    {Min: 1, Max: 1},
}
```

This creates a problem during long-running migrations. In practice, many schema migrations are purely additive (new tables, new indexes, new columns) and do not affect how the previous binary version operates. For example:

- v1.20.0 hardcodes the compatible range as `[5, 6]`
- Migrations 7 through 10 only add new label normalization tables and triggers that v1.20.0 never interacts with
- v1.20.0 works perfectly fine with schema versions 5 through 10 — but the hardcoded check rejects anything above 6

During the v1.21.3 upgrade on a production cluster, the migration job needed 48+ hours to process ~225M rows. While the migration was running (DB at schema version 7), the api-server and sink pods crashed and could not restart because v1.20.0 rejected schema version 7 — even though the code is fully compatible with it.

## Why the source of truth cannot be in the code

The upper bound of the compatibility range is only knowable **after** the migrations are written, which is **after** the binary is released. v1.20.0 was released before migrations 7-10 existed. We cannot change an already-released binary to widen its range.

The compatibility range is a property of the **relationship between the binary and the migrations**, not of either one alone. The deployer is the one who knows both sides.

### Environment variables

Add two environment variables that override the hardcoded defaults for the **current database engine** (determined by `DATABASE_KIND`):

- `SCHEMA_MIN_VERSION` → overrides `versionRange.Min`
- `SCHEMA_MAX_VERSION` → overrides `versionRange.Max`
- If not set, fall back to the hardcoded defaults (no behavior change for existing deployments)

A single pair of env vars is sufficient — the code already knows which engine is in use via `DATABASE_KIND`, so per-engine env vars (e.g. `POSTGRES_SCHEMA_MIN_VERSION`) are unnecessary and would leak implementation details into the deployment config.

### Code change

In `pkg/database/database.go`, after reading `DatabaseSchemaVersions`, check for the env vars:

```go
if minStr := os.Getenv("SCHEMA_MIN_VERSION"); minStr != "" {
    if v, err := strconv.Atoi(minStr); err == nil {
        versionRange.Min = v
    }
}
if maxStr := os.Getenv("SCHEMA_MAX_VERSION"); maxStr != "" {
    if v, err := strconv.Atoi(maxStr); err == nil {
        versionRange.Max = v
    }
}
```

### ArgoCD wave-based deployment strategy

Using a ConfigMap with `disableNameSuffixHash: true` in an early sync wave enables zero-downtime upgrades during long migrations:

- **Wave -3**: ConfigMap `kubearchive-schema-range` updated with `SCHEMA_MAX_VERSION=10`
- **Wave -2**: ConfigMap `kubearchive-schema-version` with `MIGRATION_VERSION=13`
- **Wave -1**: Migration Job runs
- **Wave 0**: Deployments update to v1.21.3 images

The Deployment's pod template references `kubearchive-schema-range` by a fixed name (no hash suffix). When the ConfigMap is updated in wave -3, the Deployment spec doesn't change — no rollout is triggered. But if a pod crashes during the migration and the ReplicaSet recreates it, the new pod resolves the `configMapKeyRef` at creation time and picks up the updated `SCHEMA_MAX_VERSION=10`. The restarted pod accepts the in-progress schema version and starts successfully.

### Deployment example

For a rolling upgrade from v1.20.0 (schema 5) to v1.21.3 (schema 13), the kustomization ConfigMap:

```yaml
configMapGenerator:
  - name: kubearchive-schema-range
    options:
      disableNameSuffixHash: true
      annotations:
        argocd.argoproj.io/sync-wave: "-3"
    literals:
      - SCHEMA_MIN_VERSION=5
      - SCHEMA_MAX_VERSION=10
```

Deployments reference it via `configMapKeyRef`:

```yaml
env:
  - name: SCHEMA_MIN_VERSION
    valueFrom:
      configMapKeyRef:
        name: kubearchive-schema-range
        key: SCHEMA_MIN_VERSION
  - name: SCHEMA_MAX_VERSION
    valueFrom:
      configMapKeyRef:
        name: kubearchive-schema-range
        key: SCHEMA_MAX_VERSION
```

After the migration completes and Deployments move to v1.21.3, the ConfigMap is updated:

```yaml
literals:
  - SCHEMA_MIN_VERSION=10
  - SCHEMA_MAX_VERSION=13
```

The overlap at version 10 ensures a smooth transition.

## Limitation: already-released versions

The current release process (`hack/release.sh`) is linear from `main` — there is no support for release branches or patch releases. This means we cannot release a v1.20.1 with env var support after v1.21.x has already been released.

The first migration to benefit from this feature will be from whatever version ships the env var support onward. For earlier versions (like the current v1.20.0 → v1.21.3 migration), the workaround is:

1. Revert the upgrade (restore the previous version's images and `MIGRATION_VERSION`)
2. The migration job runs the down scripts to roll back the DB schema
3. If the down scripts are unavailable in the old image (they were introduced in the newer version), manually set the schema version back to one accepted by the old binary:
   ```sql
   UPDATE schema_migrations SET version = 6, dirty = false;
   ```
4. Plan the migration during a maintenance window with accepted downtime, or speed up the migration to minimize the window

## Context

- Production incident: api-server and sink were down for 41+ hours during v1.21.3 migration on stone-prod-p02
- Root cause: v1.20.0 hardcodes schema version range `[5, 6]` and rejects version 7+ even though it is fully compatible

maruiz93:

Resolves https://github.com/kubearchive/kubearchive/issues/1874

## Release Note

```release-note
The database schema version compatibility range can now be overridden via `SCHEMA_MIN_VERSION` and `SCHEMA_MAX_VERSION` environment variables. This allows deployers to widen the accepted range during long-running migrations, preventing api-server and sink pods from rejecting compatible in-progress schema versions.
```

## Notes for Reviewers

After looking up the hardcoded `DatabaseSchemaVersions` range, the code now checks for `SCHEMA_MIN_VERSION` and `SCHEMA_MAX_VERSION` env vars and overrides the min/max accordingly. Invalid values are silently ignored (falls back to hardcoded defaults). Each override is logged individually.

Four new test cases cover: overriding min, overriding max, narrowing the range (rejection), and invalid env var (fallback).

### DIFF
--- a/pkg/database/database.go
+++ b/pkg/database/database.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"log/slog"
 	"maps"
+	"os"
 	"strconv"
 	"sync"
 
@@ -100,6 +101,19 @@ func newDatabase() (interfaces.Database, error) {
 			err = fmt.Errorf("no schema version range defined for database type '%s'", dbType)
 			slog.Error("No schema version range defined", "type", dbType)
 			return
+		}
+
+		if minStr := os.Getenv("SCHEMA_MIN_VERSION"); minStr != "" {
+			if v, errParse := strconv.Atoi(minStr); errParse == nil {
+				versionRange.Min = v
+				slog.Info("Schema min version overridden by environment", "min", v)
+			}
+		}
+		if maxStr := os.Getenv("SCHEMA_MAX_VERSION"); maxStr != "" {
+			if v, errParse := strconv.Atoi(maxStr); errParse == nil {
+				versionRange.Max = v
+				slog.Info("Schema max version overridden by environment", "max", v)
+			}
 		}
 
 		dbVersionInt, errParse := strconv.Atoi(dbVersion)
--- a/pkg/database/database_test.go
+++ b/pkg/database/database_test.go
@@ -16,6 +16,7 @@ func TestNewDatabase(t *testing.T) {
 	tests := []struct {
 		name          string
 		schemaVersion string
+		envVars       map[string]string
 		err           string
 	}{
 		{
@@ -37,6 +38,27 @@ func TestNewDatabase(t *testing.T) {
 			schemaVersion: "v9",
 			err:           `invalid database schema version 'v9': expected an integer: strconv.Atoi: parsing "v9": invalid syntax`,
 		},
+		{
+			name:          "env var overrides min",
+			schemaVersion: "2",
+			envVars:       map[string]string{"SCHEMA_MIN_VERSION": "2"},
+		},
+		{
+			name:          "env var overrides max",
+			schemaVersion: "6",
+			envVars:       map[string]string{"SCHEMA_MAX_VERSION": "6"},
+		},
+		{
+			name:          "env var narrows range (reject)",
+			schemaVersion: "4",
+			envVars:       map[string]string{"SCHEMA_MAX_VERSION": "3"},
+			err:           "database schema version 4 is outside accepted range [4, 3]",
+		},
+		{
+			name:          "invalid env var ignored",
+			schemaVersion: "4",
+			envVars:       map[string]string{"SCHEMA_MIN_VERSION": "abc"},
+		},
 	}
 
 	for _, test := range tests {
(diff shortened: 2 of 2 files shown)


rh-hemartin (approved):

olegbet (approved):
/lgtm

PR #1895, opened 2026-05-12
Use namespaced query in connectivity test

### LINKED ISSUE #1894: config setup: use namespaced query for connectivity test
## Problem

During `kubectl ka config setup`, the authentication connectivity test queries `/api/v1/pods?limit=1` (cluster-scoped). This generates the following PostgreSQL query:

```sql
SELECT resource.data->'metadata'->>'creationTimestamp' AS created_at,
       resource.id, resource.uuid, resource.data
FROM resource
WHERE kind = 'Pod' AND api_version = 'v1'
ORDER BY resource.data->'metadata'->>'creationTimestamp' DESC, resource.id DESC
LIMIT 2
```

The existing composite index `idx_resource_kind_apiversion_ns_ts_id(kind, api_version, namespace, creationTimestamp DESC, id DESC)` cannot be used efficiently for this query because `namespace` (the 3rd column) is skipped. PostgreSQL must scan all Pod rows across every namespace and sort them to find the top 2. On large clusters this takes **~16 minutes**, causing the setup to time out and fall through to the ServiceAccount token prompt with a misleading message.

Meanwhile, namespaced queries (e.g., `/api/v1/namespaces/{ns}/pods?limit=1`) use the full index and return in milliseconds.

## Proposed Fix

In `getTokenForKubeArchive()` (`pkg/cmd/config/interactive.go`), change `TestKubeArchiveConnectivity` to use a namespaced request instead of a cluster-scoped one. The namespace is already available via `is.namespace`.

This works even if the namespace doesn't exist in the cluster or the user lacks RBAC access to it:
- A SubjectAccessReview evaluates policies, not namespace existence — the token is still validated.
- The connectivity test already treats 401 (authenticated but unauthorized) as success, only failing on "authentication failed" (invalid token).

## Context

- `/livez` and `/readyz` are unauthenticated, so they can't be used for token validation.
- Cluster-wide queries are not a real user use case — the original choice of `/api/v1/pods?limit=1` was made because Pods exist in every cluster and no namespace knowledge was needed at that point.
- Adding a database index for cluster-scoped queries is not justified since the current KubeArchive users don't run them.

maruiz93:

Resolves #1894

## Release Note

```release-note
Fixed `oc ka config setup` connectivity check timing out on large clusters by using a namespaced query instead of a cluster-scoped one.
```

## Summary

- Change `TestKubeArchiveConnectivity` to use `/api/v1/namespaces/{ns}/pods?limit=1` instead of `/api/v1/pods?limit=1`
- The cluster-scoped query caused ~16min response times on large clusters because it couldn't efficiently use the composite index (`kind, api_version, namespace, creationTimestamp, id`)
- All callers updated to pass a namespace, falling back to `"default"` when unavailable

## Test plan

- [x] New unit tests verify namespaced URL is constructed correctly
- [x] New unit test verifies empty namespace falls back to `"default"`
- [x] All existing tests pass (100+ tests in `pkg/cmd/...`)
- [x] Manual test: `oc ka config setup` on stone-stg-rh01 completes auth check instantly (was timing out at ~16min)

## Notes for Reviewers

The fix scopes the connectivity check to a single namespace instead of querying all pods cluster-wide, which allows PostgreSQL to use the composite index efficiently.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg/cmd/cmd_test.go
+++ b/pkg/cmd/cmd_test.go
@@ -29,7 +29,7 @@ type MockConnectivityTester struct {
 }
 
 // TestKubeArchiveConnectivity mocks the connectivity test
-func (m *MockConnectivityTester) TestKubeArchiveConnectivity(host string, tlsInsecure bool, token string, caData []byte) error {
+func (m *MockConnectivityTester) TestKubeArchiveConnectivity(host string, tlsInsecure bool, namespace string, token string, caData []byte) error {
 	if m.shouldFail {
 		if m.failError != nil {
 			return m.failError
--- a/pkg/cmd/config/connectivity.go
+++ b/pkg/cmd/config/connectivity.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 	"strings"
 
 	"k8s.io/client-go/rest"
@@ -14,7 +15,7 @@ import (
 
 // ConnectivityTester defines the interface for testing KubeArchive connectivity.
 type ConnectivityTester interface {
-	TestKubeArchiveConnectivity(host string, tlsInsecure bool, token string, caData []byte) error
+	TestKubeArchiveConnectivity(host string, tlsInsecure bool, namespace string, token string, caData []byte) error
 	TestKubeArchiveLivezEndpoint(host string, tlsInsecure bool, caData []byte) error
 }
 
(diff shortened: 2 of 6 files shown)


olegbet (approved):
/lgtm

mafh314 (approved):

PR #1904, opened 2026-05-21
Add temporary index on resource.updated_at for label backfill

### LINKED ISSUE #1903: Migration 10: add temporary index on resource.updated_at for label backfill performance
## Problem

The `10_label_backfill.up.sql` migration filters all four backfill queries on `r.updated_at >= now() - INTERVAL '1 days'`, but there is no index on `resource.updated_at`. This forces a sequential scan of the entire `resource` table for each statement, which is extremely slow on databases with a large number of rows.

## Proposed solution

Use the same shell script pattern established by `07_08_populate_label_tables.sh`:

1. **Add a `09_10_create_updated_at_index.sh` script** that runs outside golang-migrate (no transaction), creating the index concurrently:
   ```sql
   CREATE INDEX CONCURRENTLY IF NOT EXISTS idx_resource_updated_at ON resource (updated_at);
   ```
   This must be a shell script because `CREATE INDEX CONCURRENTLY` cannot run inside a transaction, and golang-migrate wraps multi-statement `.sql` files in a transaction.

2. **Modify `11_drop_gin_index.up.sql`** to also drop the temporary index:
   ```sql
   DROP INDEX IF EXISTS idx_resource_updated_at;
   ```

This avoids renumbering existing migrations or changing the software's compatibility version range.

## Affected files

- New: `integrations/database/postgresql/migrations/09_10_create_updated_at_index.sh`
- Modified: `integrations/database/postgresql/migrations/11_drop_gin_index.up.sql`

maruiz93:

Resolves #1903

## Release Note

```release-note
Improved performance of the label backfill migration (10) on large databases by creating a temporary index on resource.updated_at before the backfill runs.
```

## Notes for Reviewers

The label backfill migration (10) filters on `resource.updated_at` without an index, causing full sequential scans on large databases. This PR adds:

- `09_10_create_updated_at_index.sh`: creates a B-tree index concurrently before migration 10 runs, following the same shell script pattern as `07_08_populate_label_tables.sh`
- A `DROP INDEX` in `11_drop_gin_index.up.sql` to clean up the temporary index after the backfill

The script also handles the PostgreSQL edge case where a failed `CREATE INDEX CONCURRENTLY` leaves an invalid index behind.

### DIFF
--- a/integrations/database/postgresql/migrations/09_10_create_updated_at_index.sh
+++ b/integrations/database/postgresql/migrations/09_10_create_updated_at_index.sh
@@ -1,0 +1,44 @@
+#!/bin/bash
+# Copyright KubeArchive Authors
+# SPDX-License-Identifier: Apache-2.0
+#
+# Create a temporary index on resource.updated_at to speed up the label
+# backfill in migration 10. The index is dropped later in migration 11.
+#
+# This runs as a shell script (not a .sql migration) because
+# CREATE INDEX CONCURRENTLY cannot execute inside a transaction, and
+# golang-migrate wraps multi-statement .sql files in a transaction.
+#
+# Required environment variables:
+#   DATABASE_URL      - PostgreSQL host
+#   DATABASE_PORT     - PostgreSQL port
+#   DATABASE_DB       - Database name
+#   DATABASE_USER     - Database user
+#   DATABASE_PASSWORD - Database password
+
+set -euo pipefail
+
+log() { echo "$(date '+%Y-%m-%d %H:%M:%S') $*"; }
+
+PGPASSWORD="${DATABASE_PASSWORD}"; export PGPASSWORD
+
+PSQL_OPTS="-h ${DATABASE_URL} -p ${DATABASE_PORT} -U ${DATABASE_USER} -d ${DATABASE_DB}"
+
+log "Creating index on resource.updated_at..."
+
+# Drop any invalid index left by a previously failed CONCURRENTLY build,
+# otherwise IF NOT EXISTS would match the invalid index and skip creation.
+# shellcheck disable=SC2086
+psql ${PSQL_OPTS} -t -A -c "
+  SELECT 1 FROM pg_class c JOIN pg_index i ON c.oid = i.indexrelid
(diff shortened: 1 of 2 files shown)


integrations/database/postgresql/migrations/09_10_create_updated_at_index.sh
```diff
@@ -0,0 +1,44 @@
...
+
+PGPASSWORD="${DATABASE_PASSWORD}"; export PGPASSWORD
+
+PSQL_OPTS="-h ${DATABASE_URL} -p ${DATABASE_PORT} -U ${DATABASE_USER} -d ${DATABASE_DB}"
+
+log "Creating index on resource.updated_at..."
+
+# Drop any invalid index left by a previously failed CONCURRENTLY build,
+# otherwise IF NOT EXISTS would match the invalid index and skip creation.
+# shellcheck disable=SC2086
```
mafh314, line 31: Hola @maruiz93!
What do you think about this? I'm suggestibg this just because I think is easier to read. The functionality is the same.  

```
# shellcheck disable=SC2086
INVALID=$(psql ${PSQL_OPTS} -t -A -c "
  SELECT count(*) FROM pg_class c JOIN pg_index i ON c.oid = i.indexrelid
  WHERE c.relname = 'idx_resource_updated_at' AND NOT i.indisvalid;
")

if [ "${INVALID:-0}" -gt 0 ]; then
  log "Dropping invalid index idx_resource_updated_at from a previous failed run..."
  # shellcheck disable=SC2086
  psql ${PSQL_OPTS} -c "DROP INDEX IF EXISTS idx_resource_updated_at;"
fi
```

maruiz93: it definitely is! good feedback, I'll address it :)

maruiz93: Addressed — switched to count(*) with an explicit if block as suggested.

maruiz93: done!

mafh314 (approved):

PR #1915, opened 2026-05-28
#1857 log completeness verification

### LINKED ISSUE #1857: Add log completeness verification before returning logs
When logs are retrieved shortly after a resource is archived, the log forwarding pipeline (e.g., Vector → Loki) may not have finished ingesting all log lines. This results in truncated logs being returned to the consumer, with no way for the consumer to know the data is incomplete.

Today consumers work around this by adding a delay before querying, but this is fragile — a fixed delay may be insufficient for large logs or degraded pipeline conditions, and overly conservative for normal cases.

**Proposed solution**

Implement a stabilization check in KubeArchive's log retrieval path. Before returning full logs (i.e., when no `tailLines` query parameter is set), verify that the log forwarding pipeline has finished ingesting by comparing consecutive tail queries. This verification is **only applied in full retrieval mode** — tail queries from users requesting the last N lines are not affected. Additionally, provide a post-stream integrity signal via a response header.

**The algorithm:**

1. **Check if verification can be skipped (coarse):**
   - If the log record has an `END` timestamp and `now() > END`: **skip to step 6.**
   - If the log record has no `END` but has `START` and `now() > START + 24h`: **skip to step 6.**

2. Perform a tail query with `tailLines=1` (if `timestamp-json-path` is configured) or `tailLines=VERIFICATION_TAIL_LINES` (fallback, default 10).

3. **Check if verification can be skipped (fine-grained, only if `timestamp-json-path` is configured):**
   - Extract the timestamp of the last entry using the configured `timestamp-json-path` and `timestamp-format`.
   - If the timestamp is more than 5 minutes old: logs are stable, **skip to step 6.**

4. Wait for `VERIFICATION_INTERVAL` (configurable in the ConfigMap, default 5 seconds).

5. Repeat the tail query. Compare with the previous result:
   - **Timestamp path configured:** compare the last entry's timestamp.
   - **Fallback:** compare the content of the returned lines.
   - If unchanged, logs are considered stable — save the last line and proceed to step 6.
   - If changed, repeat from step 4 (up to a max number of retries).

6. Return the full logs.

7. **Post-stream integrity check:** If a verification was performed (steps 2-5), compare the last line of the streamed response with the last line saved during verification. If they differ, set a response header (e.g., `X-Logs-Incomplete: true`) to signal that new logs arrived during the full retrieval.

<details>
<summary><b>New configuration fields</b></summary>

All fields are optional and configured under the `full` endpoint's `verification` section in the `kubearchive-logging-reader` ConfigMap (`LOG_PROVIDERS`):

| Field | Description | Default |
|-------|-------------|---------|
| `verification.tail-lines` | Number of tail lines to compare when timestamp is not available | `10` |
| `verification.interval` | Wait time between verification tail queries | `5s` |
| `verification.max-retries` | Maximum number of retry cycles before giving up | `6` |
| `verification.timestamp-json-path` | JSONPath to extract the timestamp from the tail response | not set |
| `verification.timestamp-format` | Format of the extracted timestamp (see below) | not set |

</details>

<details>
<summary><b>Recommended approach for parsing <code>timestamp-format</code></b></summary>

Go's `time.Parse` only handles string layout patterns (the `"2006-01-02..."` reference time). It cannot parse numeric epoch timestamps. The recommended approach is to reserve a few keywords for epoch-based formats and fall back to `time.Parse` for everything else:

```go
func parseTimestamp(raw string, format string) (time.Time, error) {
    switch format {
    case "unix_nano":
        n, err := strconv.ParseInt(raw, 10, 64)
        if err != nil {
            return time.Time{}, err
        }
        return time.Unix(0, n), nil
    case "unix_ms":
        n, err := strconv.ParseInt(raw, 10, 64)
        if err != nil {
            return time.Time{}, err
        }
        return time.Unix(0, n*int64(time.Millisecond)), nil
    case "unix_sec":
        n, err := strconv.ParseInt(raw, 10, 64)
        if err != nil {
            return time.Time{}, err
        }
        return time.Unix(n, 0), nil
    default:
        return time.Parse(format, raw)
    }
}
```

| Backend | Example raw value | `timestamp-format` |
|---------|-------------------|--------------------|
| Loki | `"1746879420000000000"` | `unix_nano` |
| Splunk | `"2025-05-09T12:17:00.000+00:00"` | `2006-01-02T15:04:05.000-07:00` |
| Elasticsearch | `"2025-05-09T12:17:00.000Z"` | `2006-01-02T15:04:05.000Z` |
| Datadog | `"1746879420000"` | `unix_ms` |
| Sumo Logic | `"1746879420000"` | `unix_ms` |

</details>

<details>
<summary><b>Example Loki configuration</b></summary>

```yaml
http://loki-gateway.grafana-loki.svc.cluster.local:80:
  tail:
    reverse: true
    path: /loki/api/v1/query_range
    method: GET
    params:
      query: ${QUERY}
      start: ${START}
      end: ${END}
      limit: ${TAIL_LINES}
      direction: backward
    json-path: "$.data.result[*].values[*][1]"
  full:
    reverse: false
    path: /loki/api/v1/query_range
    method: GET
    timeout: 1m
    params:
      query: ${QUERY}
      start: ${START}
      end: ${END}
      limit: 300
      direction: forward
    json-path: "$.data.result[*].values[*][1]"
    pagination:
      next-cursor: "$.data.result[*].values[*][0]"
      param: start
      cursor-increment: 1
    verification:
      interval: 5s
      max-retries: 6
      timestamp-json-path: "$.data.result[*].values[*][0]"
      timestamp-format: "unix_nano"
```

</details>

<details>
<summary><b>Behavior summary</b></summary>

| Scenario | Cost |
|----------|------|
| `now() > END` (or `START + 24h`) | Zero overhead — no verification |
| Timestamp available and > 5 min old | One extra tail query, no wait |
| Timestamp available and recent | Tail queries + wait cycles until stable |
| No timestamp configured | Tail queries (10 lines) + wait cycles, content comparison |
| Logs changed during full retrieval | `X-Logs-Incomplete: true` header set |

</details>

<details>
<summary><b>Context</b></summary>

The log forwarding pipeline in the current Loki integration (Vector) has a worst-case batching delay of ~16 seconds (10s reduce transform + 5s sink batch + ingester flush). A 30-second fixed delay covers most cases, but large log volumes or pipeline retries can push this higher. This feature replaces fixed delays with a data-driven check.

Consumers (e.g., Konflux pipeline checks) currently have no signal that logs are complete and must guess with sleep-based workarounds. The `X-Logs-Incomplete` header gives them a reliable signal to retry if needed.

</details>

olegbet:
Assisted-by: Claude

Resolves: #1857

```release-note
Implement a stabilization check in KubeArchive's log retrieval path by checking the log timestamps.
```


cmd/api/logging/verification.go
```diff
@@ -0,0 +1,319 @@
...
+
+	var jsonPathParser jp.Expr
+	if tailEndpoint.JsonPath != "" {
+		jsonPathParser, err = jp.ParseString(tailEndpoint.JsonPath)
+		if err != nil {
+			return nil, time.Time{}, fmt.Errorf("invalid tail jsonPath: %w", err)
+		}
+	}
+
+	body, err := io.ReadAll(response.Body)
```
maruiz93, line 194: **High — Unbounded `io.ReadAll` on probe response (memory exhaustion)**

`io.ReadAll(response.Body)` has no size limit on the success path. A misbehaving backend could return an arbitrarily large response. The error path at line 182 correctly uses `io.LimitReader(..., 512)`, but the success path has no equivalent bound. With MaxRetries+1 probes under concurrent load, memory impact is multiplicative.

Suggestion:
```go
body, err := io.ReadAll(io.LimitReader(response.Body, 2<<20)) // 2 MiB
```

_Consensus: 3/3 agents_

cmd/api/logging/verification.go
```diff
@@ -0,0 +1,319 @@
...
+	}
+
+	interval := verification.intervalDuration()
+
+	for attempt := range verification.MaxRetries {
+		select {
+		case <-ctx.Done():
+			slog.Warn("Verification aborted: context cancelled")
+			return verificationResult{performed: true, stable: false}
+		case <-time.After(interval):
```
maruiz93, line 282: **Medium — `time.After` leak in retry loop**

`time.After(interval)` creates a new timer on every iteration. When `ctx.Done()` fires before the timer expires, the timer leaks until it naturally fires.

Suggestion:
```go
timer := time.NewTimer(interval)
select {
case <-ctx.Done():
    timer.Stop()
    return verificationResult{performed: true, stable: false}
case <-timer.C:
}
```

_Consensus: 2/2 agents_

cmd/api/logging/logging.go
```diff
@@ -286,7 +293,7 @@ func abortNoLogsFound(c *gin.Context, record *interfaces.LogRecord, requestURL s
 // It is used when tailLines is requested (to return only the last N lines) or when the
 // backend returns entries in reverse order (to re-order them). Entries are optionally
 // reversed, flattened into individual lines, and truncated to tailLinesInt.
-func writeBufferedLogs(c *gin.Context, reader *bufio.Reader, jsonPathParser jp.Expr, record *interfaces.LogRecord, requestURL string, mode string, reverse bool, tailLinesInt int) {
+func writeBufferedLogs(c *gin.Context, reader *bufio.Reader, jsonPathParser jp.Expr, record *interfaces.LogRecord, requestURL string, mode string, reverse bool, tailLinesInt int, lastLineOut *string) {
```
maruiz93, line 296: **Medium — `lastLineOut *string` out-parameter is a Go anti-pattern**

The `*string` out-parameter threads through `writeBufferedLogs` (line 296), `streamLogs` (line 363), and `streamPaginatedLogs` (line 673) — now up to 9 params each. This hides a return value in the parameter list and couples verification concerns into streaming functions that otherwise have no verification knowledge.

Consider returning the last line as a second return value, or wrapping streaming state in a struct where `lastLine` is a field.

_Consensus: 3/3 agents_

cmd/api/logging/logging.go
```diff
@@ -520,7 +557,14 @@ func LogRetrieval() gin.HandlerFunc {
 			} else if legacyFullURL != "" {
 				slog.Warn("pagination is configured but will be ignored for legacy full URLs", "mode", mode)
 			} else {
-				streamPaginatedLogs(c, &client, endpoint, record, extraVars, headers, jsonPathParser, mode)
+				var streamedLastLine string
+				streamPaginatedLogs(c, &client, endpoint, record, extraVars, headers, jsonPathParser, mode, &streamedLastLine)
+				if verResult.performed && verResult.stable && streamedLastLine != "" &&
```
maruiz93, line 562: **Medium — Duplicated post-stream integrity check**

This identical 4-line block (compare `streamedLastLine`, set trailer, log warning) also appears in the non-paginated path at line 629. Extract to a helper:
```go
func checkPostStreamIntegrity(c *gin.Context, verResult verificationResult, streamedLastLine string) { ... }
```

_Consensus: 3/3 agents_

cmd/api/logging/logging.go
```diff
@@ -511,6 +525,29 @@ func LogRetrieval() gin.HandlerFunc {
...
+				c.Request.Context(),
+				&client,
+				providerConfig.Tail,
+				record,
+				headers,
+				endpoint.Verification,
+				time.Now,
+			)
+
+			if verResult.performed {
```
maruiz93, line 543: **Medium — Dual header/trailer signaling confuses clients**

When unstable pre-stream, `X-Logs-Incomplete: true` is set as a regular header (line 545). When the post-stream check fails, it's set as a trailer (lines 564, 631). Clients must inspect both. Also, when stable, `Trailer: X-Logs-Incomplete` is announced here but no trailer value is ever set — a promised trailer that never arrives.

Consider standardizing on trailers exclusively, and always setting a definitive value (`true`/`false`).

_Consensus: 3/3 agents_

cmd/api/logging/verification.go
```diff
@@ -0,0 +1,319 @@
...
+	hasTimestampPath := verification.TimestampJsonPath != ""
+	tailLines := verification.TailLines
+	if hasTimestampPath {
+		tailLines = 1
+	}
+
+	prevLines, prevTimestamp, err := performTailProbe(ctx, client, tailEndpoint, record, headers, tailLines, verification)
+	if err != nil {
+		slog.Warn("Verification tail probe failed, skipping verification", "error", err)
+		return verificationResult{performed: false}
```
maruiz93, line 259: **Medium — Silent fail-open with no observability signal**

When the initial tail probe fails, verification returns `performed: false` with no header. A client cannot distinguish "verification passed" from "verification not attempted." A broken tail endpoint silently degrades the feature.

Consider a separate header value (e.g., `X-Logs-Verification: skipped`) and a counter metric for probe failures.

_Consensus: 2/3 agents_

cmd/api/logging/verification.go
```diff
@@ -0,0 +1,319 @@
...
+	}
+	if record.End == "" && record.Start != "" {
+		if startTime, err := tryParseTimestamp(record.Start); err == nil && now.After(startTime.Add(coarseSkipFallbackAge)) {
+			return true
+		}
+	}
+	return false
+}
+
+func slicesEqual(a, b []string) bool {
```
maruiz93, line 138: **Medium — Reimplements `slices.Equal`**

This hand-rolled `slicesEqual` duplicates `slices.Equal` from stdlib (Go 1.21+), which is already used elsewhere in this codebase.

_Consensus: 2/2 agents_

cmd/api/logging/verification.go
```diff
@@ -0,0 +1,319 @@
...
+
+type VerificationConfig struct {
+	TailLines         int    `yaml:"tail-lines"`
+	Interval          string `yaml:"interval"`
+	MaxRetries        int    `yaml:"max-retries"`
+	TimestampJsonPath string `yaml:"timestamp-json-path"`
+	TimestampFormat   string `yaml:"timestamp-format"`
+}
+
+func (vc *VerificationConfig) defaults() {
```
maruiz93, line 38: **Low — No bounds on `MaxRetries` or minimum on `Interval`**

Validation doesn't enforce upper/lower bounds. An operator could set extreme values causing a tight loop against the backend. Also, `intervalDuration()` (line 68) silently returns 0 on parse failure, which would cause a busy-loop.

Consider adding bounds in `validate()` (e.g., MaxRetries <= 30, Interval >= 1s) and a floor in `intervalDuration()`.

_Consensus: 3/3 agents_

cmd/api/logging/verification.go
```diff
@@ -0,0 +1,319 @@
...
+	default:
+		return time.Parse(format, raw)
+	}
+}
+
+// tryParseTimestamp attempts to parse a timestamp in common formats:
+// integer-based (nanoseconds, milliseconds, seconds) and RFC3339.
+func tryParseTimestamp(s string) (time.Time, error) {
+	if n, err := strconv.ParseInt(s, 10, 64); err == nil {
+		if n > 1e15 {
```
maruiz93, line 107: **Low — Magic number thresholds undocumented**

The `1e15` and `1e12` heuristics for auto-detecting nano/milli/seconds are uncommented. They'll produce incorrect results for extreme timestamps. Consider named constants with comments, or prefer explicit format parsing since `coarseSkip` operates on fields whose format the system controls.

_Consensus: 2/2 agents_

cmd/api/logging/verification.go
```diff
@@ -0,0 +1,319 @@
...
+					}
+				}
+			}
+		}
+	}
+
+	return lines, lastTimestamp, nil
+}
+
+func performVerification(
```
maruiz93, line 233: **Low — No deduplication of concurrent verification probes**

Each incoming log request spawns its own verification loop. N concurrent clients for the same pod generate N × (MaxRetries + 1) probes. Consider `singleflight.Group` keyed on `(podName, containerName)`.

_Single agent finding_

cmd/api/logging/verification.go
```diff
@@ -0,0 +1,319 @@
...
+
+	response, err := client.Do(request)
+	if err != nil {
+		return nil, time.Time{}, fmt.Errorf("tail probe request failed: %w", err)
+	}
+	defer response.Body.Close()
+
+	if response.StatusCode != http.StatusOK {
+		body, _ := io.ReadAll(io.LimitReader(response.Body, 512))
+		return nil, time.Time{}, fmt.Errorf("tail probe got status %d: %s", response.StatusCode, body)
```
maruiz93, line 183: **Low — Backend error body logged without sanitization**

Up to 512 bytes of the backend's error body are included verbatim in the error and logged. Internal URLs, stack traces, or tokens could be persisted. Consider logging only the HTTP status code.

_Consensus: 2/3 agents_

cmd/api/logging/logging.go
```diff
@@ -511,6 +525,29 @@ func LogRetrieval() gin.HandlerFunc {
...
+
+			verResult = performVerification(
+				c.Request.Context(),
+				&client,
+				providerConfig.Tail,
+				record,
+				headers,
+				endpoint.Verification,
+				time.Now,
+			)
```
maruiz93, line 541: **High — Synchronous verification blocks response streaming (up to 30s latency)**

`performVerification` runs a polling loop before any log data is streamed. With default Loki config (5s interval, 6 retries), this adds up to 30s latency before the first byte reaches the client. The coarse skip doesn't help for recently-active pods — the most common case needing verification.

Consider running verification concurrently with streaming: start the stream immediately, collect `lastLine` during streaming, then check stability afterward via the trailer mechanism. Alternatively, use exponential backoff starting shorter (500ms, 1s, 2s, 4s). At minimum, wrap the call in a dedicated timeout context.

_Consensus: 3/3 agents_

maruiz93 (changes requested):
I've run several agents to perform the review of this PR and I reviewed the proposals and they make sense to me.

PR #1929, opened 2026-06-04
1927 replace resource label index with covering index

### LINKED ISSUE #1927: Replace resource_label(label_id) index with covering (label_id, resource_id) index
## Summary

The `idx_resource_label_key_value` index on `resource_label(label_id)` forces a bitmap heap scan to retrieve `resource_id` after finding label matches. Replacing it with a covering index on `(label_id, resource_id)` enables index-only scans, eliminating the heap access entirely.

## Problem

When querying resources by label, PostgreSQL:
1. Uses `idx_resource_label_key_value(label_id)` to find matching entries
2. Goes back to the `resource_label` heap to fetch `resource_id` for each match

In production with ~1.47B rows in `resource_label` (168GB table), this heap scan read 21,167 pages from disk and accounted for **109 seconds** of a 316-second query.

## Proposed change

Modify migration `08_label_resource_constraints` to create the covering index instead:

```sql
-- Before
CREATE INDEX IF NOT EXISTS idx_resource_label_key_value ON resource_label(label_id);

-- After
CREATE INDEX IF NOT EXISTS idx_resource_label_label_resource ON resource_label(label_id, resource_id);
```

Since migration 08 has already been included in a release, modifying it in place will only affect new installations. Clusters that have already applied migration 08 will need the index recreated manually:

```sql
CREATE INDEX CONCURRENTLY idx_resource_label_label_resource ON resource_label(label_id, resource_id);
DROP INDEX idx_resource_label_key_value;
```

Note: `CREATE INDEX CONCURRENTLY` on a large `resource_label` table can take several hours (4-5h observed on a 1.47B row table) but does not block writes.

## Results

This was tested in production and reduced the `resource_label` scan from 109s to 16ms (index-only scan, no heap access). Combined with warmer shared_buffers, the full query went from 316s to 6.3s.

olegbet:
Resolves #1927

```release-note
Improved performance of reads with the new DB normalized schema introduced in v1.21. Check how to take advantage of this fix if you have already migrated [here](https://github.com/kubearchive/kubearchive/issues/1927)
```

### DIFF
--- a/integrations/database/postgresql/migrations/08_label_resource_constraints.down.sql
+++ b/integrations/database/postgresql/migrations/08_label_resource_constraints.down.sql
@@ -2,4 +2,4 @@ ALTER TABLE resource_label ALTER COLUMN label_id SET STATISTICS -1;
 ALTER TABLE resource_label DROP CONSTRAINT IF EXISTS fk_resource_label_label_id;
 ALTER TABLE label_key_value DROP CONSTRAINT IF EXISTS fk_label_key_value_value_id;
 ALTER TABLE label_key_value DROP CONSTRAINT IF EXISTS fk_label_key_value_key_id;
-DROP INDEX IF EXISTS idx_resource_label_key_value;
+DROP INDEX IF EXISTS idx_resource_label_label_resource;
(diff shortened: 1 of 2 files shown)


maruiz93:
We need to release this as a bugfix and for that we need release notes

maruiz93 (changes requested):
Add release notes please

maruiz93 (approved):

PR #1933, opened 2026-06-07
chore(#1934): write AI skills for kubearchive

### LINKED ISSUE #1934: Write AI skills for kubearchive/kubearchive
Agent skills: Create agent skills for each repo, stored in a skills/ directory. Symlink them in a .claude/ directory (or equivalent for your agent runtime / coding assistant) so agents discover them automatically. Focus on knowledge the agent won't already have -- repo-specific procedures, not general programming knowledge.

olegbet:
Resolves #1934

```release-note
Creates agent skills for kubearchive repo, stored in a skills/ directory.
```


maruiz93:
hmmm... this is somehow duplicating the docs that we already have in adoc format. I believe we should avoid duplication. Even if the agents are the ones that will be updating the docs, I don't think we should have things duplicated. Could we consider just pointing to the proper docs instead replicating what's in the docs?

olegbet:
> hmmm... this is somehow duplicating the docs that we already have in adoc format. I believe we should avoid duplication. Even if the agents are the ones that will be updating the docs, I don't think we should have things duplicated. Could we consider just pointing to the proper docs instead replicating what's in the docs?

I suppose it does duplicate the docs a bit but it is not a part of "human" documentation.  It is more like the concentrated tips from the docs and I think this will be easier and quicker to ingest for agents instead of following the crosslinks.

maruiz93:
> > hmmm... this is somehow duplicating the docs that we already have in adoc format. I believe we should avoid duplication. Even if the agents are the ones that will be updating the docs, I don't think we should have things duplicated. Could we consider just pointing to the proper docs instead replicating what's in the docs?
> 
> I suppose it does duplicate the docs a bit but it is not a part of "human" documentation. It is more like the concentrated tips from the docs and I think this will be easier and quicker to ingest for agents instead of following the crosslinks.

Well, I disagree. I think it's not a bit, I actually see information actually duplicated from markdown files that we have (like DEVELOPMENT.md). I think that the approach should be:

* Instead of recreating all the docs, provide skills that have proper links to the existing docs, the ones with the *adoc format, I'm pretty sure the agents can understand that skill. Then, also provide proper links to the markdowns that we already have, e.g. the DEVELOPMENT.md
* If you think it's needed, add a new issue that states that you should check if the adoc documentation is good enough for the agents or if you should move on to markdown docs so they can be used by both agents and humans

Please remember that one of the advantages of the AI prompts is that they are human-readable which means that they can actually be used as human docs

.claude/effective-go
```diff
@@ -0,0 +1 @@
+./skills/effective-go/
```
maruiz93, line 1: make sure you add newlines at the end of the files. That's actually something that should be checked in our lints and done by the agents

skills/kubearchive-contributing/SKILL.md
```diff
@@ -0,0 +1,207 @@
+---
```
maruiz93: this information for example should be in the CONTRIBUTING.md. If something is missing, please add it there, and I don't think this should be a skill, but more an information in the AGENTS.md if anything

skills/kubearchive-api/SKILL.md
```diff
@@ -0,0 +1,338 @@
...
+# Create a token for a service account
+TOKEN=$(kubectl create -n <namespace> token <service-account>)
+
+# Use it
+curl -s --cacert ca.crt \
+  -H "Authorization: Bearer $TOKEN" \
+  https://localhost:8081/apis/batch/v1/namespaces/<namespace>/jobs
+```
+
+For local development, set up test users first:
```
maruiz93: this is all in DEVELOPMENT.md. If something is missing, please add it there and link it there

maruiz93:
there is sometihng else I think it's important to mention. The skills are like commands that you run, but I don't think it actually makes sense to run /effective-go or /kubearchive-api. I think we should have skills like:

`/query` in which you can provide for example a url and it extracts the query of the database given a db engine, or /how-to-query to provide a prompt of what you want to query and provides you the url or the cli command or maybe a `/configure-cel` to provide guidance on how to set up your kubearchiveconfig

skills/effective-go/SKILL.md
```diff
@@ -0,0 +1,33 @@
...
+# Effective Go
+
+Apply best practices and conventions from the official [Effective Go guide](https://go.dev/doc/effective_go) to write clean, idiomatic Go code.
+
+## When to Apply
+
+Use this skill automatically when:
+- Writing new Go code
+- Reviewing Go code
+- Refactoring existing Go implementations
```
maruiz93, line 16: this is a better skill, as it has a defined place of execution

skills/kubearchive-debugging/SKILL.md
```diff
@@ -0,0 +1,280 @@
+---
+name: kubearchive-debugging
```
maruiz93: again this fits more in the DEVELOPMENT.md

skills/kubearchive-local-setup/references/ide-debug-config.md
```diff
@@ -0,0 +1,55 @@
+# IDE Debug Configuration
```
maruiz93: Have you tried this? Can the agent actually take advantage of breakpoints on development? I think they understand better logs. One thing that could be done along this effort is making sure the project has good debug logs.

skills/kubearchive-operator-dev/SKILL.md
```diff
@@ -0,0 +1,428 @@
+---
+name: kubearchive-operator-dev
+description: Develop and modify the KubeArchive operator — reconcilers, CRDs, webhooks, CEL expressions, dynamic watches, RBAC management, and testing with envtest. Use this skill when someone is writing or modifying operator controllers, adding CRD fields, implementing webhooks, working with CEL filter expressions, managing dynamic resource watches, debugging reconciliation loops, or writing operator tests. Also trigger on "kubearchive reconciler", "KubeArchiveConfig CRD", "SinkFilter controller", "operator webhook", "CEL expression kubearchive", "envtest kubearchive", or "operator RBAC".
+---
+
+# KubeArchive Operator Development
+
+The KubeArchive operator is a controller-runtime application with three reconcilers, five CRDs, and a dynamic resource watching system. This skill covers the patterns and conventions you need to modify or extend it.
```
maruiz93: This is not a skill, it's just a description, this fits better in the CLAUDE.md and better, linked to the proper docs instead of duplicating. If something is missing we should add them on an architecture section of the human docs.

maruiz93 (approved):
As a good starting point I think the effective-go skill is ok, kubearchive repo could use more if needed while the maintainers/collaborators make use of them

PR #1939, opened 2026-06-15
#1937: limit request rate for kubearchive-api-server

### LINKED ISSUE #1937: Limit request rate for kubearchive-api-server
## Problem

To prevent DDoS attacks on KubeArchive API server the request rate per second to KubeArchive API server should include a rate limit mechanism. It should be configurable through env vars

## Expected behavior

Number of request per second to API server should be limited and configured through environment variable config. It should have a default value.

olegbet:
Assisted-by: Claude

Resolves #1937

```release-note
This change limits API server resources  and logs requests rate per second It is can be configurable by the environment variables:
        - API_RATE_LIMIT_OVERALL_RPS
	- API_RATE_LIMIT_LOG_RPS
	- API_RATE_LIMIT_OVERALL_BURST
	- API_RATE_LIMIT_LOG_BURST
	- API_MAX_CONCURRENT_REQUESTS
        - API_MAX_CONCURRENT_LOG_REQUESTS
```


cmd/api/main.go
```diff
@@ -203,6 +228,60 @@ func main() {
...
+func getRateLimitConfig() RateLimitConfig {
+	cfg := RateLimitConfig{
+		OverallRPS:       20,
+		LogRPS:           2,
+		OverallBurst:     20,
+		LogBurst:         2,
+		MaxConcurrent:    50,
+		MaxConcurrentLog: 5,
+	}
+	if v := os.Getenv(rateLimitOverallRPSEnvVar); v != "" {
```
maruiz93: **[High]** `getRateLimitConfig()` accepts any parsed value without validation. This has three failure modes:

- Negative `MaxConcurrent`/`MaxConcurrentLog` → `make(chan struct{}, -1)` **panics at runtime**, crashing the API server
- Zero values for any parameter → all requests silently rejected (self-inflicted DoS)
- Negative RPS → `rate.NewLimiter` allows all requests, **disabling rate limiting entirely**

Note that `getCacheExpirations` in the same file validates its inputs — this function should do the same. Suggestion: guard each parsed value with `if parsed <= 0 { slog.Warn(...); }` and fall back to the default.

cmd/api/main.go
```diff
@@ -73,28 +90,35 @@ func NewServer(k8sClient kubernetes.Interface, controller routers.Controller, ca
 		group.Use(auth.RBACAuthorization(k8sClient.AuthorizationV1().SubjectAccessReviews(), cache,
 			cacheExpirations.Authorized, cacheExpirations.Unauthorized))
 		group.Use(pagination.Middleware())
+		group.Use(middleware.RateLimiter(rateLimits.OverallRPS, rateLimits.OverallBurst))
```
maruiz93: **[High]** The rate limiter is placed after all auth middleware (Authentication, Impersonation, Discovery, RBAC — lines 84-92). Every rate-limited request still makes 3+ Kubernetes API calls before being rejected, so the limiter doesn't protect the K8s API from request floods.

A per-user rate limiter can be placed **before** auth by keying on available request headers: `Impersonate-User` for proxied requests, or the bearer token (or a hash of it) from the `Authorization` header for direct callers. Requests with neither get a shared anonymous bucket with stricter limits. This protects the K8s API without needing to resolve user identity first.

cmd/api/main.go
```diff
@@ -73,28 +90,35 @@ func NewServer(k8sClient kubernetes.Interface, controller routers.Controller, ca
...
+		group.Use(middleware.ConcurrentLimiter(rateLimits.MaxConcurrent))
```
maruiz93: **[Medium]** Each iteration of the `for _, group := range groups` loop (line 83) calls `middleware.RateLimiter()` and `middleware.ConcurrentLimiter()`, which create **new independent** limiter instances per call (`ratelimit.go:18`). Since the loop iterates over both `/api` and `/apis`, two independent token buckets and two independent semaphores are created. The effective server-wide limit is 2x the configured value (e.g., 40 RPS instead of 20).

Create the limiter instances once before the loop and share them across both groups, or refactor `RateLimiter`/`ConcurrentLimiter` to accept pre-created limiter objects.

pkg/middleware/ratelimit.go
```diff
@@ -0,0 +1,43 @@
...
+
+	"github.com/gin-gonic/gin"
+	"github.com/kubearchive/kubearchive/pkg/abort"
+	"golang.org/x/time/rate"
+)
+
+// RateLimiter returns middleware that enforces a token-bucket rate limit.
+// Requests exceeding the limit are rejected immediately with 429.
+func RateLimiter(rps float64, burst int) gin.HandlerFunc {
+	limiter := rate.NewLimiter(rate.Limit(rps), burst)
```
maruiz93: **[High]** The rate limiter uses a single global token bucket shared across all clients. One user's "Download all logs" on a 16-task pipeline fires 47+ concurrent log requests (empirically measured on Konflux UI staging) — this exhausts the entire log budget, and every other user gets 429s until the bucket refills. Production pipelines with 20-30+ tasks would fire 60-100+ log requests.

Consider per-user rate limiting keyed by `Impersonate-User` header (set by the nginx proxy) or bearer token for direct callers. Each user gets their own budget, so legitimate use never starves other users. A global concurrent ceiling on top still serves as a server-protection backstop.

cmd/api/main.go
```diff
@@ -203,6 +228,60 @@ func main() {
 	slog.Debug("Shutdown reached timeout")
 }
 
+func getRateLimitConfig() RateLimitConfig {
+	cfg := RateLimitConfig{
+		OverallRPS:       20,
```
maruiz93: **[Medium]** The default rate limits are too low for Konflux UI usage patterns. Empirical data from staging (16-task pipeline, "Download all task logs"):

- 47 concurrent log requests + 16 pod metadata requests in a single burst
- Current log defaults (2 RPS / burst 2 / 5 concurrent) → 42+ of 47 log requests get 429
- Production pipelines have 20-30+ tasks, meaning 60-100+ log requests per download

Additional pressure from normal browsing: namespace switches invalidate React Query cache (new query keys = fresh requests), scan results auto-drain all pages via `useEffect` loop, and `refetchOnWindowFocus` triggers refetches on tab switch. Automation scripts hit the API without any caching.

With per-user rate limiting, suggested defaults: 100 RPS / 100 burst / 100 concurrent for both overall and log limiters.

Recommendation: measure actual request volume from a Konflux UI client session (e.g., browsing pipelines, downloading logs, navigating across tenants) to validate that the chosen defaults won't break existing workflows. If using AI to assist with this PR, provide the [konflux-ui](https://github.com/konflux-ci/konflux-ui) repository as context — the request patterns are driven by its React Query hooks and `Promise.all` fan-outs.

cmd/api/main.go
```diff
@@ -203,6 +228,60 @@ func main() {
 	slog.Debug("Shutdown reached timeout")
 }
 
+func getRateLimitConfig() RateLimitConfig {
```
maruiz93: **[Medium]** `getRateLimitConfig` has no unit tests, unlike `getCacheExpirations` in the same file which has thorough coverage (`TestGetCacheExpirationsErrorPaths`, `TestGetCacheExpirationsWorkingProperly`). Given that misconfiguration can panic the server or silently block all traffic, test coverage is important. Suggested cases: defaults when no env vars are set, valid overrides, unparseable values falling back to defaults, and out-of-bounds values.

cmd/api/main.go
```diff
@@ -203,6 +228,60 @@ func main() {
...
+		MaxConcurrentLog: 5,
+	}
+	if v := os.Getenv(rateLimitOverallRPSEnvVar); v != "" {
+		if parsed, err := strconv.ParseFloat(v, 64); err == nil {
+			cfg.OverallRPS = parsed
+		} else {
+			slog.Warn("Could not parse env var, using default", "var", rateLimitOverallRPSEnvVar, "value", fmt.Sprintf("%q", v))
+		}
+	}
+	if v := os.Getenv(rateLimitLogRPSEnvVar); v != "" {
```
maruiz93: **[Medium]** `getRateLimitConfig` repeats the same parse-or-warn pattern six times. A pair of helpers (`getEnvFloat64`, `getEnvInt`) would collapse it to ~10 lines and eliminate the risk of copy-paste bugs (e.g., wrong env var name in a log message).

cmd/api/main.go
```diff
@@ -73,28 +90,35 @@ func NewServer(k8sClient kubernetes.Interface, controller routers.Controller, ca
...
 			cacheExpirations.Authorized, cacheExpirations.Unauthorized))
 		group.Use(pagination.Middleware())
+		group.Use(middleware.RateLimiter(rateLimits.OverallRPS, rateLimits.OverallBurst))
+		group.Use(middleware.ConcurrentLimiter(rateLimits.MaxConcurrent))
 	}
 
 	router.GET("/livez", controller.Livez)
 	router.GET("/readyz", controller.Readyz)
 
+	logMiddleware := []gin.HandlerFunc{
```
maruiz93: **[Low]** `logMiddleware` (line 100) is used in four `append()` calls. Currently safe since len=cap=2 forces a new allocation each time, but if a third element is ever added to the slice, the first `append` could modify the backing array and corrupt subsequent route registrations. Consider `slices.Clone(logMiddleware)` before each append.

pkg/middleware/ratelimit.go
```diff
@@ -0,0 +1,43 @@
...
+	return func(c *gin.Context) {
+		if !limiter.Allow() {
+			abort.Abort(c, errors.New("too many requests"), http.StatusTooManyRequests)
```
maruiz93: **[Medium]** 429 responses don't include a `Retry-After` header (RFC 6585 recommendation). Without it, clients may retry immediately in a tight loop. Consider setting `Retry-After` — either a static value like `"1"` or computed from `limiter.Reserve().Delay()`.

cmd/api/main.go
```diff
@@ -43,13 +50,23 @@ var (
 	date    = ""
 )
 
+// RateLimitConfig holds the configuration for the API rate limiters.
+type RateLimitConfig struct {
```
maruiz93: **[Medium]** `RateLimitConfig` and `getRateLimitConfig()` live in `main.go`, but the middleware they configure is in `pkg/middleware/`. The existing `LoggerConfig` follows the opposite (and better) pattern — it lives in `pkg/middleware/logging.go` next to its middleware. Consider moving config and constructor to `pkg/middleware/`.

pkg/middleware/ratelimit.go
```diff
@@ -0,0 +1,43 @@
...
+// ConcurrentLimiter returns middleware that enforces a maximum number of
+// in-flight requests using a semaphore channel. Requests that arrive when
+// the semaphore is full are rejected immediately with 429.
+func ConcurrentLimiter(max int) gin.HandlerFunc {
+	sem := make(chan struct{}, max)
+	return func(c *gin.Context) {
+		select {
+		case sem <- struct{}{}:
+		default:
+			abort.Abort(c, errors.New("too many requests"), http.StatusTooManyRequests)
```
maruiz93, line 84: **[Low]** Both `RateLimiter` and `ConcurrentLimiter` use the same error message `"too many requests"`. Operators seeing 429s in logs can't distinguish RPS exhaustion from concurrency exhaustion — different signals needing different remediation. Consider distinct messages like `"rate limit exceeded"` vs `"too many concurrent requests"`.

cmd/api/main_test.go
```diff
@@ -25,7 +25,12 @@ func fakeServer(k8sClient kubernetes.Interface, cache *cache.Cache) *Server {
 	}
 	controller := routers.Controller{Database: fakeDB.NewFakeDatabase(nil, nil)}
 	expirations := &routers.CacheExpirations{Authorized: 1 * time.Second, Unauthorized: 1 * time.Second}
-	return NewServer(k8sClient, controller, cache, expirations)
+	rateLimits := RateLimitConfig{
+		OverallRPS: 20, LogRPS: 2,
```
maruiz93: **[Low]** `fakeServer` in the test hardcodes the same default values (20, 2, 20, 2, 50, 5) that appear in `getRateLimitConfig`. If defaults change, the test silently uses stale values. Consider calling `getRateLimitConfig()` directly or using named constants as a single source of truth.

maruiz93 (changes requested):
Thanks for addressing this. I've posted the things that I believe should be fixed before merging but the most critical from my point of view is that we shouldn't rely on a global rate limit (we could have it, but with really large numbers), I would focus on doing rate-limit per user.

The second most important thing is to do the rate limit as soon as possible (not after the auth checks)

cmd/api/main.go
olegbet: Done

maruiz93: Thanks for addressing this. The singleton approach ensures one limiter is shared across both `/api` and `/apis` groups, which was the goal. However, it introduces a regression: `RateLimiter()` and `ConcurrentLimiter()` now silently ignore their parameters after the first call.

In `main.go`, the log-specific limiter is created first:

```go
rateLimitMiddleware := []gin.HandlerFunc{
    middleware.RateLimiter(rateLimits.LogRPS, rateLimits.LogBurst),       // first call → creates singleton (10 RPS, burst 30)
    middleware.ConcurrentLimiter(rateLimits.MaxConcurrentLog),            // first call → creates singleton (max 20)
}

for _, group := range groups {
    group.Use(middleware.RateLimiter(rateLimits.OverallRPS, rateLimits.OverallBurst))  // second call → returns the SAME handler (10 RPS, not 100)
    // ...
    group.Use(middleware.ConcurrentLimiter(rateLimits.MaxConcurrent))                  // second call → returns the SAME handler (max 20, not 200)
}
```

Because the singleton is set by the first call, the overall limiter silently uses log-level limits: 10 RPS instead of 100, and 20 concurrent instead of 200.

A simpler approach: remove the singleton, let each call create an independent handler, and control sharing from the caller:

```go
// ratelimit.go — no globals, each call returns an independent handler
func RateLimiter(rps float64, burst int) gin.HandlerFunc {
    limiter := rate.NewLimiter(rate.Limit(rps), burst)
    return func(c *gin.Context) {
        if !limiter.Allow() {
            abort.Abort(c, errors.New("too many requests"), http.StatusTooManyRequests)
            return
        }
        c.Next()
    }
}

// main.go — create each limiter once, reuse across groups
overallRL := middleware.RateLimiter(rateLimits.OverallRPS, rateLimits.OverallBurst)
overallCL := middleware.ConcurrentLimiter(rateLimits.MaxConcurrent)
logRL     := middleware.RateLimiter(rateLimits.LogRPS, rateLimits.LogBurst)
logCL     := middleware.ConcurrentLimiter(rateLimits.MaxConcurrentLog)

for _, group := range groups {
    group.Use(overallRL)   // same handler for both /api and /apis ✓
    // ...
    group.Use(overallCL)
}
// log routes use logRL and logCL
```

This keeps the fix (one limiter shared across groups) while supporting distinct limits for overall vs log routes.

cmd/api/main.go
maruiz93: The rate limiter was moved before auth — thanks. However, `ConcurrentLimiter` (line 85) is still the last middleware in the chain, after authentication, impersonation, discovery, RBAC, and pagination. Every request that hits the concurrency check has already made 3+ K8s API calls.

Moving `ConcurrentLimiter` next to `RateLimiter` (before auth) would let the server cheaply reject excess traffic before any expensive I/O.

pkg/middleware/ratelimit.go
```diff
@@ -0,0 +1,135 @@
...
+	var result float64 = defaultVal
+	if v := os.Getenv(envVar); v != "" {
+		if r, err := strconv.ParseFloat(v, 64); err == nil {
+			result = r
+		} else {
+			slog.Warn("Could not parse env var, using default", "var", defaultVal, "value", fmt.Sprintf("%q", v))
+			result = defaultVal
+		}
+	}
+	return result, err
```
maruiz93: **[High]** Both `getEnvFloat64` and `getEnvInt` have a variable shadowing bug: `var err error` on the outer scope is never assigned — the inner `if r, err := strconv.ParseFloat(...)` uses `:=` which creates a new `err` scoped to the `if` block. The returned error is always `nil`.

The callers already discard it with `_ =`, so this is benign in practice, but the function signature is misleading. Since the functions already log warnings and fall back to defaults internally, the simplest fix is to drop the error return:

```go
func getEnvFloat64(envVar string, defaultVal float64) float64 {
    if v := os.Getenv(envVar); v != "" {
        if r, err := strconv.ParseFloat(v, 64); err == nil {
            return r
        }
        slog.Warn("Could not parse env var, using default", "var", envVar, "value", v, "default", defaultVal)
    }
    return defaultVal
}
```

pkg/middleware/ratelimit.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+func getEnvFloat64(envVar string, defaultVal float64) (float64, error) {
+	var err error
+	var result float64 = defaultVal
+	if v := os.Getenv(envVar); v != "" {
+		if r, err := strconv.ParseFloat(v, 64); err == nil {
+			result = r
+		} else {
+			slog.Warn("Could not parse env var, using default", "var", defaultVal, "value", fmt.Sprintf("%q", v))
```
maruiz93: **[Medium]** `"var", defaultVal` logs the numeric default value (e.g., `100`) as the variable identifier instead of the environment variable name. Same bug on line 98.

Should be `"var", envVar` so operators can identify which env var failed to parse.

pkg/middleware/ratelimit.go
```diff
@@ -0,0 +1,135 @@
...
+	cfg.OverallBurst, _ = getEnvInt(rateLimitOverallBurstEnvVar, cfg.OverallBurst)
+	cfg.LogBurst, _ = getEnvInt(rateLimitLogBurstEnvVar, cfg.LogBurst)
+	cfg.MaxConcurrent, _ = getEnvInt(maxConcurrentRequestsEnvVar, cfg.MaxConcurrent)
+	cfg.MaxConcurrentLog, _ = getEnvInt(maxConcurrentLogRequestsEnvVar, cfg.MaxConcurrentLog)
+
+	if cfg.OverallRPS <= 0 || cfg.LogRPS <= 0 ||
+		cfg.OverallBurst <= 0 || cfg.LogBurst <= 0 ||
+		cfg.MaxConcurrent <= 0 || cfg.MaxConcurrentLog <= 0 {
+		slog.Error("Invalid rate limit configuration: all values must be positive")
+		os.Exit(1)
```
maruiz93: **[Medium]** `os.Exit(1)` in a `pkg/` library function breaks testability and project conventions — every other `os.Exit` in `pkg/` is in test code, and `getCacheExpirations()` in the same `main.go` file follows the correct pattern (returns an error).

This also makes it impossible to test the validation path — the test suite has no negative-value test cases, likely for this reason.

Suggestion: return `(RateLimitConfig, error)` and let the caller in `main.go` handle it:

```go
rateLimits, err := middleware.GetRateLimitConfig()
if err != nil {
    slog.Error(err.Error())
    os.Exit(1)
}
```

pkg/middleware/ratelimit.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	cfg.OverallRPS, _ = getEnvFloat64(rateLimitOverallRPSEnvVar, cfg.OverallRPS)
+	cfg.LogRPS, _ = getEnvFloat64(rateLimitLogRPSEnvVar, cfg.LogRPS)
+	cfg.OverallBurst, _ = getEnvInt(rateLimitOverallBurstEnvVar, cfg.OverallBurst)
+	cfg.LogBurst, _ = getEnvInt(rateLimitLogBurstEnvVar, cfg.LogBurst)
+	cfg.MaxConcurrent, _ = getEnvInt(maxConcurrentRequestsEnvVar, cfg.MaxConcurrent)
+	cfg.MaxConcurrentLog, _ = getEnvInt(maxConcurrentLogRequestsEnvVar, cfg.MaxConcurrentLog)
+
+	if cfg.OverallRPS <= 0 || cfg.LogRPS <= 0 ||
```
maruiz93, line 145: **[Medium]** `strconv.ParseFloat` successfully parses `"Inf"`, `"+Inf"`, `"-Inf"`, and `"NaN"`. The validation check `cfg.OverallRPS <= 0` doesn't catch these: `Inf > 0` is true, and `NaN <= 0` is false (all NaN comparisons return false).

Setting `API_RATE_LIMIT_OVERALL_RPS=Inf` would disable rate limiting entirely. Setting it to `NaN` produces undefined limiter behavior.

Suggestion: add a check after parsing float values:

```go
if math.IsInf(result, 0) || math.IsNaN(result) {
    slog.Warn("Invalid float value for env var, using default", "var", envVar, "value", v)
    result = defaultVal
}
```

maruiz93 (changes requested):
I've added some comments on threads that were already opened and a couple more reviews, please take a look

maruiz93 (changes requested):
Two things about the overall architecture:

**1. Per-user rate limiting should apply to all endpoints, not just log routes.**

Currently `UserRateLimiter` is only used for log routes — the overall rate limiter (`RateLimiter`) is still global. This means one user can still starve others on the resource endpoints (e.g., a script querying pipeline runs across many namespaces). The same `Impersonate-User` / Bearer / IP key extraction already works — it just needs to be wired up for the overall limiter too.

**2. The concurrent limiters (`ConcurrentLimiter`) can be removed.**

Concurrent request limiting is already handled at the layers that actually enforce the bottleneck:
- For non-log requests, database connection pooling (https://github.com/kubearchive/kubearchive/pull/1946) caps concurrent DB access via `SetMaxOpenConns`. Requests that can't get a connection wait or timeout — no need for a separate semaphore in the middleware.
- For log requests, the concurrency constraint is in Loki (or whatever logging backend is configured), not in our API server. Adding a semaphore here is guessing at the backend's capacity from the wrong layer.

The `ConcurrentLimiter` at 200 slots doesn't add meaningful protection when the DB pool allows 25 connections — it just adds complexity without changing the actual bottleneck.

maruiz93 (changes requested):
Also, by mistake you updated the .claude files, please undo that

olegbet:
> Also, by mistake you updated the .claude files, please undo that

Done

olegbet:
> Two things about the overall architecture:
> 
> **1. Per-user rate limiting should apply to all endpoints, not just log routes.**
> 
> Currently `UserRateLimiter` is only used for log routes — the overall rate limiter (`RateLimiter`) is still global. This means one user can still starve others on the resource endpoints (e.g., a script querying pipeline runs across many namespaces). The same `Impersonate-User` / Bearer / IP key extraction already works — it just needs to be wired up for the overall limiter too.
> 
> **2. The concurrent limiters (`ConcurrentLimiter`) can be removed.**
> 
> Concurrent request limiting is already handled at the layers that actually enforce the bottleneck:
> 
> * For non-log requests, database connection pooling (#1946) caps concurrent DB access via `SetMaxOpenConns`. Requests that can't get a connection wait or timeout — no need for a separate semaphore in the middleware.
> * For log requests, the concurrency constraint is in Loki (or whatever logging backend is configured), not in our API server. Adding a semaphore here is guessing at the backend's capacity from the wrong layer.
> 
> The `ConcurrentLimiter` at 200 slots doesn't add meaningful protection when the DB pool allows 25 connections — it just adds complexity without changing the actual bottleneck.

Done

pkg/middleware/ratelimit.go
```diff
@@ -0,0 +1,146 @@
...
+//
+// Requests exceeding the per-user limit are rejected immediately with 429 and
+// a Retry-After header indicating how many seconds to wait (RFC 6585 §4).
+//
+// TODO: evict stale entries if user population grows unbounded.
+func UserRateLimiter(rps float64, burst int) gin.HandlerFunc {
+	var limiters sync.Map
+	return func(c *gin.Context) {
+		key := userKey(c)
+		v, _ := limiters.LoadOrStore(key, rate.NewLimiter(rate.Limit(rps), burst))
```
maruiz93: `LoadOrStore` allocates a new `rate.Limiter` on every request, even for existing users — the allocation is immediately GC'd on cache hit. Use `Load` first, fall back to `LoadOrStore` on miss:

```go
v, ok := limiters.Load(key)
if !ok {
    v, _ = limiters.LoadOrStore(key, rate.NewLimiter(rate.Limit(rps), burst))
}
```

docs/modules/ROOT/pages/configuration/rate-limiting.adoc
```diff
@@ -0,0 +1,70 @@
...
+To modify rate limiting settings, perform the following steps:
+
+. Edit or patch the `kubearchive-api-server` Deployment present on the `kubearchive`
+namespace:
++
+[source,yaml]
+----
+---
+apiVersion: v1
+kind: Deployment
```
maruiz93, line 43: `apiVersion: v1` should be `apiVersion: apps/v1` for a Deployment.

pkg/middleware/ratelimit.go
```diff
@@ -0,0 +1,146 @@
...
+		limiter := v.(*rate.Limiter)
+		r := limiter.Reserve()
```
maruiz93, line 75: `Reserve()` is called without checking `r.OK()`. With the current validation this is always true, but a defensive check would be more robust.

pkg/middleware/ratelimit_test.go
```diff
@@ -0,0 +1,445 @@
...
+			req := httptest.NewRequest(http.MethodGet, "/log", nil)
+			if header != "" {
+				req.Header.Set(header, value)
+			}
+			router.ServeHTTP(w, req)
+			switch w.Code {
+			case http.StatusOK:
+				ok++
+			case http.StatusTooManyRequests:
+				limited++
```
maruiz93, line 37: The `Retry-After` header is set on 429 responses but no test verifies its presence or value.

maruiz93 (approved):
I provided several comments that would be nice to fix but they are not blockers. I've also created a follow up for the TODO you introduced about not evicting the map with the ratelimit information:  https://github.com/kubearchive/kubearchive/issues/1984

pkg/middleware/ratelimit.go
olegbet: Done

docs/modules/ROOT/pages/configuration/rate-limiting.adoc
olegbet: Done

pkg/middleware/ratelimit.go
olegbet: Done

PR #1945, opened 2026-06-18
fix(1928): add configurable statement timeout for database read queries

### LINKED ISSUE #1928: Add configurable statement timeout for database read queries
## Summary

The API server currently has no timeout on database read queries. Long-running queries can run indefinitely (observed: 16+ hours), degrading performance for all other queries by thrashing shared_buffers and saturating I/O.

## Problem

The HTTP server intentionally does not set `ReadTimeout` to support long-running streaming responses. The database role has `statement_timeout = 0` (disabled). When a client or upstream proxy (nginx, HAProxy) times out, the database query continues running if the upstream connection is not fully closed.

In production, 24 concurrent namespace-less resource listing queries ran for 16+ hours, each scanning the full 1.1TB `resource` table. This caused:
- **Cache thrashing**: shared_buffers (1.9GB) were filled with cold data, evicting hot pages. Normal queries went from 6.3s to 316s+.
- **I/O saturation**: 24 concurrent full-table scans consumed all disk bandwidth.
- **Parallel worker starvation**: all query plans showed `Workers Planned: 2, Workers Launched: 0`.

## Proposed change

Wrap database read queries with a context deadline in `GetResources` (and similar handlers in `cmd/api/routers/routers.go`):

```go
ctx, cancel := context.WithTimeout(context.Request.Context(), queryTimeout)
defer cancel()
```

Pass `ctx` instead of `context.Request.Context()` to `QueryResources` and `StreamResources`.

The timeout should:
- Default to **5 minutes**
- Be configurable via an environment variable (e.g. `KUBEARCHIVE_QUERY_TIMEOUT`)
- Apply to the database query, not the full HTTP response (streaming log retrieval has its own timeout)

## Notes

- This is a safety net independent of proxy behavior — it does not rely on nginx or HAProxy closing the upstream connection
- `database/sql` respects context cancellation and sends a cancel signal to PostgreSQL
- Setting `statement_timeout` on the database role is a complementary measure but should not be the primary mechanism, as it's outside the application's control

olegbet:
Assisted-by: Claude

Resolves #1928

## Release Note

```release-note
These changes sets the timeout for database requests to prevent long-running queries to drain db connection pool. The timeout will be configured by KUBEARCHIVE_QUERY_TIMEOUT env variable and defaulted to 5 minutes.
```

### DIFF
--- a/cmd/api/main.go
+++ b/cmd/api/main.go
@@ -158,26 +158,7 @@ func main() {
 		}
 	}(db)
 
-	queryTimeout := defaultQueryTimeout
-	if s := os.Getenv(queryTimeoutEnvVar); s != "" {
-		d, parseErr := time.ParseDuration(s)
-		if parseErr != nil {
-			slog.Error("invalid query timeout, using default",
-				"env", queryTimeoutEnvVar,
-				"value", s,
-				"default", defaultQueryTimeout,
-				"error", parseErr.Error(),
-			)
-		} else if d <= 0 {
-			slog.Error("invalid query timeout, using default",
-				"env", queryTimeoutEnvVar,
-				"value", s,
-				"default", defaultQueryTimeout,
-			)
-		} else {
-			queryTimeout = d
-		}
-	}
+	queryTimeout := getQueryTimeout()
 	slog.Info("Database query timeout configured", "timeout", queryTimeout)
 
 	controller := routers.Controller{Database: db, CacheConfiguration: *cacheExpirations, QueryTimeout: queryTimeout}
@@ -236,6 +217,30 @@ func main() {
 	// This blocks until the context expires
 	<-ctx.Done()
 	slog.Debug("Shutdown reached timeout")
+}
+
+func getQueryTimeout() time.Duration {
+	queryTimeout := defaultQueryTimeout
+	if s := os.Getenv(queryTimeoutEnvVar); s != "" {
+		d, parseErr := time.ParseDuration(s)
+		if parseErr != nil {
+			slog.Error("invalid query timeout, using default", //nolint:gosec
+				"env", queryTimeoutEnvVar,
+				"value", s,
+				"default", defaultQueryTimeout,
+				"error", parseErr.Error(),
+			)
+		} else if d <= 0 {
+			slog.Error("invalid query timeout, using default", //nolint:gosec
+				"env", queryTimeoutEnvVar,
+				"value", s,
+				"default", defaultQueryTimeout,
+			)
+		} else {
+			queryTimeout = d
+		}
+	}
+	return queryTimeout
 }
 
 func getCacheExpirations() (*routers.CacheExpirations, error) {
--- a/cmd/api/main.go
+++ b/cmd/api/main.go
@@ -230,7 +230,7 @@ func getQueryTimeout() time.Duration {
 				"default", defaultQueryTimeout,
 				"error", parseErr.Error(),
 			)
-		} else if d <= 0 {
+		} else if d < 0 {
 			slog.Error("invalid query timeout, using default", //nolint:gosec
 				"env", queryTimeoutEnvVar,
 				"value", s,
--- a/cmd/api/main.go
+++ b/cmd/api/main.go
@@ -35,6 +35,8 @@ const (
 	otelServiceName                   = "kubearchive.api"
 	cacheExpirationAuthorizedEnvVar   = "CACHE_EXPIRATION_AUTHORIZED"
 	cacheExpirationUnauthorizedEnvVar = "CACHE_EXPIRATION_UNAUTHORIZED"
+	queryTimeoutEnvVar                = "KUBEARCHIVE_QUERY_TIMEOUT"
+	defaultQueryTimeout               = 5 * time.Minute
 )
 
 var (
@@ -156,7 +158,23 @@ func main() {
 		}
 	}(db)
 
-	controller := routers.Controller{Database: db, CacheConfiguration: *cacheExpirations}
+	queryTimeout := defaultQueryTimeout
+	if s := os.Getenv(queryTimeoutEnvVar); s != "" {
+		d, parseErr := time.ParseDuration(s)
+		if parseErr != nil {
+			slog.Error("invalid query timeout, using default",
+				"env", queryTimeoutEnvVar,
+				"value", s,
+				"default", defaultQueryTimeout,
+				"error", parseErr.Error(),
+			)
+		} else {
+			queryTimeout = d
+		}
+	}
+	slog.Info("Database query timeout configured", "timeout", queryTimeout)
+
+	controller := routers.Controller{Database: db, CacheConfiguration: *cacheExpirations, QueryTimeout: queryTimeout}
 	k8sClient, err := k8sclient.NewInstrumentedKubernetesClient()
 	if err != nil {
 		slog.Error("Could not create instrumented kubernetes client", "error", err.Error())
--- a/cmd/api/main.go
+++ b/cmd/api/main.go
@@ -168,6 +168,12 @@ func main() {
 				"default", defaultQueryTimeout,
 				"error", parseErr.Error(),
 			)
+		} else if d <= 0 {
+			slog.Error("invalid query timeout, using default",
+				"env", queryTimeoutEnvVar,
+				"value", s,
+				"default", defaultQueryTimeout,
+			)
 		} else {
 			queryTimeout = d
 		}
--- a/cmd/api/routers/routers.go
+++ b/cmd/api/routers/routers.go
@@ -39,12 +39,12 @@ type Controller struct {
 // database query execution only — not for the full response lifecycle — so that callers which
 // stream results can pass it solely to the initial cursor-open call; returns DeadlineExceeded
 // immediately if the parent context is already expired.
-func (c *Controller) queryContext(ctx context.Context) (context.Context, context.CancelFunc, error) {
+func (c *Controller) queryContext(ctx context.Context) (context.Context, context.CancelFunc) {
 	if c.QueryTimeout <= 0 {
-		return ctx, func() {}, nil
+		return ctx, func() {}
 	}
 	newCtx, cancel := context.WithTimeout(ctx, c.QueryTimeout)
-	return newCtx, cancel, nil
+	return newCtx, cancel
 }
 
 func (c *Controller) GetResources(context *gin.Context) {
@@ -126,11 +126,7 @@ func (c *Controller) GetResources(context *gin.Context) {
 			abort.Abort(context, errors.New("cannot use 'continue' with 'count'"), http.StatusBadRequest)
 			return
 		}
-		queryCtx, cancel, queryErr := c.queryContext(context.Request.Context())
-		if queryErr != nil {
-			abort.Abort(context, queryErr, http.StatusGatewayTimeout)
-			return
-		}
+		queryCtx, cancel := c.queryContext(context.Request.Context())
 		defer cancel()
 		count, countErr := c.Database.CountResources(
 			queryCtx, kind, apiVersion, namespace, name, labelFilters,
@@ -150,11 +146,7 @@ func (c *Controller) GetResources(context *gin.Context) {
 
 	// Single resource by exact name - no streaming needed
 	if name != "" && !strings.Contains(name, "*") {
-		queryCtx, cancel, queryErr := c.queryContext(context.Request.Context())
-		if queryErr != nil {
-			abort.Abort(context, queryErr, http.StatusGatewayTimeout)
-			return
-		}
+		queryCtx, cancel := c.queryContext(context.Request.Context())
 		defer cancel()
 		var resources []labelFilter.Resource
 		resources, err = c.Database.QueryResources(
@@ -193,11 +185,7 @@ func (c *Controller) GetResources(context *gin.Context) {
 	w := context.Writer
 
 	reqCtx := context.Request.Context()
-	queryCtx, queryCancel, err := c.queryContext(reqCtx)
-	if err != nil {
-		abort.Abort(context, err, http.StatusGatewayTimeout)
-		return
-	}
+	queryCtx, queryCancel := c.queryContext(reqCtx)
 	defer queryCancel()
 	err = c.Database.StreamResources(
 		queryCtx, reqCtx, kind, apiVersion, namespace, name, id, date, labelFilters,
@@ -295,11 +283,7 @@ func (c *Controller) GetLogURL(context *gin.Context) {
 		apiVersion = fmt.Sprintf("%s/%s", group, version)
 	}
 
-	queryCtx, cancel, err := c.queryContext(context.Request.Context())
-	if err != nil {
-		abort.Abort(context, err, http.StatusGatewayTimeout)
-		return
-	}
+	queryCtx, cancel := c.queryContext(context.Request.Context())
 	defer cancel()
 	var logRecord *interfaces.LogRecord
 	if name != "" {
@@ -348,11 +332,7 @@ func (c *Controller) GetResourceByUID(context *gin.Context) {
 		return
 	}
 
-	queryCtx, cancel, err := c.queryContext(context.Request.Context())
-	if err != nil {
-		abort.Abort(context, err, http.StatusGatewayTimeout)
-		return
-	}
+	queryCtx, cancel := c.queryContext(context.Request.Context())
 	defer cancel()
 	resource, err := c.Database.QueryResourceByUID(queryCtx, kind, apiVersion, namespace, uid)
 	if err != nil {
--- a/cmd/api/routers/routers.go
+++ b/cmd/api/routers/routers.go
@@ -41,7 +41,7 @@ type Controller struct {
 // immediately if the parent context is already expired.
 func (c *Controller) queryContext(ctx context.Context) (context.Context, context.CancelFunc, error) {
 	if c.QueryTimeout <= 0 {
-		return ctx, func() {}, errors.New("Query timeout must be greater than zero")
+		return ctx, func() {}, nil
 	}
 	newCtx, cancel := context.WithTimeout(ctx, c.QueryTimeout)
 	return newCtx, cancel, nil
@@ -137,7 +137,11 @@ func (c *Controller) GetResources(context *gin.Context) {
 			creationTimestampAfter, creationTimestampBefore)
 
 		if countErr != nil {
-			abort.Abort(context, countErr, http.StatusInternalServerError)
+			if errors.Is(countErr, dbErrors.ErrQueryTimeout) {
+				abort.Abort(context, countErr, http.StatusGatewayTimeout)
+			} else {
+				abort.Abort(context, countErr, http.StatusInternalServerError)
+			}
 			return
 		}
 		context.JSON(http.StatusOK, gin.H{"count": count})
@@ -157,7 +161,11 @@ func (c *Controller) GetResources(context *gin.Context) {
 			queryCtx, kind, apiVersion, namespace, name, id, date, labelFilters,
 			creationTimestampAfter, creationTimestampBefore, 2)
 		if err != nil {
-			abort.Abort(context, err, http.StatusInternalServerError)
+			if errors.Is(err, dbErrors.ErrQueryTimeout) {
+				abort.Abort(context, err, http.StatusGatewayTimeout)
+			} else {
+				abort.Abort(context, err, http.StatusInternalServerError)
+			}
 			return
 		}
 		if len(resources) == 0 {
@@ -215,7 +223,11 @@ func (c *Controller) GetResources(context *gin.Context) {
 
 	if err != nil {
 		if !headerWritten {
-			abort.Abort(context, err, http.StatusInternalServerError)
+			if errors.Is(err, dbErrors.ErrQueryTimeout) {
+				abort.Abort(context, err, http.StatusGatewayTimeout)
+			} else {
+				abort.Abort(context, err, http.StatusInternalServerError)
+			}
 			return
 		}
 		// Response already started - log the error, client will see truncated JSON
@@ -303,7 +315,11 @@ func (c *Controller) GetLogURL(context *gin.Context) {
 		return
 	}
 	if err != nil {
-		abort.Abort(context, err, http.StatusInternalServerError)
+		if errors.Is(err, dbErrors.ErrQueryTimeout) {
+			abort.Abort(context, err, http.StatusGatewayTimeout)
+		} else {
+			abort.Abort(context, err, http.StatusInternalServerError)
+		}
 		return
 	}
 
@@ -340,7 +356,11 @@ func (c *Controller) GetResourceByUID(context *gin.Context) {
 	defer cancel()
 	resource, err := c.Database.QueryResourceByUID(queryCtx, kind, apiVersion, namespace, uid)
 	if err != nil {
-		abort.Abort(context, err, http.StatusInternalServerError)
+		if errors.Is(err, dbErrors.ErrQueryTimeout) {
+			abort.Abort(context, err, http.StatusGatewayTimeout)
+		} else {
+			abort.Abort(context, err, http.StatusInternalServerError)
+		}
 		return
 	}
 
--- a/cmd/api/routers/routers.go
+++ b/cmd/api/routers/routers.go
@@ -126,9 +126,9 @@ func (c *Controller) GetResources(context *gin.Context) {
 			abort.Abort(context, errors.New("cannot use 'continue' with 'count'"), http.StatusBadRequest)
 			return
 		}
-		queryCtx, cancel, err := c.queryContext(context.Request.Context())
-		if err != nil {
-			abort.Abort(context, err, http.StatusGatewayTimeout)
+		queryCtx, cancel, queryErr := c.queryContext(context.Request.Context())
+		if queryErr != nil {
+			abort.Abort(context, queryErr, http.StatusGatewayTimeout)
 			return
 		}
 		defer cancel()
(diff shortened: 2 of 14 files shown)


cmd/api/main.go
```diff
@@ -145,7 +147,23 @@ func main() {
...
+	if s := os.Getenv(queryTimeoutEnvVar); s != "" {
+		d, parseErr := time.ParseDuration(s)
+		if parseErr != nil {
+			slog.Error("invalid query timeout, using default",
+				"env", queryTimeoutEnvVar,
+				"value", s,
+				"default", defaultQueryTimeout,
+				"error", parseErr.Error(),
+			)
+		} else {
```
maruiz93: **[Medium]** `time.ParseDuration` accepts negative durations (e.g. `"-5m"`). A negative value parses successfully but `queryContext` treats `<= 0` as "no timeout", silently disabling the protection this PR adds. Consider validating after parse:
```go
} else if d <= 0 {
    slog.Error("query timeout must be positive, using default",
        "env", queryTimeoutEnvVar,
        "value", s,
        "default", defaultQueryTimeout,
    )
} else {
    queryTimeout = d
}
```

cmd/api/routers/routers.go
```diff
@@ -157,8 +170,10 @@ func (c *Controller) GetResources(context *gin.Context) {
 	var headerWritten bool
 	w := context.Writer
 
+	streamCtx, streamCancel := c.queryContext(context.Request.Context())
```
maruiz93: **[Medium]** The timeout context wraps the entire `StreamResources` call including row iteration and HTTP response writes. If the timeout fires after headers are already written, the client receives truncated JSON with HTTP 200 and no way to detect the failure. This scenario becomes more likely with a configurable timeout, especially for large result sets.

Consider applying the timeout only to the initial query execution (opening the cursor) rather than the full iteration — once the DB starts returning rows, the connection pool pressure is relieved. Alternatively, document this behavior in the release notes so operators are aware.

cmd/api/routers/routers.go
```diff
@@ -31,6 +32,14 @@ type CacheExpirations struct {
 type Controller struct {
 	Database           interfaces.DBReader
 	CacheConfiguration CacheExpirations
+	QueryTimeout       time.Duration
+}
+
+func (c *Controller) queryContext(ctx context.Context) (context.Context, context.CancelFunc) {
```
maruiz93, line 40: **[Medium]** When the query timeout fires, `context.DeadlineExceeded` is passed to `abort.Abort` with 500. HTTP 504 Gateway Timeout is more accurate here — the server understood the request but the database was too slow. Distinguishing 504 from 500 also helps monitoring and alerting (slow queries vs actual errors). Consider checking for `context.DeadlineExceeded` before calling `abort.Abort` and using `http.StatusGatewayTimeout` instead.

cmd/api/routers/routers_test.go
```diff
@@ -865,3 +868,158 @@ func TestCountResourcesWithInvalidParams(t *testing.T) {
...
+
+func (s *slowDBReader) QueryLogURLByUID(ctx context.Context, _, _, _, _, _ string) (*interfaces.LogRecord, error) {
+	return nil, s.block(ctx)
+}
+
+func (s *slowDBReader) Ping(_ context.Context) error                    { return nil }
+func (s *slowDBReader) QueryDatabaseSchemaVersion(_ context.Context) (string, error) {
+	return "", nil
+}
+func (s *slowDBReader) CloseDB() error                  { return nil }
```
maruiz93: **[Medium]** `setupRouterWithTimeout` (line 918) is a near-copy of `setupRouter` (line 47) — same routes, same middleware, only difference is the `QueryTimeout` field. Future route changes need to be applied in both places. Consider extracting a shared helper that accepts a `Controller`, with `setupRouter` and `setupRouterWithTimeout` as thin wrappers.

cmd/api/routers/routers.go
```diff
@@ -31,6 +32,14 @@ type CacheExpirations struct {
...
+	if c.QueryTimeout <= 0 {
+		return context.WithCancel(ctx)
```
maruiz93: **[Low]** When `QueryTimeout <= 0`, `queryContext` calls `context.WithCancel(ctx)` which allocates a new child context on every request. Since the parent request context already has its own cancellation, you could return it directly with a no-op cancel (callers `defer cancel()`, so it can't be nil):
```go
if c.QueryTimeout <= 0 {
    return ctx, func() {}
}
```

maruiz93 (dismissed):
Left a couple of things that I think we should improve in this PR, they are no major things but it's worth fixing them now

cmd/api/routers/routers.go
maruiz93: I see that you have addressed this, nicely done :)

cmd/api/main.go
maruiz93: already addressed

cmd/api/routers/routers.go
maruiz93: I see it's addressed

cmd/api/routers/routers.go
maruiz93: The pre-check in `queryContext` handles the edge case where the parent is already expired — nice. But the common case (timeout fires mid-query) still returns 500. For example in `GetResources`:

```go
count, countErr := c.Database.CountResources(queryCtx, ...)
if countErr != nil {
    abort.Abort(context, countErr, http.StatusInternalServerError) // ← this is the timeout path
```

The error here is `context.DeadlineExceeded` — checking for it and returning 504 would give clients a meaningful signal that the query was too slow, not that the server is broken. Something like:

```go
if countErr != nil {
    if errors.Is(countErr, context.DeadlineExceeded) {
        abort.Abort(context, countErr, http.StatusGatewayTimeout)
    } else {
        abort.Abort(context, countErr, http.StatusInternalServerError)
    }
    return
}
```

The test at line 1010 currently asserts `StatusInternalServerError` for timeouts, which confirms the current behavior.

cmd/api/routers/routers_test.go
```diff
@@ -865,3 +868,158 @@ func TestCountResourcesWithInvalidParams(t *testing.T) {
...
+			assert.Less(t, elapsed, blockFor,
+				"response took %v but blockFor is %v — timeout should have fired first", elapsed, blockFor)
+		})
+	}
+}
+
+// TestQueryTimeoutDisabled verifies that when QueryTimeout is zero the handler
+// does NOT impose a timeout — the fast fake DB returns normally.
+func TestQueryTimeoutDisabled(t *testing.T) {
+	router := setupRouterWithTimeout(fake.NewFakeDatabase(testResources, testLogUrls), false, 0)
```
maruiz93, line 997: `QueryTimeout: 0` makes `queryContext` call `context.WithTimeout(ctx, 0)`, which returns an immediately-expired context. This test passes only because the fake DB ignores context cancellation — it returns data without checking `ctx.Err()`. With a real database, every query would fail immediately.

If zero is meant to mean "no timeout," `queryContext` needs an explicit check:

```go
func (c *Controller) queryContext(ctx context.Context) (context.Context, context.CancelFunc, error) {
    if errors.Is(ctx.Err(), context.DeadlineExceeded) {
        return nil, func() {}, context.DeadlineExceeded
    }
    if c.QueryTimeout <= 0 {
        return ctx, func() {}, nil
    }
    newCtx, cancel := context.WithTimeout(ctx, c.QueryTimeout)
    return newCtx, cancel, nil
}
```

If zero is NOT meant to be valid (main.go already defaults to 5m), consider removing this test or documenting that `QueryTimeout` must be positive.

cmd/api/routers/routers.go
maruiz93: The `queryContext` change (erroring on `QueryTimeout <= 0`) fixes the config-validation path, but the common case — timeout firing mid-query — still returns 500. A small helper fixes all call sites:

```go
func dbErrorCode(err error) int {
    if errors.Is(err, context.DeadlineExceeded) {
        return http.StatusGatewayTimeout
    }
    return http.StatusInternalServerError
}
```

Then replace the hardcoded 500 after each DB call:

```go
abort.Abort(context, countErr, dbErrorCode(countErr))
```

This applies to 5 sites: `CountResources`, `QueryResources`, `StreamResources`, `QueryLogURLByName`/`QueryLogURLByUID`, and `QueryResourceByUID`. The streaming case is safe — query timeout only governs cursor-open (`queryCtx`), which fires before headers are written.

The test should then expect 504:

```go
assert.Equal(t, http.StatusGatewayTimeout, res.Code,
    "expected 504 when DB call times out")
```

Also note: `TestQueryTimeoutDisabled` is now broken by the `queryContext` change — `QueryTimeout: 0` triggers the `<= 0` error path and returns 504, but the test expects 200.

olegbet: Done

docs/modules/ROOT/pages/configuration/database-query-timeout.adoc
```diff
@@ -0,0 +1,84 @@
...
+
+For streaming responses (collection queries that return multiple
+resources), the timeout governs the initial database cursor open.
+Once rows begin streaming to the client, the response uses the
+original request context so that a short query timeout does not
+truncate an otherwise healthy transfer.
+
+== Disabling the timeout
+
+Setting `KUBEARCHIVE_QUERY_TIMEOUT` to `0` disables the timeout
```
maruiz93, line 63: **[Medium]** The documentation states that setting `KUBEARCHIVE_QUERY_TIMEOUT` to `0` disables the timeout entirely, but `getQueryTimeout()` treats `d <= 0` as invalid and falls back to the 5-minute default. The `queryContext()` method already handles `QueryTimeout <= 0` correctly (returns the parent context with no deadline), but this path is unreachable via the env var.

Consider changing the guard from `d <= 0` to `d < 0` so that zero is accepted as a valid "disable" value, matching the documented behavior. Alternatively, if disabling is not intended, remove the "Disabling the timeout" section from the docs.

pkg/database/sql/query.go
```diff
@@ -23,24 +24,30 @@ func (q queryPerformer[T]) performSingleRowQuery(ctx context.Context, builder sq
...
 	err := sqlx.SelectContext(ctx, q.querier, &res, query, args...)
-	return res, err
+	return res, dbErrors.WrapQueryError(ctx, err)
 }
 
-func (q queryPerformer[T]) performStreamQuery(ctx context.Context, builder sqlbuilder.Builder, fn func(T) error) error {
+// performStreamQuery executes the query using queryCtx (e.g. one with a deadline) and
+// iterates rows under iterCtx. Separating the two contexts lets callers apply a timeout
+// only to the query-execution phase without aborting row iteration mid-stream.
+func (q queryPerformer[T]) performStreamQuery(queryCtx, iterCtx context.Context, builder sqlbuilder.Builder, fn func(T) error) error {
```
maruiz93: **[Medium]** After a closer look at Go's `database/sql` internals: the dual-context separation (`queryCtx` for cursor-open, `iterCtx` for row iteration) doesn't achieve its goal. `database/sql` binds the `Rows` object to the context passed to `QueryxContext` — an internal goroutine closes the Rows when that context is cancelled, regardless of `iterCtx`. Since `queryCtx` is also a child of the request context, client disconnection is already propagated through it, making `iterCtx` effectively unused.

Consider simplifying back to a single context and documenting this as a known limitation: the configured timeout covers both query execution and row iteration, though it's unlikely that streaming takes a noticeable amount of time in practice. Users who need a timeout scoped strictly to query execution can configure it at the database engine level (e.g. `statement_timeout` in PostgreSQL) as a complement.

pkg/database/errors/errors.go
```diff
@@ -3,6 +3,40 @@
...
+// WrapQueryError wraps database query errors to detect timeout conditions.
+// Returns ErrQueryTimeout if the error was caused by context deadline or PostgreSQL query cancellation.
+func WrapQueryError(ctx context.Context, err error) error {
+	if err == nil {
+		return nil
+	}
+
+	// Check if context deadline was exceeded
+	if errors.Is(err, context.DeadlineExceeded) {
+		return ErrQueryTimeout
```
maruiz93: **[Low]** `WrapQueryError` returns the bare `ErrQueryTimeout` sentinel, discarding the original error. Callers that log the error will always see the generic "context deadline exceeded" message with no information about the underlying cause. Using `fmt.Errorf("%w: %w", ErrQueryTimeout, err)` (Go 1.20+) preserves both the sentinel for `errors.Is()` and the original error for diagnostics.

pkg/database/errors/errors.go
```diff
@@ -3,6 +3,40 @@
...
+	}
+
+	// Check context error first in case query was cancelled
+	if ctxErr := ctx.Err(); ctxErr == context.DeadlineExceeded {
+		return ErrQueryTimeout
+	}
+
+	// Check for PostgreSQL error code 57014 (query_canceled)
+	var pqErr *pq.Error
+	if errors.As(err, &pqErr) && pqErr.Code == "57014" {
```
maruiz93: **[Low]** The string literal `"57014"` is a PostgreSQL error code for query cancellation. A named constant would make the intent clear without requiring a lookup, e.g. `const pqQueryCanceled pq.ErrorCode = "57014"`.

maruiz93 (dismissed):
`getLogsForPodSelector` at `pkg/database/sql/reader.go:171` wraps errors with `fmt.Errorf("could not retrieve resource '%s/%s': %s", ..., err.Error())`. The `%s` with `.Error()` discards the error chain, so when `performResourceQuery` returns `ErrQueryTimeout`, the wrapping breaks `errors.Is()`. The handler's check at `routers.go:302` fails, and log URL timeout errors return HTTP 500 instead of the intended 504. While this is pre-existing code, the new timeout feature depends on `errors.Is` working through the call chain. Change `%s` to `%w` and remove `.Error()` at lines 171 and 182 so the sentinel is preserved.

cmd/api/main.go
```diff
@@ -214,6 +219,30 @@ func main() {
 	slog.Debug("Shutdown reached timeout")
 }
 
+func getQueryTimeout() time.Duration {
+	queryTimeout := defaultQueryTimeout
+	if s := os.Getenv(queryTimeoutEnvVar); s != "" {
+		d, parseErr := time.ParseDuration(s)
+		if parseErr != nil {
+			slog.Error("invalid query timeout, using default", //nolint:gosec // trusted env var
```
mafh314: why this comment is needed? (`//nolint:gosec // trusted env var`)

cmd/api/main.go
```diff
@@ -214,6 +219,30 @@ func main() {
...
+				"env", queryTimeoutEnvVar,
+				"value", s,
+				"default", defaultQueryTimeout,
+				"error", parseErr.Error(),
+			)
+		} else if d <= 0 {
+			slog.Error("invalid query timeout, using default", //nolint:gosec // trusted env var
```
mafh314: same here: 
why this comment is needed? (`//nolint:gosec // trusted env var`)

mafh314:
I added a small comment. But I agreed with Marta's comments. Especially this one: https://github.com/kubearchive/kubearchive/pull/1945#discussion_r3558013602

cmd/api/main.go
olegbet: This were meant to handle linter and Qodo errors related to env var presence in error message

cmd/api/main.go
olegbet: Here is the same as previous.

pkg/database/errors/errors.go
olegbet: Compiler complaints that pq.ErrorCode is deprecated, so I'll just introduce the constant.

olegbet:
> `getLogsForPodSelector` at `pkg/database/sql/reader.go:171` wraps errors with `fmt.Errorf("could not retrieve resource '%s/%s': %s", ..., err.Error())`. The `%s` with `.Error()` discards the error chain, so when `performResourceQuery` returns `ErrQueryTimeout`, the wrapping breaks `errors.Is()`. The handler's check at `routers.go:302` fails, and log URL timeout errors return HTTP 500 instead of the intended 504. While this is pre-existing code, the new timeout feature depends on `errors.Is` working through the call chain. Change `%s` to `%w` and remove `.Error()` at lines 171 and 182 so the sentinel is preserved.

Done

docs/modules/ROOT/pages/configuration/database-query-timeout.adoc
olegbet: Done

pkg/database/errors/errors.go
olegbet: Done

pkg/database/errors/errors.go
olegbet: Moved value to a constant

cmd/api/main.go
olegbet: Removed the comment.

cmd/api/main.go
olegbet: Removed the comment.

pkg/database/sql/query.go
olegbet: Done.

cmd/api/main.go
mafh314: I think here could be useful to add a comment with the rule suppressed `//G115` (if I'm not mistaken)

pkg/database/errors/errors.go
```diff
@@ -3,6 +3,43 @@
...
+const pgQueryCancelled = "57014"
+
+// WrapQueryError wraps database query errors to detect timeout conditions.
+// Returns ErrQueryTimeout if the error was caused by context deadline or PostgreSQL query cancellation.
+func WrapQueryError(ctx context.Context, err error) error {
+	if err == nil {
+		return nil
+	}
+
+	// Check context error first in case query was cancelled
```
mafh314, line 30: I think this is not correct. `context.Canceled` is what you get when the client disconects, not when we have timeout. I think we need to treat this separately.

pkg/database/errors/errors.go
```diff
@@ -3,6 +3,43 @@
...
+	"errors"
+	"fmt"
 
-var ErrResourceNotFound = errors.New("resource not found")
+	"github.com/lib/pq"
+)
+
+var (
+	ErrResourceNotFound = errors.New("resource not found")
+	ErrQueryTimeout     = errors.New("context deadline exceeded")
```
mafh314: This error is identical to `context.DeadlineExceeded.Error()` (line29). So in line 30 we are producing:
`context deadline exceeded: context deadline exceeded`
Maybe you can change this error to:
`ErrQueryTimeout     = errors.New("database query timeout")`

cmd/api/routers/routers.go
```diff
@@ -31,6 +32,19 @@ type CacheExpirations struct {
 type Controller struct {
 	Database           interfaces.DBReader
 	CacheConfiguration CacheExpirations
+	QueryTimeout       time.Duration
+}
+
+// queryContext creates a child context bounded by the configured QueryTimeout for use during
+// database query execution only — not for the full response lifecycle — so that callers which
+// stream results can pass it solely to the initial cursor-open call; returns DeadlineExceeded
```
mafh314: I think this is not correct: `returns DeadlineExceeded immediately if the parent context is already expired` . The function doesn't return DeadlineExceeded, it returns a context.Context.
Maybe something like this:
```
// queryContext returns a child context with the configured QueryTimeout applied.
// If QueryTimeout is zero or negative, the parent context is returned unchanged.
```

pkg/database/errors/errors.go
maruiz93: Agreed — `context.Canceled` (client disconnect) shouldn't be treated as a query timeout. The fix should go further than just removing `Canceled` from the check: move it before everything else as an early return, so the 57014 path on line 40 doesn't catch client disconnects either. Since `ctx.Err()` already tells us whether it's a timeout or a disconnect, that knowledge should be authoritative:

```go
if ctx.Err() == context.Canceled {
    return err
}
```

before the `DeadlineExceeded` and pq error checks. This also relates to the unwrapped `ctx.Err()` in `performStreamQuery` (query.go:49) — once `WrapQueryError` handles `Canceled` correctly, that path can safely wrap through it too.

pkg/database/errors/errors.go
maruiz93: Agreed — changing the message to `"database query timeout"` would avoid the `"context deadline exceeded: context deadline exceeded"` duplication and give clients a clearer signal. The error response example in `database-query-timeout.adoc` (line 80) should be updated to match.

cmd/api/routers/routers.go
maruiz93: Agreed — the current wording reads as if the function returns an error. The suggested rewording is clearer.

pkg/database/errors/errors.go
maruiz93: Taking a deeper look, lib/pq v1.12.3 (already in go.mod) exports `pqerror.QueryCanceled` and a `pq.As()` helper that combines the type assertion and code check:

```go
if pqErr := pq.As(err, pqerror.QueryCanceled); pqErr != nil {
    return fmt.Errorf("%w: %w", ErrQueryTimeout, pqErr)
}
```

We should probably just use that instead of the local constant.

docs/modules/ROOT/pages/configuration/database-query-timeout.adoc
```diff
@@ -0,0 +1,86 @@
...
+
+To modify the timeout, perform the following steps:
+
+. Edit or patch the `kubearchive-api-server` Deployment in the
+`kubearchive` namespace:
++
+[source,yaml]
+----
+---
+apiVersion: v1
```
maruiz93: Nit: `apiVersion: v1` should be `apiVersion: apps/v1` for a Deployment resource.

maruiz93:
The label-resolution helpers in `postgresql.go` (`resolveKeyIDs`, `resolveLabelIDs`, `resolveLabelIDsMulti`) run SQL under the timeout context but wrap errors with plain `fmt.Errorf` instead of `WrapQueryError`. If one of those queries hits the deadline, the error won't be recognized as a timeout and the API will return 500 instead of 504. Since these are called from `ApplyLabelFilters` → `buildResourceListQuery`/`buildResourceCountQuery`, the unwrapped error propagates up through `CountResources`, `QueryResources`, and `StreamResources`.

pkg/database/errors/errors.go
olegbet: Done

pkg/database/errors/errors.go
olegbet: Done

cmd/api/routers/routers.go
olegbet: Done

docs/modules/ROOT/pages/configuration/database-query-timeout.adoc
olegbet: Done

olegbet:
> The label-resolution helpers in `postgresql.go` (`resolveKeyIDs`, `resolveLabelIDs`, `resolveLabelIDsMulti`) run SQL under the timeout context but wrap errors with plain `fmt.Errorf` instead of `WrapQueryError`. If one of those queries hits the deadline, the error won't be recognized as a timeout and the API will return 500 instead of 504. Since these are called from `ApplyLabelFilters` → `buildResourceListQuery`/`buildResourceCountQuery`, the unwrapped error propagates up through `CountResources`, `QueryResources`, and `StreamResources`.

Handled the query errors for those methods.

cmd/api/routers/routers.go
```diff
@@ -31,6 +32,20 @@ type CacheExpirations struct {
 type Controller struct {
 	Database           interfaces.DBReader
 	CacheConfiguration CacheExpirations
+	QueryTimeout       time.Duration
+}
+
+// queryContext creates a child context bounded by the configured QueryTimeout for use during
```
mafh314: I think the first three lines should be deleted now. With only this comment should be enough.
```
// queryContext returns a child context with the configured QueryTimeout applied.
// If QueryTimeout is zero or negative, the parent context is returned unchanged.
```

olegbet: Done

pkg/database/sql/query.go
```diff
@@ -23,24 +24,30 @@ func (q queryPerformer[T]) performSingleRowQuery(ctx context.Context, builder sq
...
 func (q queryPerformer[T]) performQuery(ctx context.Context, builder sqlbuilder.Builder) ([]T, error) {
 	var res []T
 	query, args := builder.BuildWithFlavor(q.flavor)
 	err := sqlx.SelectContext(ctx, q.querier, &res, query, args...)
-	return res, err
+	return res, dbErrors.WrapQueryError(ctx, err)
 }
 
+// performStreamQuery executes the query using queryCtx (e.g. one with a deadline) and
+// iterates rows under iterCtx. Separating the two contexts lets callers apply a timeout
```
maruiz93: **[Low]** This comment describes a dual-context approach (`queryCtx` + `iterCtx`) that was removed in `28f0e4b8`. The function takes a single `ctx` — consider replacing with something like:
```go
// performStreamQuery executes the query and iterates rows under ctx.
// If ctx has a deadline, it covers both query execution and row iteration.
```

docs/modules/ROOT/pages/configuration/database-query-timeout.adoc
```diff
@@ -0,0 +1,86 @@
...
+connection indefinitely, potentially exhausting the connection pool.
+====
+
+== Error responses
+
+When a query exceeds the timeout the API returns:
+
+[source,json]
+----
+{"message": "context deadline exceeded"}
```
maruiz93: **[Low]** After the `WrapQueryError` changes, the actual error message is no longer bare `"context deadline exceeded"` — it's wrapped with the `ErrQueryTimeout` sentinel (e.g. `"context deadline exceeded: query timeout: context deadline exceeded"`). Update the example to match what the API actually returns.

maruiz93 (commented):
Added a couple more comments regarding the latest changes

docs/modules/ROOT/pages/configuration/database-query-timeout.adoc
olegbet: Done

pkg/database/sql/query.go
olegbet: Done

pkg/database/errors/errors.go
olegbet: Done

mafh314 (approved):

mafh314:
@olegbet Thank you for all your work with this PR!

skoved (approved):

PR #1946, opened 2026-06-18
feat(issue-1936):  Add database connection pool configuration

### LINKED ISSUE #1936: Limit database connection number
## Problem

When receiving too many requests API server pod logs `pq: remaining connection slots are reserved for non-replication superuser connections` errors  — meaning the external PostgreSQL database ran out of available connections because the `max_connections` limit on database instance was hit.

## Expected behavior

Number of request per second to the database should be limited and configured through environment variable config. It should have a default value.

## Recommended Fix

  The primary fix is adding connection pool configuration in pkg/database/sql/connection.go:

```
  db.SetMaxOpenConns(10)       // Per pod: 2 API + 1 sink = 30 max total
  db.SetMaxIdleConns(5)
  db.SetConnMaxLifetime(5 * time.Minute)
  db.SetConnMaxIdleTime(2 * time.Minute)
```

mafh314:

Resolves #1936 

## Release Note

```release-note
- Added database connection pool configuration to prevent connection exhaustion.
- Each pod now limits open connections (default: 10) and idle connections (default: 5).
- Configurable via environment variables:
DATABASE_MAX_OPEN_CONNS
DATABASE_MAX_IDLE_CONNS 
DATABASE_CONN_MAX_LIFETIME
DATABASE_CONN_MAX_IDLE_TIME

```

## Notes for Reviewers

Assisted-by: Cursor (Claude Opus 4.6 High).

### DIFF
--- a/docs/modules/ROOT/pages/configuration/database-connection-pool.adoc
+++ b/docs/modules/ROOT/pages/configuration/database-connection-pool.adoc
@@ -1,0 +1,92 @@
+= Database Connection Pool
+
+KubeArchive limits the number of concurrent connections each pod opens to the
+database. This prevents connection exhaustion errors when the application
+receives a high volume of requests.
+
+The following environment variables control connection pool behavior:
+
+* `DATABASE_MAX_OPEN_CONNS`: maximum number of open connections to the database.
+    If not set, the default value is **25**.
+* `DATABASE_MAX_IDLE_CONNS`: maximum number of idle connections kept in the pool.
+    If not set, the default value is **10**.
+* `DATABASE_CONN_MAX_LIFETIME`: maximum time a connection can be reused.
+    If not set, the default value is **5m** (5 minutes).
+* `DATABASE_CONN_MAX_IDLE_TIME`: maximum time a connection can remain idle before being closed.
+    If not set, the default value is **2m** (2 minutes).
+
+Duration values must be parseable by
+link:https://pkg.go.dev/time#ParseDuration[time.ParseDuration]
+(e.g. `5m`, `30s`, `2h`).
+
+[WARNING]
+====
+The `DATABASE_MAX_OPEN_CONNS` value is **per pod**. The total number of
+connections to the database is this value multiplied by the number of pods
+that access the database (by default: 2 API + 1 sink = 3 pods).
+
+Ensure the total does not exceed your database's `max_connections` setting
+(PostgreSQL default: 100).
+====
+
+To modify the connection pool settings, edit or patch the relevant Deployment
+in the `kubearchive` namespace:
+
+[source,yaml]
+----
+---
+apiVersion: apps/v1
+kind: Deployment
+spec:
+  template:
+    spec:
+      containers:
+      - name: kubearchive-api-server
+        env:
+        - name: DATABASE_MAX_OPEN_CONNS
+          value: "15"
+        - name: DATABASE_MAX_IDLE_CONNS
+          value: "7"
+        - name: DATABASE_CONN_MAX_LIFETIME
+          value: "10m"
+        - name: DATABASE_CONN_MAX_IDLE_TIME
+          value: "3m"
+----
+
+Apply the patch to both deployments that access the database:
+
+[source,bash]
+----
+kubectl patch -n kubearchive deployment kubearchive-api-server --patch-file path/to/patch.yaml
+kubectl patch -n kubearchive deployment kubearchive-sink --patch-file path/to/patch.yaml
+----
+
+[NOTE]
+====
+The same environment variables apply to both `kubearchive-api-server` and
+`kubearchive-sink` since both connect to the database. Ensure you configure
+both deployments consistently.
+====
+
+== Edge Cases
+
+* **Invalid, zero, or negative values**: If an environment variable is set to a
+  non-numeric string, an unparseable duration, zero, or a negative value, the
+  application logs a warning and falls back to the default. Zero is rejected
+  because in Go's `database/sql`, `SetMaxOpenConns(0)` means unlimited
+  connections and `SetConnMaxLifetime(0)` means connections are never recycled.
+
+* **Very high `DATABASE_MAX_OPEN_CONNS`**: Setting a large value (e.g. 1000)
+  effectively disables pool limiting and may exhaust the database's
+  `max_connections` (PostgreSQL default: 100), causing the same connection
+  errors this feature is designed to prevent.
+
+* **Very long `DATABASE_CONN_MAX_LIFETIME` or `DATABASE_CONN_MAX_IDLE_TIME`**:
+  Setting durations too high (e.g. hours or days) can lead to stale
+  connections that are not recycled after a database failover, network
+  interruption, or configuration change. The defaults (5m lifetime, 2m idle)
+  balance connection reuse with timely recovery.
+
+* **`DATABASE_MAX_IDLE_CONNS` greater than `DATABASE_MAX_OPEN_CONNS`**: The
+  idle count is capped internally by Go's `database/sql` to never exceed the
+  max open count. Setting a higher idle value has no effect.
--- a/pkg/database/sql/connection.go
+++ b/pkg/database/sql/connection.go
@@ -6,16 +6,18 @@ package sql
 import (
 	"database/sql"
 	"log/slog"
+	"strconv"
 	"strings"
 	"time"
 
 	"github.com/XSAM/otelsql"
 	"github.com/avast/retry-go/v5"
 	"github.com/jmoiron/sqlx"
+	"github.com/kubearchive/kubearchive/pkg/database/env"
 	semconv "go.opentelemetry.io/otel/semconv/v1.26.0"
 )
 
-func establishConnection(driver, connectionString string) (*sqlx.DB, error) {
+func establishConnection(driver, connectionString string, dbEnv map[string]string) (*sqlx.DB, error) {
 	slog.Info("Establishing database connection")
 	configs := []retry.Option{
 		retry.Attempts(10),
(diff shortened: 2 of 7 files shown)


pkg/database/sql/connection.go
```diff
@@ -6,15 +6,29 @@ package sql
...
 )
 
+// WARNING: These defaults assume a typical deployment of 2 API pods + 1 sink pod = 30 total
+// connections against the database. If you override these via environment variables, ensure
+// the total across all pods does not exceed your database's max_connections setting
+// (PostgreSQL default: 100, max: 262,143).
+const (
+	defaultMaxOpenConns    = 10
+	defaultMaxIdleConns    = 5
+	defaultConnMaxLifetime = 5 * time.Minute
```
maruiz93: I'm wondering how this value interacts with the behaviour introduced in https://github.com/kubearchive/kubearchive/pull/1945/changes

olegbet: I think it will work with the timeout logic fine, it will wait for open connection or be closed by timeout.

pkg/database/sql/connection.go
```diff
@@ -59,6 +76,63 @@ func establishConnection(driver, connectionString string) (*sqlx.DB, error) {
...
+	if !exists || val == "" {
+		return defaultValue
+	}
+	parsed, err := strconv.Atoi(val)
+	if err != nil {
+		slog.Warn("Invalid integer value for environment variable, using default",
+			"key", key, "value", val, "default", defaultValue)
+		return defaultValue
+	}
+	if parsed < 0 {
```
maruiz93: **[Medium]** `getEnvInt` rejects negative values but allows 0. In Go's `database/sql`, `SetMaxOpenConns(0)` means **unlimited** connections — the opposite of what an operator would expect, and it undermines this feature's purpose of preventing connection exhaustion. Similarly, `SetConnMaxLifetime(0)` and `SetConnMaxIdleTime(0)` mean connections are never recycled.

Consider either treating 0 the same as negative values (reject and fall back to default), or documenting clearly in the edge cases section that 0 means unlimited/never-expire per Go's `database/sql` semantics.

pkg/database/sql/connection.go
```diff
@@ -59,6 +76,63 @@ func establishConnection(driver, connectionString string) (*sqlx.DB, error) {
 	return sqlx.NewDb(conn, driver), nil
 }
 
+func configureConnectionPool(db *sql.DB) {
```
maruiz93: **[Medium]** `configureConnectionPool` reads env vars directly via `os.LookupEnv`, but the existing pattern in this codebase reads all DB env vars upfront in `pkg/database/env/environment.go` via `NewDatabaseEnvironment()` and passes them through `Init()` as a `map[string]string`. The pool env var constants are defined in the `env` package but never included in `DbEnvVars` or that flow.

Consider adding the pool env vars to the environment map flow (or passing a config struct into `configureConnectionPool`) so pool configuration follows the same path as every other database setting. This would also improve testability — tests could pass a map instead of using `t.Setenv`.

pkg/database/sql/connection_test.go
```diff
@@ -104,3 +107,159 @@ func TestExtractConnectionInfoEdgeCases(t *testing.T) {
...
+			}
+			got := getEnvDuration(key, tt.defaultValue)
+			if got != tt.expected {
+				t.Errorf("getEnvDuration() = %v, want %v", got, tt.expected)
+			}
+		})
+	}
+}
+
+func TestConfigureConnectionPoolEnvVars(t *testing.T) {
```
maruiz93: **[Medium]** This test sets env vars and calls `getEnvInt`/`getEnvDuration` individually, which are already covered by the table-driven tests above. It never calls `configureConnectionPool()` with a `*sql.DB`, so the integration point (env vars actually wired to pool settings) is not verified.

Consider either removing it (redundant with the table-driven tests) or making it a true integration test that calls `configureConnectionPool()` and asserts `db.Stats().MaxOpenConnections`. Note: if integration tests belong in `test/integration/` per project convention, this could go there instead.

pkg/database/sql/connection.go
```diff
@@ -6,15 +6,29 @@ package sql
...
 	"time"
 
 	"github.com/XSAM/otelsql"
 	"github.com/avast/retry-go/v5"
 	"github.com/jmoiron/sqlx"
+	"github.com/kubearchive/kubearchive/pkg/database/env"
 	semconv "go.opentelemetry.io/otel/semconv/v1.26.0"
 )
 
+// WARNING: These defaults assume a typical deployment of 2 API pods + 1 sink pod = 30 total
```
maruiz93: **[Medium]** The comment reads "2 API pods + 1 sink pod = 30 total connections" which skips the intermediate step and reads as "2 + 1 = 30." Consider rephrasing to show the multiplication, e.g.: `// These defaults assume a typical deployment of 2 API pods + 1 sink pod (3 pods × 10 connections = 30 total).`

docs/modules/ROOT/pages/configuration/database-connection-pool.adoc
```diff
@@ -0,0 +1,82 @@
...
+(PostgreSQL default: 100).
+====
+
+To modify the connection pool settings, edit or patch the relevant Deployment
+in the `kubearchive` namespace:
+
+[source,yaml]
+----
+---
+apiVersion: v1
```
maruiz93: **[Low]** `apiVersion: v1` is incorrect for a Deployment — should be `apiVersion: apps/v1`.

docs/modules/ROOT/pages/configuration/database-connection-pool.adoc
```diff
@@ -0,0 +1,82 @@
...
+kind: Deployment
+spec:
+  template:
+    spec:
+      containers:
+      - name: kubearchive-api-server
```
maruiz93, line 44: **[Low]** The example and `kubectl patch` command only reference `kubearchive-api-server`, but the pool settings apply equally to the sink (and the warning already counts it in "2 API + 1 sink = 3 pods"). Users following this doc would only configure the API server. Consider adding a note that the same env vars should also be set on the sink deployment, or showing a second example.

pkg/database/sql/connection_test.go
```diff
@@ -104,3 +107,159 @@ func TestExtractConnectionInfoEdgeCases(t *testing.T) {
...
+			setEnv:       true,
+			defaultValue: 10,
+			expected:     10,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			key := "TEST_GET_ENV_INT"
+			t.Cleanup(func() { t.Setenv(key, "") })
```
maruiz93: **[Low]** `t.Cleanup(func() { t.Setenv(key, "") })` is redundant — `t.Setenv` already registers its own cleanup to restore the original value when the subtest ends. The manual cleanup also sets the var to empty string rather than unsetting it, which subtly changes semantics. Same applies in `TestGetEnvDuration` at line 231.

pkg/database/sql/connection.go
```diff
@@ -6,15 +6,29 @@ package sql
...
+// connections against the database. If you override these via environment variables, ensure
+// the total across all pods does not exceed your database's max_connections setting
+// (PostgreSQL default: 100, max: 262,143).
+const (
+	defaultMaxOpenConns    = 10
```
maruiz93: **[Medium]** The default of 10 max open connections per pod may be too conservative given observed Konflux UI traffic patterns. A single "Download all logs" click generates 47+ concurrent requests to the API (empirical data from staging — see PR #1939 review for details). With 2 API pods × 10 connections = 20 total DB connections for the API layer, most requests in a burst would queue for a pool connection. Combined with PR #1945's context timeout logic, queued requests could receive timeout errors rather than being served.

Consider whether the defaults should be higher (e.g., 25-50 per pod), or document the expected interaction between connection pool limits and concurrent request volume so operators can tune for their workload.

maruiz93 (changes requested):
Good PR! Left some comments, the one thing I'm more concerned about is about the new defaults given the calls from Konflux UI. We can always tune the correct value when deploying in staging, but we need to be mindful about that and test it out thoroughly before going to the prod clusters.

pkg/database/sql/connection.go
mafh314: you are right. I thought about the case with negative numbers but I forgot to treat 0 as a special case!

pkg/database/sql/connection.go
mafh314: Changed!

docs/modules/ROOT/pages/configuration/database-connection-pool.adoc
mafh314: Changed!

pkg/database/sql/connection_test.go
mafh314: Changed! (Also in 231).

pkg/database/sql/connection.go
mafh314: changed:
```
if parsed <= 0 {
		slog.Warn("Zero or negative duration for environment variable is not allowed, using default",
			"key", key, "value", parsed, "default", defaultValue)
		return defaultValue
}
```

pkg/database/sql/connection_test.go
mafh314: Hello! I've removed this tests and I created an integration test here: test/integration/connection_pool_test.go

docs/modules/ROOT/pages/configuration/database-connection-pool.adoc
mafh314: Added a note (line 64)

pkg/database/sql/connection.go
mafh314: Done!

pkg/database/sql/connection.go
mafh314: Changed to:
```
DbDefaultMaxOpenConns = 25
DbDefaultMaxIdleConns = 10
```

mafh314:
@maruiz93 I think I addressed all your comments. Let me know what you think :)  many thanks!

maruiz93 (approved):
looks good!

